PR #103, opened 2015-01-09
Update NodeAdminController.php

tentwofour:
Bugfix for undefined method ActionsMenuBuilder::setIsEditableNode in NodeAdminController

Looks like the method was renamed in dc7f701596d18a6a468ed8ff310b82d10eb1159c, but this call was missed in the refactoring...?


roderik:
Thx for the fix!

Issue #125, opened 2015-01-22
Has-space on tabs


indrikenens:
When there's not enough space the tabs should only be 'content' and  More ▾

Now: 
![screen shot 2015-01-22 at 17 43 03](https://cloud.githubusercontent.com/assets/772236/5860077/40e103b6-a25f-11e4-9cdd-66fa1bdce807.png)

PR #134, opened 2015-01-30
fix more dropdown tab in nav if necessary

dbeerten:
#125


IbeVanmeenen:
We should split this in a autocollapseTabs.init function and a autocollapseTabs.resize function. This way we will minimize the impact on window resize (witch has a huge performance impact).

PR #142, opened 2015-02-05
Bugfix for render_pagetemplate()

tentwofour:
- Merge passed parameters with the current twig context - allows passing of view variables when calling render_pagetemplate()
 - Argument was previously unused

### DIFF
--- a/src/Kunstmaan/PagePartBundle/Twig/Extension/PageTemplateTwigExtension.php
+++ b/src/Kunstmaan/PagePartBundle/Twig/Extension/PageTemplateTwigExtension.php
@@ -72,7 +72,7 @@ class PageTemplateTwigExtension extends \Twig_Extension
 
         $template = $this->environment->loadTemplate($pageTemplate->getTemplate());
 
-        return $template->render($twigContext);
+        return $template->render(array_merge($twigContext, $parameters));
     }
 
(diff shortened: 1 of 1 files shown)


tentwofour:
Not sure if I'm missing another way to do this, but calling 
```twig
render_pagetemplate(page, { 'customVariable': 'myGreatValue' }
```
wasn't passing customVariable to my page template.

Not sure why the argument is available, but not merged with the Twig context in the extension - if I'm missing something obvious, let me know :)

src/Kunstmaan/PagePartBundle/Twig/Extension/PageTemplateTwigExtension.php
```diff
@@ -72,7 +72,7 @@ public function renderPageTemplate(array $twigContext, HasPageTemplateInterface
 
         $template = $this->environment->loadTemplate($pageTemplate->getTemplate());
 
-        return $template->render($twigContext);
+        return $template->render(array_merge($twigContext, $parameters));
```
krispypen: It's better to put $parameters as the first element. The keys from the first array will be preserved. If an array key exists in both arrays, then the element from the first array will be used and the matching key's element from the second array will be ignored.

PR #170, opened 2015-02-24
Do not prefix database when using SQLite

WouterJ:
SQLite does not support database prefixes like this.

After this, I got a working demo site with SQLite :tada:


roderik:
seems like this broke some tests, could you could you check them out please?

WouterJ:
Tests should be fixed now

Issue #174, opened 2015-02-26
Node translations not getting a URL set when using structure nodes


JoakimLofgren:
We use structure pages to create section headers in our menu.
This causes an issue with the NodeTranslationListener which is supposed to set the URL column
on the node translations when creating pages.

In a foreach at [line 79](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/NodeBundle/EventListener/NodeTranslationListener.php#L79) in NodeTranslationListener.php when checking if the node is a StructureNode it does `return` instead of `continue`, causing the remaining node translations to be skipped.

This results in the remaining node translations not getting a URL set, which in turn gives a random root page when browsing the site.

PR #175, opened 2015-02-26
Fix nodetranslation listener

JoakimLofgren:
Proposed solution for issue #174 

I am unsure if the fix for the `autoload.php.dist` is the way to go.
I copied it from [`app/autoload.php`](https://github.com/Kunstmaan/KunstmaanBundlesStandardEdition/blob/master/app/autoload.php).

I couldn't run the tests locally without it though.
I have PHP 5.5.9 installed.

### DIFF
--- a/autoload.php.dist
+++ b/autoload.php.dist
@@ -1,14 +1,12 @@
 <?php
 
-if (version_compare(PHP_VERSION, '5.4', '>=') && gc_enabled()) {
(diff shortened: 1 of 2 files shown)


JoakimLofgren:
When will you release this?

roderik:
End of march I will tag 3.1, this will be in it.

PR #193, opened 2015-03-11
New frontend

indrikenens:


src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/defaultsite/Resources/views/Pages/ContentPage/pagetemplate.html.twig
```diff
@@ -10,6 +10,28 @@
                 </div>
             {% endif %}
 
+            {# Pageparts #}
```
krispypen: does these lines need to be here? looks for me that you where just testing things?

PR #203, opened 2015-03-17
Only export formsubmissions of the same language (aka NodeTranslation)

Lisje:
When exporting all FormSubmissions of a certain NodeTransation, the csv or excel file contains all submissions for all translations of the Node. This results in a file exported with only certain fields filled in (depending on the first language):

image

This was an export on the Dutch NodeTranslation.

As the method states it should only export all the entries of the NodeTranslation and not the Node. I propose to filter on the NodeTranslation language as well.


krispypen:
not all websites has languages, does this also works for single-language projects?

PR #213, opened 2015-03-18
ConsoleExceptionListener

woutervandamme:
added a ConsoleExcetpionListener to the adminBundle, this service can be toggled on or off in the app/config.yml file by adding the line: enable_console_exception_listener: true 
to the kunstmaan_admin section.

### DIFF
--- a/src/Kunstmaan/AdminBundle/DependencyInjection/KunstmaanAdminExtension.php
+++ b/src/Kunstmaan/AdminBundle/DependencyInjection/KunstmaanAdminExtension.php
@@ -48,7 +48,7 @@ class KunstmaanAdminExtension extends Extension implements PrependExtensionInter
         $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('services.yml');
 
-        if (!empty($config['enable_console_listener']) && $config['enable_console_listener']) {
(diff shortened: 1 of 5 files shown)


src/Kunstmaan/AdminBundle/DependencyInjection/KunstmaanAdminExtension.php
```diff
@@ -47,6 +47,10 @@ public function load(array $configs, ContainerBuilder $container)
 
         $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('services.yml');
+
+        if (!empty($config['enable_console_listener']) && $config['enable_console_listener']) {
```
jockri: `enable_console_listener` should be replaced by `enable_console_exception_listener`

src/Kunstmaan/AdminBundle/Resources/config/console_listener.yml
```diff
@@ -0,0 +1,7 @@
+services:
+    kunstmaan_admin.consolelogger.listener:
+        class: Kunstmaan\AdminBundle\EventListener\ConsoleExceptionListener
```
jockri: The classname should be configurable via a parameter. This makes it easier to overwrite the listener when needed.

roderik:
After merging this, please enable the listener by default in the StandardEdition config.yml files.

PR #215, opened 2015-03-19
migrate to google/apiclient

arneruy:


roderik:
This is a lot more complex than it seems, i worked around it for now in #220

Issue #212, opened 2015-03-18
Add wysiwyg formtype


denbatte:
At the moment a wysiwyg is done with:

```php
$builder->add(
        		'text1',
        		'textarea',
        		array(
        			'required' => false, 'attr' => array('rows' => 10, 'cols' => 600, 'class' => 'rich_editor'),
        		)
        );
```

This should be changed to something like this:
```php
$builder->add(
        		'text1',
        		'wysiwyg',
        		array(
        			'required' => false),
        		)
        );
```
So we need a WysiwygType, like we already made a MediaType

Also make sure the generators that are generating this code are changed.

woutervandamme:
https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/222

PR #233, opened 2015-03-31
Wysiwyg formtype added

denbatte:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #212

TODO
- [x] Submit changes to the documentation

WHAT
A new wysiwyg form type has been created in the Adminbundle. 
This makes it possible to use 'wysiwig' in the formbuilder as a type and results in a text area with a wysiwyg editor on top. There are two extra options you can pass as a data attribute:
- simple: will only render a basic wysiwyg editor
- full: will render a wysiwyg editor with extra buttons
The wysiwyg formtype also replaces the rich text type when generating a new pagepart using the kunstmaan generatorbundle.

WHY
Before this feature there was:
```php
        $builder->add('name', 'textarea',
                array(
                    'attr' => array(
                        'rows'  => 5,
                        'cols'  => 600,
                        'class' => 'js-rich-editor rich-editor'
                    )
                )
        );
```
now you just have to use:
```php
        $builder->add('name', 'wysiwyg');
```

### DIFF
--- a/src/Kunstmaan/AdminBundle/Form/WysiwygType.php
+++ b/src/Kunstmaan/AdminBundle/Form/WysiwygType.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Kunstmaan\AdminBundle\Form;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+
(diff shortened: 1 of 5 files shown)


roderik:
@denbatte please add a changeling item? Check with @Devolicious on how

PR #236, opened 2015-03-31
Feature/media files originalname

Devolicious:
Keeps the original name on uploaded files and stores it in a uuid directory to prevent duplicate filenames.
Also allow to configure blacklisted extensions. Defaults are .php and .htaccess

### DIFF
--- a/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
+++ b/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
@@ -33,6 +33,8 @@ class FileHandler extends AbstractMediaHandler
      */
     public $mimeTypeGuesser = null;
 
+    CONST MEDIA_PATH = '/uploads/media/';
+
     /**
      * Constructor
      */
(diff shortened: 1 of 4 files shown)


wimvds:
Did you test what happens when you upload 2 different docs that happen to have the same filename?

roderik:
Love it, but please follow this soon to be released contribution guidelines for your pull request: https://github.com/roderik/KunstmaanBundlesCMS/blob/master/docs/06-02-pull-requests.md

Also use a descriptive name for the pull request since these will go into the changeling for each release instead of all the separate commits.

src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
```diff
@@ -183,13 +200,28 @@ public function saveMedia(Media $media)
...
+    private function getFilePath(Media $media)
+    {
+        $filename = $media->getOriginalFilename();
+
+        if (!empty($this->blacklistedExtensions)) {
+            $filename = preg_replace('/\.('.join('|', $this->blacklistedExtensions).')$/', '.txt', $filename);
+        }
+
+        return sprintf(
+            '/%s/%s',
```
wimvds: Please don't create a folder per file, but group them (ie. using a variant of the old OpenMercury naming scheme).

UUID : 12345678
file : original_filename.ext

=> /1/2/3/4/12345678-original_filename.ext (UUID has to be used here to prevent overwriting files)

krispypen: Yes, if you use fat32 then the limit of files in a folder is 65,534. This way you work arround it. If we would drop fat32 this was not needed :)

PR #266, opened 2015-04-03
Require at least PHP 5.4

wimvds:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

Since PHP 5.3 is no longer maintained, we decided to drop PHP 5.3 support (if you're still using it : you *really* should upgrade!).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.9",
+        "php": ">=5.4.0",
 
         "symfony/symfony": "~2.3",
         "doctrine/orm": "~2.2,>=2.2.3",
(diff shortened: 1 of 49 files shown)


roderik:
:+1:

krispypen:
@wimvds can you also update the .travis.yml file in the root folder?

krispypen:
we should also update the requirements here: http://bundles.kunstmaan.be/documentation

roderik:
@krispypen these requirements are parsed from here: https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/docs/03-01-system-requirements.md Updates should be part of the merge request

PR #272, opened 2015-04-07
[NodeBundle] Add classname to admin if user is a super admin

kimausloos:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This adds the classname of the entity as an abbr to the admin is the user has the SUPER_ADMIN role.

add-classname


roderik:
@kimausloos fixed the title, please think of the title as the line in the changelog, it should mean something next year. Also add a [] with the touched bundles

@krispypen  :+1: for merge

Issue #251, opened 2015-04-02
[Mediabundle] Add folder option at pop-up


denbatte:
It would be a help if there was a "add folder" option at the pop-up you get when adding media.

PR #288, opened 2015-04-08
[MediaBundle] Add add/delete new folders while inside the media-chooser popup

arneruy:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #251 

This feature makes it possible to add/delete new folders while inside the media-chooser. This way, you can re-arrange the folder structure without having to exit the media-chooser.


roderik:
- [ ] Rebase on top of master

After that :+1: for merge

Issue #309, opened 2015-04-13
[MediaBundle] When adding new (local) media items, the url path contains double slashes


jockri:
This causes problems is some cases.

![screenshot 2015-04-13 09 40 57](https://cloud.githubusercontent.com/assets/517574/7111577/a69ec624-e1c1-11e4-9be3-8d3819c1c35a.png)

PR #323, opened 2015-04-14
[KunstmaanBundlesCMS] [MediaBundle] Fix single slash in media path

arneruy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #309 

Media url path does not contain double slashes anymore.


roderik:
@krispypen :+1:

krispypen:
Actually this is a fix, except that the MEDIA_PATH should be configurable and then the code should check if it ends witch / or not.

PR #330, opened 2015-04-15
[SitemapBundle] Make use of a sitemapindex and sub sitemaps

denbatte:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #4

WHAT
This feature generates one central sitemap.xml file at your document root to define a sub sitemap for each language available.
You can now have up to 50000 urls for each language you define.
As a bonus, you only have to register the sitemap index (/sitemap.xml) at the search engines to register all available languages.

HOW
In order to upgrade, you need a new route in your app/config/routes.yml.
```yml
KunstmaanSitemapBundle_sitemapIndex:
    resource: "@KunstmaanSitemapBundle/Controller/SitemapController.php"
    type:     annotation
```

TODO
- [ ] add a note in UPGRADE-3.2.md file

### DIFF
--- a/UPGRADE-3.2.md
+++ b/UPGRADE-3.2.md
@@ -1,0 +1,16 @@
+# UPGRADE FROM 3.1 to 3.2
+
+## Manual upgrade needed in app/routes to refactor sitemaps
+
+### WHAT
+This feature generates one central sitemap.xml file at your document root to define a sub sitemap for each language available.
+You can now have up to 50000 urls for each language you define.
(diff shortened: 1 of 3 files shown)


krispypen:
@roderik :+1:

roderik:
@krispypen we need to defer this to 3.2, it features a bc break because it adds a new route. Currently i'm using master as 3.1.1-dev. Let's keep this one open for a while.

roderik:
@denbatte please create an UPGRADE-3.2.md doc and add this info

PR #331, opened 2015-04-15
[AdminBundle, AdminListBundle] Add an admin page title when the adminlist is not included in the menu + make it possible to override the admin page title + fix action button position when there is no admin page title

jockri:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | []

This PR fixes these issues:
* When there was no admin page title, the action buttons (from the right top) where moved to the center of the page
* When an adminlist was not added to the `Modules` menu, no title was shown 
* Make it possible to overwrite the admin page title without duplicating the `extra_actions_header` block

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Default/layout.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Default/layout.html.twig
@@ -99,13 +99,15 @@
                     {% block header %}
                         <header class="app__content__header">
                             <div class="row">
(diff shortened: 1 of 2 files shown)


krispypen:
@roderik :+1:

roderik:
@jockri Please use a descriptive title, with bundle prefix. This title comes in the changelog and needs to be clear. Cfr the contributing guidelines

Issue #281, opened 2015-04-08
[MediaBundle] Media chooser type parameter does not work anymore


jockri:
Media chooser example url:  /nl/admin/media/chooser?type=image
This should only show images, and not other file types (like pdf)

![screenshot 2015-04-08 11 04 42](https://cloud.githubusercontent.com/assets/517574/7042096/6fcadfc4-dddf-11e4-9dcd-ff206aed32ed.png)

PR #333, opened 2015-04-15
[MediaBundle] Media chooser type parameter now works

woutervandamme:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #281

### DIFF
--- a/src/Kunstmaan/MediaBundle/Entity/Media.php
+++ b/src/Kunstmaan/MediaBundle/Entity/Media.php
@@ -291,6 +291,17 @@ class Media extends AbstractEntity
     }
 
     /**
+     * @return mixed
+     */
+    public function getType()
+    {
+        $contentType = $this->contentType;
+        $array       = explode("/", $contentType);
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/MediaBundle/Entity/Media.php
```diff
@@ -291,6 +291,17 @@ public function getContentType()
     }
 
     /**
+     * @return mixed
+     */
+    public function getType()
+    {
+        $contentType = $this->contentType;
+        $array       = explode("/", $contentType);
+
```
jockri: This will not work: for example pdf's (content-type: application/pdf) will not be shown when type 'file' is configured.
Before the CMS upgrade the type worked, can you check how this was done in the previous version?

jockri:
The media types that should be supported: image, video, file (everything that has location local).

PR #347, opened 2015-04-17
[SeoBundle] Suggest a default value in the robots.txt field when empty

denbatte:
This is a upgrade for the robots.txt. It warns you that no robots.txt has been defined and makes a default suggestion.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /


krispypen:
@roderik :+1:

PR #357, opened 2015-04-21
[SeoBundle] Fix hanging flash message, it will now disappear after save

denbatte:
The flash message did not go away after save.
Refactored the code to solve this issue

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

### DIFF
--- a/src/Kunstmaan/SeoBundle/Controller/Admin/SettingsController.php
+++ b/src/Kunstmaan/SeoBundle/Controller/Admin/SettingsController.php
@@ -30,6 +30,7 @@ class SettingsController extends BaseSettingsController
         $repo = $this->getDoctrine()->getRepository("KunstmaanSeoBundle:Robots");
         $robot = $repo->findOneBy(array());
         $default = $this->container->getParameter('robots_default');
+        $isSaved = true;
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/SeoBundle/Controller/Admin/SettingsController.php
```diff
@@ -68,6 +51,11 @@ public function robotsSettingsAction(Request $request)
             }
         }
 
+        if ($robot->getRobotsTxt() === $default) {
```
jockri: I think this is not the correct check. When someone saves the default text, there is an entry in the database, but `$robot->getRobotsTxt() === $default`is still true...

krispypen: you should show the warning only when the value is null, and if the value is null, the default should be used in the "front"

krispypen:
@denbatte what is the status of this pr?

krispypen:
@roderik :+1:

Issue #360, opened 2015-04-22
SetCookie on homepage


krispypen:
https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/37e7a27a6f84be4ee0586bbf6b1c98984bd344f4/src/Kunstmaan/AdminBundle/EventListener/SessionSecurityListener.php#L62

There should be a check added: $request->getSession()->isStarted()

PR #361, opened 2015-04-23
[AdminBundle] Prevent session hijacking checks from starting a session and preventing pages to be cached in Varnish

kimausloos:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #360


krispypen:
@roderik :+1:

PR #363, opened 2015-04-23
[GeneratorBundle] Use real bundle name if it’s not the same as the namespace

mlebkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | N/A 

A bundle may have it’s own name, not composed of it’s namespace.


roderik:
@krispypen :+1: for merge

Issue #358, opened 2015-04-21
[SearchBundle] Search page does not work anymore after the service method was replaced with a controller action


jockri:
Related commit:  https://github.com/Kunstmaan/KunstmaanBundlesCMS/commit/a83e96b628c52f8f0f251768b71fc3fe66f82c4f

Also write test to make sure this cannot happen anymore.

PR #372, opened 2015-04-24
[NodeBundle, NodeSearchBundle] Handle case when RenderContext is not an array but ArrayObject

kimausloos:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #358


krispypen:
@roderik :+1:

PR #374, opened 2015-04-25
[GeneratorBundle] Added missing check for AbstractPagePart during page part generation

stefvaessen:
Hi, in the correct repo now :) My bad, I was a bit too quick and did not read the contribution guide. Hope everything is allright now.

When generating a pagepart using Kunstmaan v3.1 the check for an AbstractPagePart only contained a string and did not check if the file existed which resulted in an error when trying to generate a pagepart. Added a file_exists to the condition.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /


roderik:
@stefvaessen great! We have a two person merging agreement, but :+1: from me :)

@krispypen :+1: for merge

Issue #376, opened 2015-04-27
Deprecated service method introduced BC break


woutervandamme:
nodeTranslation in renderContext has been changed to _nodeTranslations, this breaks some stuff

PR #377, opened 2015-04-27
[NodeBundle] Don't BC break nodeTranslation introduced when we depreciated the service method

woutervandamme:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #376


krispypen:
@roderik :+1:

kimausloos:
The case is wrong, it should be 'nodetranslation', all lowercase.

krispypen:
indeed, not tested? :)

kimausloos:
@krispypen :+1:  for me now :)

krispypen:
@roderik also for me :+1:

Issue #188, opened 2015-03-05
Custom Media Handler - Priority in MediaManager#addHandler


tentwofour:
I've created a new Media Handler (Music, locally stored, to be streamed using .htaccess for pseudo-secured streaming to MEJS player).

It's working...almost; here's the issue:

MediaHandlerCompilerPass#process looks for all services tagged with 'kunstmaan_media.media_handler', and registers them as handlers using MediaManager#addHandler.

That works, but my handler is always last in the chain, so when a Media is created, the FileHandler (default) will always be called before my MusicHandler.

Is there a way to set the priority/positioning of my handler? Maybe using the 'priority' key in the service definition, like any kernel event tagged service, and having the MediaHandlerCompilerPass#process respect that?

tentwofour:
Edit: I can put my bundle *before* Kunstmaan's in AppKernel, and my service will be found before, and injected first, so that's a temporary fix I guess ;)

krispypen:
That can indeed be a temporary fix, but we need a priority there indeed.

PR #394, opened 2015-05-05
[MediaBundle] Add priority to media handlers

arneruy:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #188

Handlers with higher priority will be preferred over handlers with lower priority when adding media.


krispypen:
@arneruy can you check the travis build log?

wimvds:
Why don't we use standard Symfony behavior for this? ie. (optionally) add a priority to the tag of the handler in the config, and sort handlers according to that priority, just like the kunstmaan_admin.menu.adaptor handling.

PR #406, opened 2015-05-13
[GeneratorBundle] Refactor pagepart media validation

arneruy:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Adding validation for media objects in pageparts is no longer obligated. You no longer have to set minimum/maximum dimension values for images if you don't want to. Instead you can skip the validation part and add it later directly into the form files if you want to. This makes for a more userfriendly experience when generating pageparts.

screen shot 2015-05-13 at 11 09 06


roderik:
@krispypen :+1: (the failed test was a scrutinizer error when transmitting, not a test)

PR #434, opened 2015-05-26
[SeoBundle] Fixed issue that OG image can not be deleted.

bureaublauwgeel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 428


roderik:
@krispypen :+1:

src/Kunstmaan/SeoBundle/Entity/Seo.php
```diff
@@ -350,7 +350,7 @@ public function getOgDescription()
      *
      * @return Seo
      */
-    public function setOgImage(Media $ogImage)
```
krispypen: Removing the type is not the fix for a null-exception. Can you change this to "Media $ogImage = null" ?

bureaublauwgeel: Should be changed.

wimvds:
Replaced with https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/452

PR #476, opened 2015-06-04
[DashboardBundle] unified the css output paths for assetic

virtualize:
Q | A
--- | ---
Bug fix? | no
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | /

previously, dashboard bundle css files were dumped to `web/css`, now they behave like every other bundle and get generated to `web/KunstmaanDashboardBundleAssets/css`, where its JS files already reside


roderik:
Thanks!

Issue #293, opened 2015-04-09
Save page on Publish


alexborton:
Currently, when creating a new page and adding content, the user must remember to Save the page _before_ hitting Publish.

If they forget, the content they have added to that page is lost AND the page is published as an empty shell.

Ideally, the page will Save automatically if the user Publishes without saving.

sandergo90:
@jockri close

PR #480, opened 2015-06-08
[NodeBundle] add button to menu so pages are saved when published

woutervandamme:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |  #293 

I've added a "save on publish" button to the menu.

 Another option would be to show a popup when publishing a page that tells the user that the page has not yet been saved.

- [ ] feedback

### DIFF
--- a/src/Kunstmaan/NodeBundle/Controller/NodeAdminController.php
+++ b/src/Kunstmaan/NodeBundle/Controller/NodeAdminController.php
@@ -592,6 +592,12 @@ class NodeAdminController extends Controller
             $page = $nodeVersion->getRef($this->em);
         }
 
+        $saveandpublish =$request->get('saveandpublish');
+
+        if($saveandpublish){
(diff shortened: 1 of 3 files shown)


denbatte:
The ideal solution would be to just save on publish.
Adding a extra option to the menu just adds complexity imo.
I don't know if a "are you sure" modal window is such a good solution.
Maybe we should have a talk about this in the next backend meeting?

krispypen:
@woutervandamme these buttons should be there: cancel - save - publish now - publish later 

and the "save",  "publish now" and "publish later" should save the page "now"

roderik:
@denbatte werk jij af? :)

FVKVN:
No validation on page publish, this saves and publishes the page even when there are blank required fields

jockri:
When you have unsaved pageparts and you click the publish button, a popup should appear.
Will create issue for this.

PR #484, opened 2015-06-10
Extending the admin User entity 3.2.*@dev

wimvds:
Hello,
I've updated kunstmaan bundles from 3.0 to 3.2.*@dev.
all works great but i have a problem with custom user entity.
i have extended the original entity by this article:

    http://bundles.kunstmaan.be/news/extending-the-admin-user-entity

but in new version i have some errors:
1 on user page it tries to show entities from original db table 
2 on add/edit i have this error 

     Error: Call to protected method Sandbox\WebsiteBundle\Entity\User::getFormTypeClass() from context 'Kunstmaan\UserManagementBundle\Controller\UsersController' 

and if i change the protected method to public i get 

     Attempted to load class "UserType" from namespace "SandBox\WebsiteBundle\Form".
Did you forget a "use" statement for "Kunstmaan\AdminBundle\Form\UserType"?

my method : 

        public function getFormTypeClass()
        {
            return 'SandBox\WebsiteBundle\Form\UserType';
        }

that file with that namespace exist(and it worked in prev version)

    Stack Trace
    in vendor/kunstmaan/bundles-cms/src/Kunstmaan/UserManagementBundle/Controller/UsersController.php at line 89  -
            /* @var $em EntityManager */
            $em                = $this->getDoctrine()->getManager();
            $user              = $this->getUserClassInstance();
            $formTypeClassName = $user->getFormTypeClass();
            $formType          = new $formTypeClassName();
            $formType->setLangs($this->container->getParameter('kunstmaan_admin.admin_locales'));

how can i fix this?

### DIFF
--- a/src/Kunstmaan/AdminBundle/Entity/BaseUser.php
+++ b/src/Kunstmaan/AdminBundle/Entity/BaseUser.php
@@ -165,6 +165,19 @@ abstract class BaseUser extends AbstractUser
         )));
     }
 
-    abstract protected function getFormTypeClass();
-    abstract protected function getAdminListConfiguratorClass();
+    /**
+     * Return class name of form type used to add & edit users
+     *
+     * @return string
+     */
+    abstract public function getFormTypeClass();
+
+    /**
+     * Return class name of admin list configurator used for users
+     *
+     * @return string
+     *
+     * @deprecated Use the kunstmaan_user_management.user_admin_list_configurator.class parameter instead!
+     */
+    abstract public function getAdminListConfiguratorClass();
 }
--- a/src/Kunstmaan/AdminBundle/composer.json
+++ b/src/Kunstmaan/AdminBundle/composer.json
@@ -14,7 +14,7 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/symfony": "~2.3",
+        "symfony/symfony": "~2.5,<2.7",
         "doctrine/orm": "~2.2,>=2.2.3",
         "friendsofsymfony/user-bundle": "2.0.*@dev",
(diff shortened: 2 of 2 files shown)


krispypen:
hmm, looks all ok, maybe you have a folder named "form" instead of "Form" and you are using OSX? :)

konstantinsp:
@krispypen 
folder names are okey. it was working correctly in previous version...

konstantinsp:
also it is not working only on /admin/settings/users/ page.
because i can log in to admin with my custom entity.
will you check what is wrong with that page?

markmedia:
Any more help with this one? We cannot manage users at the moment at all.
Adding users = Fatal error: Class 'SandBox\WebsiteBundle\Form\UserType' not found in vendor/kunstmaan/bundles-cms/src/Kunstmaan/UserManagementBundle/Controller/UsersController.php on line 90.

mtnorthrop:
I was having a similar problem with the old admin users showing up on the settings page, instead of my custom, extended user entities.

In the end, the problem was that I forgot to include this method in my custom UserAdminListConfigurator:

```
    /**
     * Get bundle name
     *
     * @return string
     */
    public function getBundleName()
    {
        return 'CompanyYourBundle';
    }
```

When I added the getBundleName() method and returned my bundle name, everything worked correctly.

I'm just leaving this here in case anybody in the future runs into a similar problem.

KevinSleegers:
I have followed all the steps of https://bundles.kunstmaan.be/news/extending-the-admin-user-entity, and checked that the getBundleName() function is present, but still the old admin users are showing up in my settings page.

Everything else is working fine, but the new users just don't show up.

PR #499, opened 2015-06-18
Make AdminLocaleListener also update locale on subrequests

bureaublauwgeel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

Since Symfony 2.6 it seems to be necessary to also update the locale on subrequests. This becomes apparent when viewing a backend pag with a subrequest while the request locale (by virtue of the default locale) is not equal to an admin locale of a backend user. In my case I had my default locale set to Dutch (nl) and the backend locale set to English (en). This resulted in the dashboard page using Dutch translations, while the rest of the backend uses English translations.

Apparently this is caused by a refactor of the TranslatorListener in Symfony 2.6, which also acts on subrequests. I'm not certain if checking for a subrequest is still needed in Symfony 2.5, if so we need to look for another solution.

### DIFF
--- a/src/Kunstmaan/AdminBundle/EventListener/AdminLocaleListener.php
+++ b/src/Kunstmaan/AdminBundle/EventListener/AdminLocaleListener.php
@@ -3,7 +3,6 @@
 namespace Kunstmaan\AdminBundle\EventListener;
 
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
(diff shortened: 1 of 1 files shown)


roderik:
Let's assume it works until someone yells at me for breaking it :)

PR #526, opened 2015-07-07
[MediaBundle] Change MediaController createAndRedirect() from private to protected

mwoynarski:
In the interest of extensibility, this method should be protected, not private.  If you want to extend this bundle and make changes to either this method, or methods that call it, you have to copy large swaths of code because a child class can't access private methods from the parent.  And if you do that, you now can't get updates to the class because you've redefined almost all the methods.


mwoynarski:
Fantastic, thanks!

Issue #109, opened 2015-01-21
Update the router to support country/language and domain based routing


roderik:
À generic version of the Mxxxxxx version
Also add http://inchoo.net/online-marketing/google-shows-meta-description-wrong-website/

yoshz:
Any update about language specific domain routing? 
We are looking into a solution to have different domains for translated pages, but the admin must work on each domain with the _locale prefix or query parameter.

roderik:
Coming in the next month or two.

yoshz:
Any idea's about this yet? 
Otherwise we will just submit a pull request with our changes.

jockri:
Solved with https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/633

PR #570, opened 2015-07-17
[NodeBundle] Added support for locale domains

yoshz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #109 

This pull request adds support for different domains for each locale (as mentioned in #109).

Users have to configure `localedomains` parameter in parameters.yml like:
```
localedomains:
  domain.com: en
  domain.fr: fr
  domain.nl: nl
  domain.de: de
```
The parameter is not mandatory for now, so it shouldn't break existing installations.

I'm not a big fan of getting parameters from the container directly, but this was already the case for multilanguage and defaultlocale so I used the same method.

**Other points to think about:**
- [ ] Refactor multilanguage parameter. Both parameters could be set but work independently.
- [ ] Update documentation
- [ ] Add parameter to standard edition

### DIFF
--- a/src/Kunstmaan/NodeBundle/Router/SlugRouter.php
+++ b/src/Kunstmaan/NodeBundle/Router/SlugRouter.php
@@ -19,7 +19,6 @@ use Symfony\Component\Routing\RouterInterface;
  */
 class SlugRouter implements RouterInterface
 {
-
     /** @var RequestContext */
     private $context;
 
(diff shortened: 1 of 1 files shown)


roderik:
@kimausloos @jockri let's compare this with our own plan and see how to learn form this approach

yoshz:
Keep me updated. Any feedback is welcome!

yoshz:
@roderik It would be nicer to have some kind of multidomain support.
If you can assign one or more hostnames to a root node it also fits the solution for different locale domains.

roderik:
I think our version will be ready next week, it goed quite a bit further than your proposal and will feature:

* domains to rootnodes mapping (so you can do multisite)
* extend locale to country-locale (nl_NL nl_BE -> both dutch, but different writingstyle and country info)
* map a domain to a locale -> http://mysite.nl/ (single lang, nl_NL locale)  http://mysite.be/nl http://mysite.be/fr (multi lang, nl_BE and fr_BE locales)

And completely configured in a YAML file

roderik:
The pull request is up, check it out #633

yoshz:
Looks like exactly what we need :+1:

PR #576, opened 2015-07-20
[AdminBundle] use `AuthorizationChecker` interface instead of a concrete implementation

mlebkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

In case someone overrides authorization checker service. Same goes for `TokenStorage` and `Router`


krispypen:
@roderik :+1:

PR #581, opened 2015-07-23
[NodeBundle] Re-add the nodemenu again and make it lazy loading

krispypen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

Removing the nodemenu from the rendercontext brought lots of problems in:

* everywhere you need it you first need to check if it's already created. You absolutely don't want to create it twice for one page rendering.
* you had to create it in twig (!= template logic) and you sometimes need it already in the service method where you get it from the rendercontext. So also there you needed to create it
* it's not always possible to create the nodemenu correct because you do not have all parameters (i.e. if it's a preview context or not)

My solution:
* make it default available again, but made the contructor empty and use lazy loading

For the sluglistener-controllerserviceaction the nodemenu was never removed (see SlugSecurityListener L 76) , so these will also become faster since this lazy loading strategy.

### DIFF
--- a/src/Kunstmaan/NodeBundle/Helper/NodeMenu.php
+++ b/src/Kunstmaan/NodeBundle/Helper/NodeMenu.php
@@ -83,6 +83,11 @@ class NodeMenu
     private $nodesByInternalName = array();
 
     /**
+     * @var bool
+     */
+    private $initialized = false;
+
+    /**
      * @param EntityManagerInterface   $em                   The entity manager
      * @param SecurityContextInterface $securityContext      The security context
      * @param AclHelper                $aclHelper            The ACL helper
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/NodeBundle/Helper/NodeMenu.php
```diff
@@ -170,6 +184,7 @@ public function getTopNodes()
      */
     public function getBreadCrumb()
     {
+        $this->init();
```
mlebkowski: I don’t think you need to `init()` the whole menu to get breadcrumbs

krispypen: Indeed, thanks!

PR #591, opened 2015-07-28
[AdminBundle] fix admin locale listener

Devolicious:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

The securitycontext getToken method also provides NULL when there is no token found.


src/Kunstmaan/AdminBundle/EventListener/AdminLocaleListener.php
```diff
@@ -75,7 +75,7 @@ public function onKernelRequest(GetResponseEvent $event)
      *
      * @return bool
      */
-    private function isAdminToken(TokenInterface $token, $providerKey)
+    private function isAdminToken(TokenInterface $token = null, $providerKey)
```
wimvds: If you assign a default value, you should revert the order of the parameters as well ...

http://php.net/manual/en/functions.arguments.php#example-171

PR #592, opened 2015-07-29
[NodeSearchBundle] fix escaping the search query

mlebkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Search would fail if the query contained an unmatched quote for an instance.


roderik:
Thx!

PR #621, opened 2015-08-09
[TranslatorBundle] Don't load database translations if the database does not exist.

roderik:


jverdeyen:
:+1:

roderik:
Still weird, only fails when using symfony 2.7

PR #630, opened 2015-08-13
[NodeBundle] Removed fallback logic for simple route loaders

jockri:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 

With these changes, it is not possible anymore to use custom route loaders (`LoaderInterface`) for defining extra routes that are handled via the controller of a `Page`. You should use a custom router (`RouterInterface`) instead. This is a BC break!

Example use case:
 - `/video` -> VideoOverviewPage with custom controller
 - `/video/123` -> no page linked tot this url, **a custom defined route** (`/{url}/{videoid}`) that is also using the controller of the VideoOverviewPage

(as discussed with @kimausloos and @wimvds)

### DIFF
--- a/UPGRADE-3.3.md
+++ b/UPGRADE-3.3.md
@@ -61,3 +61,10 @@ Refer to the DefaultAdminPanelAdaptor source to see how this is done.
 - ```get_backend_locales()``` - returns the admin (back-end) locales for the current host
 
 For multi-site / multi-domain you should make sure you no longer use the requiredlocales
+`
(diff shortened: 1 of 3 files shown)


roderik:
@jockri if it's a bc break, you need a mention in the upgrade file

jockri:
@roderik documentation added

PR #634, opened 2015-08-14
[SearchBundle] Add the locale to the request context when populating the search index

jockri:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When the search index gets populated, their was no `_locale` parameter in the `RenderContext` because the command is executed from the command line. The `_locale` parameter is in the context for browser requests, so it is also needed when populating the search index.

Example use case:
When you have this peace of code in twig, it works for browser requests, but not when populating the search index.

```
<a href="{{ path('_slug', { url: nodeTranslation.url }) }}">click here</a>
```


krispypen:
@roderick :+1:

PR #635, opened 2015-08-17
[PagePartBundle] Make it possible to override the twig template for a custom PagePartWidget

jockri:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/PagePartBundle/Helper/FormWidgets/PagePartWidget.php
+++ b/src/Kunstmaan/PagePartBundle/Helper/FormWidgets/PagePartWidget.php
@@ -3,7 +3,6 @@
 namespace Kunstmaan\PagePartBundle\Helper\FormWidgets;
 
 use Doctrine\ORM\EntityManager;
-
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/PagePartBundle/Helper/FormWidgets/PagePartWidget.php
```diff
@@ -128,11 +128,19 @@ public function getFormErrors(FormView $formView)
     }
 
     /**
+     * @param string $template
```
wimvds: Why did you add these here?They are already in the FormWidget class (& interface)that this class inherits from...

jockri:
@wimvds fixed it

krispypen:
@roderik :+1:

tarjei:
Simple question. If I understand this correctly, then this can be used to override the template shown in the Admin section of the site. What should I do in my code to use this?

PR #641, opened 2015-08-19
[MultiDomainBundle] Fix host override

wimvds:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

The host override didn't work because the session was not started, refactored so it uses a cookie instead.


krispypen:
@roderik :+1:

PR #642, opened 2015-08-19
[SeoBundle] Logical operators should be avoided

anvanza:
Logical operators should be avoided
You should use '&&' instead of 'and' and '||' instead of 'or'


krispypen:
@roderik :+1:

PR #657, opened 2015-08-25
[FixturesBundle] Pass arguments to Faker provider

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

Currently no arguments are passed to Faker when using its providers in the FixturesBundle. Because Faker uses `__call` to handle calls to its providers, the methods are not found using reflection. The method is however callable, but was called without the parsed arguments. This PR makes sure those arguments are passed.


roderik:
Thx!

PR #677, opened 2015-09-01
[MenuBundle] Refactor MenuBundle to use a MenuService

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

In order to create MenuItems via fixtures, the logic of creating MenuItems should be in a service class instead of the controller.

This PR removes this logic from the controller and moves it into a service. I've created a wrapper function in the controller to avoid BC breaks.


jockri:
:+1:

src/Kunstmaan/MenuBundle/Controller/MenuAdminListController.php
```diff
@@ -66,47 +66,14 @@ public function indexAction(Request $request)
     }
 
     /**
-     * Make sure the menu objects exist in the database for each locale.
+     * Wrapper function for BC breaking
      */
     private function makeSureMenusExist()
```
mlebkowski: the method is private, you can safely remove it

roderik:
@jverdeyen kan je die opmerking nog fixen? Dan gaat hij er in. Ik zou graag 3.3 taggen deze week

jverdeyen:
done!

PR #679, opened 2015-09-01
Hide pagepart dropdown when there are no options configured

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Sometimes u use a context as a placeholder (for a very custom context) on a page.
A user can't add any pageparts to it, so there shouldn't be a useless dropdown with no select options.


jockri:
Btw, you can also use a template file as a region in the backend. This is probably what you are looking for.
```
name: "Test page"
rows:
    - regions:
        - { name: "region_1", span: 12 }
        - { name: "description", span: 12, template: "YourCompanyWebsiteBundle:Pages:TestPage/description.html.twig" }
        - { name: "region_2", span: 12 }
```

PR #682, opened 2015-09-03
[RedirectBundle] Add domain field when configuring redirects

jockri:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When you have a website that has multiple domains, it was not possible to use the same origin url and point it to a different target url for each domain. This PR solves this by allowing the administrator to choose the domain for which the redirect should work. It is also possible to choose "All domains".

When there is only a single domain configured, the domain input field will not be shown and the domain column will not be visible in the adminlist.

Eg:
domain.com/cola -> domain.com/actions/cola
domain.nl/cola -> domain.nl/acties/cola


krispypen:
@roderik :+1:

PR #688, opened 2015-09-04
[MultiDomainBundle] Replaced host override cookie with session

yoshz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

I had a problem with the host override cookie was unset by Varnish because it is prefixed with underscore.
As it is a session cookie it is easy replaced by using the request session instead of an extra cookie.

### DIFF
--- a/src/Kunstmaan/MultiDomainBundle/Controller/SiteSwitchController.php
+++ b/src/Kunstmaan/MultiDomainBundle/Controller/SiteSwitchController.php
@@ -4,7 +4,6 @@ namespace Kunstmaan\MultiDomainBundle\Controller;
 
 use Kunstmaan\MultiDomainBundle\Helper\DomainConfiguration;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
-use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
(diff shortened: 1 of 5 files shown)


jockri:
All admin routes (/admin/xx) should never be cached by Varnish. So the host override cookie should work as expected. Or do you have an other use case?

yoshz:
Varnish doesn't cache the response, so that isn't the problem.
Our Varnish VCL strips all _* cookies from the request. The convention is to prefix a cookie with an underscore only when you use the cookie clientside with javascript, like Google Analytics does. Otherwise varnish will not cache the response when a cookie is set.

This pull request is just another way of overriding the host without removing the underscore which solves my usecase and probably that of others.

jockri:
@wimvds What was the reason why you moved from using sessions to cookies?

roderik:
@wimvds ping

wimvds:
A session was created on every request (because $session->has starts a session when no session is started yet). Putting it inside a $session->isStarted() test did not seem to work (and I didn't have time to find out why). That's why I refactored it to use a cookie instead.

mlebkowski:
@wimvds, use [`$request->hasPreviousSession()`](http://api.symfony.com/2.0/Symfony/Component/HttpFoundation/Request.html#method_hasPreviousSession)

jockri:
@yoshz can you rebase please.

Using the session is ok for us. It looks like there is some sort of convention that cookies that start with underscores can be removed in Varnish. So, or we remove the underscore, or we use the session.

yoshz:
Done

roderik:
@yoshz the tests failed

yoshz:
Uhh.. I guess someone added tests in meantime :). 
I will look into it.

yoshz:
All checks have passed! Please merge otherwise I keep rebasing :)

roderik:
done :)

PR #689, opened 2015-09-04
Make it possible to add extra JS to the Admin area

tarjei:
Without extending layout.html.twig.

Since layout.html.twig is an important file, you do not want to extend
it. Instead extend the
KunstmaanAdminBundle:Default:_js_extra_footer.html.twig file to add an
extra JS file or two.

Usefull for example if you are creating advanced form types to include
in editing a PagePart.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Default/layout.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Default/layout.html.twig
@@ -141,5 +141,10 @@
 
     <!-- JS - Extra -->
     {% block extrajavascript %}{% endblock %}
+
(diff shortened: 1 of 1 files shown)


krispypen:
@tarjei there was a PR yesterday for exactly the same, have a look at #683

tarjei:
Fantastic! And a much better PR as well :+1: 

When do you think there will be a release containg that pr?

krispypen:
This should be today or monday :)

Issue #701, opened 2015-09-09
Problems with StructureNodes


wesleylancel:
I'm currently using a few StructureNodes to group certain pages in the CMS. This works pretty good, however I've started running in to some issues. As a StructureNode doesn't have any NodeTranslations that are online, it'll be skipped by the getChildren in NodeMenu by default. As this function is used by the SitemapBundle and the SearchBundle, any children of StructureNodes will not be visible in the sitemap or be indexed for searching. Is there an existing work around for this? If not, how would I go about writing a solution?

mlebkowski:
Yeah, I ran into this so I have a listener:

```php
<?php
use Kunstmaan\NodeBundle\Event\Events;
use Kunstmaan\NodeBundle\Event\NodeEvent;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;

class StructureNodeListener implements EventSubscriberInterface
{
    public function onNodeSave(NodeEvent $event)
    {
        $page = $event->getPage();
        if (false === $page->isStructureNode()) {
            return;
        }

        $node = $event->getNode();
        $nodeTranslation = $event->getNodeTranslation();

        $node->setHiddenFromNav(true);
        $nodeTranslation->setOnline(true);
        $nodeTranslation->setUrl('');
    }

    public static function getSubscribedEvents()
    {
        return [
            Events::PRE_PERSIST => 'onNodeSave',
            Events::ADD_NODE => 'onNodeSave',
        ];
    }
}
```

I believie this could be the default. I can prepare a PR with this in the future, but currently I’m a little swamped.

roderik:
Seems like a good solution, a PR is welcome

jverdeyen:
I think most of these "decisions" should be in listeners. As the default behaviour for adding paging will diverge from the fixtures' bundle behaviour. For example, the possible children check when adding a page should also throw an exception when using fixtures. My two cents ;-)

:+1:

PR #702, opened 2015-09-10
[NodeBundle] Option to fix `StructureNodes` on save

mlebkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #701

I don’t have my project env set up, so I wasn’t able to test it, but it seems to be straight forward enough to make a PR anyway.

And since the behaviour may cause problems to some, I added a config switch for it (off by default).

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/doc/FormCollectionAdditions.md
+++ b/src/Kunstmaan/AdminBundle/Resources/doc/FormCollectionAdditions.md
@@ -82,7 +82,7 @@ form type has a getter and setter for a weight field if you don't set it).
 ## Many to Many relations
 
 If you have a many-to-many relation, you wouldn’t want to delete the related entity when removing it from page part.
-In order to skip this, set 'nested_deletable => false` attribute, i.e.:
+In order to skip this, set `nested_deletable => false` attribute, i.e.:
 
 ```
 'attr' => array(
--- a/src/Kunstmaan/NodeBundle/DependencyInjection/Configuration.php
+++ b/src/Kunstmaan/NodeBundle/DependencyInjection/Configuration.php
@@ -50,6 +50,8 @@ class Configuration implements ConfigurationInterface
                 ->end()
             ->end();
 
+        $root->children()->booleanNode('fix_structure_nodes')->defaultFalse();
+
         return $treeBuilder;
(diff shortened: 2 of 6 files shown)


jverdeyen:
Maybe we shouldn't create settings for bug fixes to stay away from BC break?

mlebkowski:
I don’t really have an opinion on that topic. I can refactor this into the core if you decide that’s a better solution.

denbatte:
Hi @mlebkowski ,

Thank you for your pull request. We discussed this internally and came to the conclusion that there are some issues with it:

- The config option "fix..." doesn't really tell other users what it actually does.
- The actual problem you are experiencing is better solved by implementing a custom search configuration.

wesleylancel:
@denbatte I agree that the config option is this PR might not be the best option, but I would expect this behaviour to be default regardless of any configuration. The problem is not only with searching but also happens in for example the SitemapBundle, because the actual problem is deeper inside the NodeMenu. All children of StructureNode's are discarded as being offline because the StructureNode itsself doesn't have any online NodeTranslations. As far as I know StructureNodes are meant to be used to add children to, so that multiple pages can be logically grouped. It seems weird to skip these children in other parts of the bundles.

mlebkowski:
I’m not sure that’s the case in the latest version, but somewhere around 3.1 there was an issue with StructureNodes appearing in breadcrumbs. I believe it doesn’t have much sense for the majority of users.

Anyway, I myself am ambivalent about this change. It works perfectly fine as a listener in my projects.

Maybe there should be a list of bundles / plugins / patches that may prove useful for users, but don’t fit for the core package?

roderik:
@krispypen don't know if I can merge this, could you check?

kimausloos:
@wesleylancel The StructureNode was originally used to group nodes that should not appear online or show up in search, they just contained content we needed to place somewhere and use somewhere else. For example, we created StructureNodes to store custom menu information, temporary landing pages (not to be indexed), just pages of content needed on different locations in long flows, ....

All those use cases required the content to be both not indexed and not appear in the sitemap, hence StructureNodes don't even have a NodeTranslation...

Issue #684, opened 2015-09-04
[MultiDomainBundle] Can not generate url to a node on a different host


yoshz:
I used the MultiDomainBundle #633 to successfully implement multilingual site with a different domain for each locale.
However I'm not sure how I can easily generate an url to a translation on a different hostname.
This functionality seems te be missing in DomainBasedLocaleRouter.

In pull request #570 I was achieving this by setting the host parameter when you generate url.

@wimvds, @krispypen: Any thoughts about a proper solution? Im about to implement my way by overridding the hostname with the first found hostname by current locale in generateUrl().

yoshz:
Could anyone please comment on this issue? I would like to spend time to solve this issue, but it would be nice if I know the solution i'm thinking about is the right solution.

denbatte:
@jockri

yoshz:
Yesterday I tried to find a proper solution. But it isn't that easy to generate an url inside the router to a different host as you don't have scope on the nodeTranslation and don't know the root node.
It would be easier if it would be allowed to set the node or nodeTranslation as parameter when you generate an url. From the node it is easier to determine the root node and use the internal name to find the first host which matches. But I think it is to much effort for now.

For the current usecase I only need to determine the right host for a translation of the current active node. So I'm thinking about adding a `get_translation_host(locale)` twig function. What do you think?

PR #703, opened 2015-09-10
[NodeBundle] Added get_translation_url() twig method

yoshz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #684

With this method you can generate a translation url for a node from a twig template.
It also works (and especially) when the MultiDomainBundle is enabled.

The SlugRouter is modified to allow host parameter to be set, which allows you to change the hostname when generating an absolute url.

This could probably also be used to generate url's to nodes on different hosts by setting the host parameter. However you probably want a wrapper function to determine the right hostname for a node.


denbatte:
Hi @yoshz ,

Thank you for this pull request.
We discussed this pull request internally and decided not to merge it, because it handles a specific edge case. You can override the existing services and router to implement this behaviour.

yoshz:
I implemented the twig function in our own bundle without overriding the router. 
You can safely ignore this pull request ;)

PR #704, opened 2015-09-10
[MultiDomainBundle] Added host aliases support

yoshz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

Added the ability to configure aliases for each host. 
With this configuration each host can have multiple domain names without adding a new host configuration for each domain.


denbatte:
Hi @yoshz ,

Just a quick question: why would you define multiple aliases?
Could you explain your use case a bit more?

yoshz:
If you have multiple domains for the same site you only have to add an alias and avoid an extra site in the admin selector.
Users don't understand why a website is in the selector twice but uses the same page tree.

I currently use this functionality to add a test domainname alias next to the live domainname to a existing site.

yoshz:
I implemented a different way to solve our use case by overriding DomainConfiguration.
You still can merge this pull request if you think it usefull, otherwise you may close it.

roderik:
@jockri this is the feature i was talking about :) I want this too :)

jockri:
Looks like a good solution. 
When you want different `extra` parameter values for each (aliased) domain, you can always define them separately.

roderik:
@yoshz can you rebase your branch so i can merge this?

PR #706, opened 2015-09-14
[Nodebundle] [critical] Fix services.yml error in NodeBundle

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

```
  Catchable Fatal Error: Argument 2 passed to Symfony\Component\DependencyInjection\Definition::addMethodCall() must be of the type array, object given, called in /vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php on line 253 and defined  
```

This is caused by services.yml in the NodeBundle.


jverdeyen:
Note: I'm using sf 2.7.3

PR #710, opened 2015-09-15
[NodeBundle] performance fixes

wimvds:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

This will fix performance issues with sites that contain lots of nodes that were created with the page creator. You will still have to run the following query on these projects once after updating though (using a migration or by hand on the DB) : ```UPDATE kuma_nodes SET internal_name = NULL WHERE internal_name = ''```


krispypen:
@roderik :+1:

PR #715, opened 2015-09-16
[All Bundles] removed deprecated notices

piotrbelina:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | |no
| Deprecations? | no
| Fixed tickets |


mlebkowski:
hold on, hold on! `choice_label` for example is only available in Symfony 2.7, this is a breaking change.

mlebkowski:
I guess that changing the FormTypes methods are also a breaking change:
```diff
-    public function setDefaultOptions(OptionsResolverInterface $resolver)
+    public function configureOptions(OptionsResolver $resolver)
```

This way if someone extends `setDefaultOptions` and uses `parent::setDefaultOptions`… ?

roderik:
ow crap

roderik:
@piotrbelina could you check this out?

piotrbelina:
@mlebkowski thank you for noticing it
@roderik it will revert setDefaultOptions and choice_label changes

Issue #586, opened 2015-07-27
[Admin] Media subfolder position are not saved via drag and drop.


dimitrizetzsche:
As a user I can drag and drop subfolders in the media (images) folder so I can sort them accordingly. Unfortunately the sorting order of the folders isn't saved to the database. When I refresh my screen the three structure is sorted back according it's sorting algorithm (alphabetically?).

You can reproduce this problem by performing following steps:
1. Create following folders in de media images folder: folder 1, folder 2, folder 3. You should have a three structure like following example 
![step1](https://cloud.githubusercontent.com/assets/3658214/8899844/e3d1d0fc-3439-11e5-9151-77c66ced33bd.png)
2. Drag and drop folder 3 to the first position, above folder 1. This will give you a three structure like following example 
![step2](https://cloud.githubusercontent.com/assets/3658214/8899854/fca02566-3439-11e5-8ea8-8aa005439afd.png)
3. Refresh you screen.
4. The folders are sorted back to their initial state like in step 1.

PR #718, opened 2015-09-17
[MediaBundle] fix folder position saving via drag and drop

piotrbelina:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #586


krispypen:
Drag and drop to move folders to other parents yes, but I think we also want them to be sorted alphabetically?

piotrbelina:
@krispypen I fixed only the folder moving as described in #586

krispypen:
@roderik :+1:

PR #726, opened 2015-09-23
[KunstmaanAdminBundle] Refactor DomainConfiguration

wimvds:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 

This PR fixes projects that don't use the KunstmaanNodeBundle, so everything works for them as it should again. The DomainConfiguration had to be moved to KunstmaanAdminBundle, so this is a breaking change (it will only impact users that already used the domain configuration in their own projects though).

Upgrade instructions :
- replace all occurences of ```kunstmaan_node.domain_configuration``` with ```kunstmaan_admin.domain_configuration```
- replace all references to ```Kunstmaan\NodeBundle\Helper\DomainConfigurationInterface``` with ```Kunstmaan\AdminBundle\Helper\DomainConfigurationInterface```


roderik:
@wimvds I want to tag 3.4.0 but I need a UPGRADE-3.4.md file with the change to go from 3.3 to 3.4 in regards to this pull request.

PR #731, opened 2015-09-28
[AdminBundle] Fix admin sidebar when NodeBundle is not used

jockri:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Default/sidebar.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Default/sidebar.html.twig
@@ -14,7 +14,7 @@
 <!-- Tree -->
 {% set currentMenuItem = adminmenu.current %}
 <nav role="navigation" id="app__sidebar__navigation" class="app__sidebar__module app__sidebar__navigation"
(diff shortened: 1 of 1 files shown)


wimvds:
As is, this will break all sites. It needs a better solution then this.

jockri:
This will not break all site. I just used the same check like in the main layout.html.Twig: https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/AdminBundle/Resources/views/Default/layout.html.twig#L27
The variable is set in the Nodebundle: https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/7b216e569f05a14b1df597b667474a3994e80d0f/src/Kunstmaan/NodeBundle/DependencyInjection/KunstmaanNodeExtension.php#L44

I agree that we should find a better solution for the variable, but for the moment it is not possible to remove the NodeBundle from the stack without this quick change.

wimvds:
Ok, didn't know about that global. The whole sidebar handling in the admin still needs a better solution though...

krispypen:
@roderik :+1:

PR #740, opened 2015-10-04
[AdminBundle] Make PagePartEditor emit events on add, edit and delete

### LINKED ISSUE #695: Can the admin pageparts emit an event when a new patepart is created?
Hi, I've made a pagepart that uses some JS in the admin form to show / hide fields and also to run a select2 selector on one of the fields. My problem now is in having an event to hook up some init code when a new pagepart is created.

Would it be possible for _pagepart_editor.js to emit an event when a new pagepart is created containing some metadata (pagepart class for example) that external scripts could listen to?

tarjei:
This can be used if you need advanced JS functionality for custom
pageparts.

Fixes #695

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #695

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/ui/js/_pagepart-editor.js
+++ b/src/Kunstmaan/AdminBundle/Resources/ui/js/_pagepart-editor.js
@@ -2,7 +2,13 @@ var kunstmaanbundles = kunstmaanbundles || {};
 
 kunstmaanbundles.pagepartEditor = (function(window) {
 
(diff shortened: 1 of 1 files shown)


roderik:
Thx!

tarjei:
Great :-)
A point release with this pr would be just swell.
4. okt. 2015 4.44 p.m. skrev "Roderik van der Veer" <
notifications@github.com>:

roderik:
We just need a UPGRADE-3.4 file from @wimvds and i can tag :)

PR #761, opened 2015-10-16
[SeoBundle] Extra article seo meta tags

denbatte:
This PR adds extra fields to the SEO OG tab. The fields however are conditional. 
They only show up if the user selects "Article" in the OG Type field.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /


roderik:
@denbatte add a small note to the Upgrade 3.4 file that mentions you need to make migrations for 3.4.1

PR #779, opened 2015-10-23
[NodeBundle] Make Doctrine connection configurable

joker806:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

Makes database connection used by `gedmo.listener.tree` configurable. Previously the `default` connection was hardcoded.

Configuration:
```
    kunstmaan_node:
        connection: content_database
```

I tried to update just service definition in `services.yml` like this:
```
    gedmo.listener.tree:
        class: Gedmo\Tree\TreeListener
        tags:
            - { name: doctrine.event_subscriber, connection: %kunstmaan_node.connection% }
        calls:
            - [ setAnnotationReader, [ @annotation_reader ] ]
```
but for some reason, the parameter value doesn't get resolved so i moved definition to to extension class.


mlebkowski:
@joker806, you can do it the way you described, you just have to convert the configuration variable to a parameter:

``` php
# KunstmaanNodeExtension.php
$container->setParameter('kunstmaan_node.connection', $config['connection']);
```

jockri:
Well I think that the connection is already configurable, as the `config.yml` is a configuration file where you just can update the connection name used. Or am I missing something?

Btw, the tree extension is not only used in the NodeBundle but also in the MediaBundle and even in your own bundles.

joker806:
@mlebkowski already tried that, but when I set a parametrized attribute in tag (`connection: %kunstmaan_node.connection%`) I always get error like this:
```
RuntimeException in RegisterEventListenersAndSubscribersPass.php line 131:
The Doctrine connection "%kunstmaan_node.connection%" referenced in service "gedmo.listener.tree" does not exist. Available connections names: connection1, connection2
```

Even if I try setting some dummy parameter value directly in `services.yml`. The parameter value never gets resolved.

@jockri that's true, but we need these bundles to use other connection than `default`.

joker806:
Are there any more thought about this?

jockri:
I don't understand the use case. Can you explain your problem a bit more?

joker806:
We use multiple database connections in our application :
 * `content_database` for `kuma_*` (and other content) tables
 * `default` for shop related data

The `default` connection can't be changed for historical reasons.

The problem is that `gedmo.listener.tree` has hardcoded `default` connection. So it attaches to shop database instead of the content.

joker806:
In theory, the the `default` is just name of connection and it doesn't need to exist at all.

jockri:
Ah, I think I understand the problem now.

The gedmo tree extension is used in the NodeBundle, but also in the MediaBundle. This means that the service should not be defined in the NodeBundle. Can you try to remove the service definition from the NodeBundle, and put it in the main `config.yml` file. By doing this, it is also easy to replace the database connection that should be used.

```
gedmo.listener.tree:
        class: Gedmo\Tree\TreeListener
        tags:
            - { name: doctrine.event_subscriber, connection: default }
        calls:
            - [ setAnnotationReader, [ @annotation_reader ] ]
````

joker806:
Yes, this works and I consider it to be a cleanest solution. But as it would introduce a breaking change, I rather tried to find another solution - make existing service configurable.

jockri:
Yes, it is a breaking change, but the best solution.

Can you update this PR (or start a new one) and also document the BC break in the latest UPDATE file?
The `config.yml` should be changed in the https://github.com/Kunstmaan/KunstmaanBundlesStandardEdition , so an extra PR is required.

Thanks!

roderik:
@jockri @joker806 please create a UPGRADE-3.5.md file for this, since breacking changes can only be merged in 3.x releases.

joker806:
Replaced by #846

PR #789, opened 2015-10-29
[NodeBundle] Change publish date stepping to 15 minutes

yoshz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no

A user can schedule a node publication every minute.
We rather let the you user schedule the publication at 0, 15, 30, 45 minutes so we only have to run the cronjob 4 times an hour instead of 60 times an hour.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/ui/js/_datepicker.js
+++ b/src/Kunstmaan/AdminBundle/Resources/ui/js/_datepicker.js
@@ -11,7 +11,8 @@ kunstmaanbundles.datepicker = (function($, window, undefined) {
         defaultCollapse = true,
         defaultKeepOpen = false,
         defaultMinDate = false,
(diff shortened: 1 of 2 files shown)


roderik:
@krispypen @krispypen personally i'm ok with this (+ a PR on the standard edition). Are there any usecases for every minute?

jockri:
We should also do this for the unpublish later then.
 
What about creating bundle config properties for this? By default we can set it to 5 minutes, but it will be easy to override for each project.

Something like:
```
kunstmaan_node:
    publish_stepping: 15
    uppublish_stepping: 15
```

yoshz:
(y)

jockri:
Closed in favor of #1191

Issue #810, opened 2015-11-10
Attempted to call an undefined method named "validateTargetDir"


nochecksum:
SensioGeneratorBundle Validator's support for validateTargetDir() [was removed](https://github.com/sensiolabs/SensioGeneratorBundle/commit/baab95e63abd6b0b39619888b0d83e5c0171d9f3#diff-ca9769fcce13abde4b5b0e1878f28c56L90) but called from [GenerateBundleCommand](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/GeneratorBundle/Command/GenerateBundleCommand.php#L86)

Steps to repeat:
`$ composer create-project kunstmaan/bundles-standard-edition myproject` with defaults
`$ app/console kuma:gen:bun`
Namespace `Nochecksum/WebsiteBundle` with defaults

```
PHP Fatal error:  Call to undefined method Sensio\Bundle\GeneratorBundle\Command\Validators::validateTargetDir() in /screencasts/myproject/vendor/kunstmaan/bundles-cms/src/Kunstmaan/GeneratorBundle/Command/GenerateBundleCommand.php on line 165
PHP Stack trace:
PHP   1. {main}() /screencasts/myproject/app/console:0
PHP   2. Symfony\Component\Console\Application->run() /screencasts/myproject/app/console:27
PHP   3. Symfony\Bundle\FrameworkBundle\Console\Application->doRun() /screencasts/myproject/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:120
PHP   4. Symfony\Component\Console\Application->doRun() /screencasts/myproject/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:96
PHP   5. Symfony\Component\Console\Application->doRunCommand() /screencasts/myproject/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:189
PHP   6. Symfony\Component\Console\Command\Command->run() /screencasts/myproject/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:846
PHP   7. Kunstmaan\GeneratorBundle\Command\GenerateBundleCommand->interact() /screencasts/myproject/vendor/symfony/symfony/src/Symfony/Component/Console/Command/Command.php:241
[...]

PR #812, opened 2015-11-11
Fixed wrong dependency for sensio/generator-bundle 

yoshz:
As it is not compatible with v3.0

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #810


jockri:
Merged it to fix the problem, but we also should remove the `validateTargetDir ` function like mentioned in https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/810

PR #816, opened 2015-11-13
Videopp

morenonaitana:
Checking for either Youtube or Vimeo or Dailymotion for the URL link


FVKVN:
can you squash all your commits into 1 commit and include the pr table from the coding guidelines

roderik:
Also add all these favicons and manifest files. Create the bundles favicon by using http://realfavicongenerator.net  

also document that they can be generared with this tool http://realfavicongenerator.net

PR #818, opened 2015-11-13
new favicons

morenonaitana:
Adding new favicons line


roderik:
Also add all these favicons and manifest files. Create the bundles favicon by using http://realfavicongenerator.net  

also document that they can be generared with this tool http://realfavicongenerator.net

morenonaitana:
Apparently the favicons are not linked to the GeneratorBundle except for the Twig. Instead, Ruud and Dries told me than we should swap the favicons from this repo https://github.com/Kunstmaan/KunstmaanBundlesStandardEdition

PR #821, opened 2015-11-13
[GeneratorBundle] Videopp adding compatibility for Vimeo and Dailymotion

morenonaitana:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


FVKVN:
I think the placeholder image should come directly from the video provider as well?

Sambego:
Yep, what Kevin said, the placeholder image should come from the video provider's api, unless the user specifies an image, in that case I think we can override it.

jockri:
The placeholder image from the video provider can be added in an other PR.

PR #822, opened 2015-11-13
CSV export: Fix issue where only first row below the header has content

RX01:
Fix for the issue where only the first row below the header row has actual table row content

...so rows actually contain the information they should contain instead of only the very first result.

This fix is tested on Kunstmaan 3.2.2 so that is why it is commited on branch 3.2.

FYI: the Excel export has the same issue.


jockri:
@RX01 can you remove the leading whitespaces from the twig file? We don't want those in the csv export. Thanks!

PR #847, opened 2015-11-29
Update LiipImagineBundle

joker806:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

Update legacy LiipImagineBundle 0.20.2 to 1.0+.

We use this with version 1.3.1 in production for about 2 weeks and no problem occured so far.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "lunetics/locale-bundle": "~2.4",
         "doctrine/doctrine-fixtures-bundle": "~2.2",
         "stof/doctrine-extensions-bundle": "~1.1",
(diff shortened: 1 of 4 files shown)


roderik:
Until there is a good solution for https://github.com/liip/LiipImagineBundle/issues/584, we cannot upgrade

joker806:
Oh, I hit this problem too but I considered it to be my fail because I allowed user to pick an PNG image where JPG was expected by the filter.

I'm closing this PR for now.

PR #850, opened 2015-11-30
[AdminBundle] Make sure folder is always a property

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

This fixes a bug introduced in this commit: https://github.com/Kunstmaan/KunstmaanBundlesCMS/commit/7e60d159e029b7055f07088e975fc2493e705370. When the folder is not explicitly set on the MenuItem, the property will not be present. The is the case on the Settings page in the admin, which causes an error.


jockri:
Thanks!

PR #853, opened 2015-12-10
[GeneratorBundle] Add FE styling to pageparts in CMS

dbeerten:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /


src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/edit.html.twig
```diff
@@ -70,3 +70,7 @@
     {% include 'KunstmaanNodeBundle:NodeAdmin:_modals.html.twig' %}
 
 {% endblock %}
+
+{% block extracss %}
+    <link rel="stylesheet" href="/frontend/css/admin-style.min.css">
```
jockri: We need to check if the file exists, because when you update a site without admin styles, the file will not exist. Also use asset() function.

Issue #857, opened 2015-12-16
Datepicker is not initialised on new date fields that have been created in nested forms


mtnorthrop:
When a new nested form is added to a form, the datepicker plugin is not initialised on date and datetime fields.

This is similar to the bug that was fixed in commit 53368907d8fa3817fd02a0a03d80840536a46741, but that only fixed the issue when new *pageparts* are added; not when a **nested form** is added inside a pagepart or admin list.

It seems that it should be enough to call `kunstmaanbundles.datepicker.reInitialise()` inside the `kunstmaanbundles.nestedForm.addNewItem()` method.

[Update] After posting this, I realised that this bug has already been mentioned in issue #823. I think we're both talking about the same thing.

Issue #823, opened 2015-11-13
Datetime Sub Entities Date/Time picker


alexborton:
It would appear that the date/time pickers are not instantiated on sub-entities as they are 'added' to the pagepart.

This means relying on the user to input the data in the correct format (always risky!)

PR #859, opened 2015-12-16
[AdminBundle] Added datepicker reinit for nested forms

mtnorthrop:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #857, #823

The datepicker plugin was not being initialized on nested forms that were dynamically added to a page. This fix calls the datepicker's reInit method to fix this.


jockri:
Thanks!

PR #873, opened 2016-01-02
[All] Allow Symfony 2.8 and fix some deprecations

WouterJ:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

 * If a bundle works on 2.7, it works on 2.8 too, no reason to not allow installing this bundle with Symfony 2.8.
 * Improve Travis setup to run tests quicker and test more
 * Removed some deprecated feature usages (the ones that don't require BC layers)

Is there any roadmap for Symfony 3 support? E.g. support both Symfony 2 and 3 with a BC layer like [the one in SonataCoreBundle](https://github.com/sonata-project/SonataCoreBundle/blob/master/Form/FormHelper.php) or do you only support 2.8 and 3.0 (which would be possible without BC layers), etc. ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 
 cache:
   directories:
-    - $HOME/.composer/cache
+    - $HOME/.composer/cache/files
 
 php:
   - 5.4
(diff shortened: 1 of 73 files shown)


jockri:
Hi @WouterJ 

The tests are failing because you added `symfony/phpunit-bridge` and this makes that deprecation warnings give errors in the tests.

Can your remove this dependency? The package will be added again when preparing and moving for SF3.0.

WouterJ:
Hi @jockri. Instead of removign the package, I've set the `SYMFONY_DEPRECATIONS_HELPER` env variable to 16 (current number of deprecations). This way, if there are 16 or less deprecations, tests will still pass. This means that you are aware when merging PRs introducing more deprecated usages.

I've also updated the composer.json files in the separate bundles and made sure they all used PSR-4 instead of PSR-0 + target-dir.

jockri:
Great solution. Thanks for the contribution!

Issue #623, opened 2015-08-10
Download pagepart: Can't choose media, except images & videos


dbeerten:
In the default download pagepart, it is not possible to select other files than images/videos (tested with .pdf and .xlsx)

![screen shot 2015-08-10 at 15 20 06](https://cloud.githubusercontent.com/assets/5577291/9172411/b6a995a2-3f73-11e5-8fc1-095a5a3b350c.png)

Proof that the files exist:
![screen shot 2015-08-10 at 15 23 37](https://cloud.githubusercontent.com/assets/5577291/9172430/d74b9a4e-3f73-11e5-828d-6eeadc95145a.png)

PR #881, opened 2016-01-12
[MediaBundle] Choose of any type, in pagepart : Download

whitewhidow:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/623


jockri:
Does this break the functionality that you can only choose images on the ImagePagePart etc?

whitewhidow:
@jockri yep it seems so, seems they both use the same code .

jockri:
The solution is probably that the FileHandler should be able the "handle" ppt filets etc

PR #901, opened 2016-01-15
[MediaBundle] Fix img url retina which broke when an svg is uploaded

FVKVN:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


src/Kunstmaan/AdminBundle/Resources/views/Form/fields.html.twig
```diff
@@ -48,15 +48,17 @@
 {% spaceless %}
     {% set type = type|default('text') %}
     {% set attr = attr|merge({ 'class': (attr.class|default('') ~ ' form-control' ~ (attr['maxlength'] is defined ? ' js-max-length-input' : ''))|trim }) %}
-    <input type="{{ type }}" {{ block('widget_attributes') }} {% if value is not empty %}value="{{ value }}"{% endif %}>
+    <input type="{{ type }}" {{ block('widget_attributes') }} {% if value is not empty %}value="{{ value }}"{% endif %}"{% endif %}>
```
diskwriter: Double endif ?

Issue #865, opened 2015-12-23
[AdminBundle] Add up/down arrows on nested forms


Devolicious:
Nested forms should also have the up/down arrows when you use the nested_sortable option. Maybe it's even better to remove the drag/drop functionality and only use up/down arrows since it has some odd behavior (dragging outside of the page part with nested forms).

Issue #216, opened 2015-03-19
Add up and down buttons to move pageparts, and move subentities in pageparts


krispypen:
We can add move up and move down buttons to the left of the delete button of each pagepart and each subentity. This is because when pageparts are too big, the drag-and-drop is almost unusable

wimvds:
An up/down button does not solve the issue imho, since the pagepart will be moved after a click on the button, you will have to scroll again and click on the button again if you want to move it more then 1 position up (or if auto scrolling was implemented it would be quite slow to reorder). I think a toggle button (per region?) to minify pageparts to 1 line which you can drag/drop would be both easier to implement and more user friendly. For pageparts containing sub entities you could add a similar button to minify the subentities in the containing pagepart for easier drag/drop reordering.

IbeVanmeenen:
A button to minify layout is in my opinion a good idea, not only for subentities, but also for the pageparts in general.

tentwofour:
+1 

A minify all within a region would be handy.

Up/down would temporarily fix drag/drop issues, even if it means scrolling back up. In that regard, a simple $(el).offset() call on the page part that was moved, and auto-scrolling the window to its new position might be feasible?

PR #908, opened 2016-01-19
[AdminBundle] Added up/down arrows on nested forms, fix rich editor when moving elements in DOM

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #865 #216

Nested forms can now also be sorted by using the up and down arrow.

Also a fix for the rich editor. When moving a element around in the DOM by sorting it, the ckeditor gets disabled and the content is removed. This is a known bug in Ckeditor. I have fixed it by destroying all the rich editors and reinitialise them when moving them around.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/ui/js/_pagepart-editor.js
+++ b/src/Kunstmaan/AdminBundle/Resources/ui/js/_pagepart-editor.js
@@ -8,7 +8,7 @@ kunstmaanbundles.pagepartEditor = function (window) {
         delete: []
     };
 
-    var init, addPagePart, editPagePart, deletePagePart, movePagePartUp, movePagePartDown, subscribeToEvent, unSubscribeToEvent, executeEvent;
+    var init, addPagePart, editPagePart, deletePagePart, movePagePartUp, movePagePartDown, subscribeToEvent, unSubscribeToEvent, executeEvent, reInit, updateDisplayOrder;
 
     init = function () {
         var $body = $('body');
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/AdminBundle/Resources/ui/js/_pagepart-editor.js
```diff
@@ -293,6 +305,28 @@ kunstmaanbundles.pagepartEditor = function (window) {
             cb({target: target})
         })
     };
+    reInit = function($el) {
+        $($el).find('*[data-reinit-js]').each(function() {
```
diskwriter: $el is already a jQuery element here

src/Kunstmaan/AdminBundle/Resources/ui/js/_pagepart-editor.js
```diff
@@ -184,6 +184,9 @@ kunstmaanbundles.pagepartEditor = function (window) {
 
         var currentPp = $('#' + targetId + '-pp-container');
```
diskwriter: Could you add the $var syntax for these variables?

src/Kunstmaan/AdminBundle/Resources/ui/js/_pagepart-editor.js
```diff
@@ -208,6 +214,9 @@ kunstmaanbundles.pagepartEditor = function (window) {
...
```
diskwriter: Could you add the $var syntax for these variables?

sandergo90:
@diskwriter good catch, done.

PR #925, opened 2016-01-20
[MenuBundle] Do not require online NodeTranslation for url_link MenuItems

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

### DIFF
--- a/src/Kunstmaan/MenuBundle/Repository/MenuItemRepository.php
+++ b/src/Kunstmaan/MenuBundle/Repository/MenuItemRepository.php
@@ -3,6 +3,7 @@
 namespace Kunstmaan\MenuBundle\Repository;
 
 use Gedmo\Tree\Entity\Repository\NestedTreeRepository;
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/MenuBundle/Repository/MenuItemRepository.php
```diff
@@ -25,7 +26,8 @@ public function getMenuItemsForLanguage($menuName, $locale)
             ->setParameter('locale', $locale)
             ->andWhere('m.name = :name')
             ->setParameter('name', $menuName)
-            ->andWhere('nt.online = 1');
+            ->andWhere('nt.online = 1 OR mi.type = :url_type')
+            ->setParameter('url_type', MenuItem::TYPE_URL_LINK);
```
jockri: It is better to use `BaseMenuItem::TYPE_URL_LINK` because the MenuItem entity can be overridden via configuration.

PR #933, opened 2016-01-22
[TaggingBundle] Lost old tags after adding new one to an entity

mdxpl:
The Tags collection need to be cloned before saveTagging
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


jockri:
Thanks

PR #945, opened 2016-01-26
[AllBundles]: use fully qualified class names for forms

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/PagePartBundle/Controller/PagePartAdminController.php
+++ b/src/Kunstmaan/PagePartBundle/Controller/PagePartAdminController.php
@@ -2,11 +2,13 @@
 
 namespace Kunstmaan\PagePartBundle\Controller;
 
+use Doctrine\Common\Util\ClassUtils;
 use Kunstmaan\PagePartBundle\Helper\PagePartConfigurationReader;
 use Kunstmaan\PagePartBundle\PagePartAdmin\PagePartAdmin;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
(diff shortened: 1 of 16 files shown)


src/Kunstmaan/PagePartBundle/Controller/PagePartAdminController.php
```diff
@@ -51,17 +53,16 @@ public function newPagePartAction(Request $request)
...
         $data                         = $formBuilder->getData();
         $data['pagepartadmin_' . $id] = $pagePart;
         $adminType                    = $pagePart->getDefaultAdminType();
-        if (!is_object($adminType) && is_string($adminType)) {
-            $adminType = $this->container->get($adminType);
-        }
-        $formBuilder->add('pagepartadmin_' . $id, $adminType);
+        $adminTypeFqn                 = ClassUtils::getClass($adminType);
+
+        $formBuilder->add('pagepartadmin_' . $id, $adminTypeFqn);
```
fchris82: Why? I use a form type service who needs a service:

```
    webtown_kunstmaan_chart.form.highchart_page_part.admin_type:
        class: Webtown\KunstmaanChartBundle\Form\PageParts\HighchartPagePartAdminType
        arguments:
            - '@webtown_kunstmaan_chart.configuration.parser_manager'  # <-- service!
```

I cannot set this service anywhere.

---

Otherwise I don't understand this solution and logic. There is the form type in the `$adminType` variable. Why is it wrong? Why get the FQN from the **existing** FormType then create it **again**? The `$formBuilder->add('pagepartadmin_' . $id, $adminType);` seems like a good solution for me.

wesleylancel:
@fchris82 I agree that the code is a little weird, but I think it'll instantiate the service anyway if it's tagged as a form type.

Issue #765, opened 2015-10-19
Social images should go through a imagine filter (jpg + optim)


roderik:

tentwofour:
Should they also likely be resized for FB purposes (max: 1200 x 630, as per https://developers.facebook.com/docs/sharing/best-practices#images)?

PR #962, opened 2016-02-01
[SeoBundle]: add imagine filter for seo images, refs #765

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #765


jockri:
Better to call the `getimagesize` function only once for each image.

PR #999, opened 2016-02-11
Adminlist viewaction

whitewhidow:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets |


src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
```diff
@@ -245,6 +246,39 @@ protected function doEditAction(AbstractAdminListConfigurator $configurator, $en
         );
     }
 
+
+    protected function doViewAction(AbstractAdminListConfigurator $configurator, $entityId, Request $request = null)
```
jockri: request should be required

src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
```diff
@@ -245,6 +246,39 @@ protected function doEditAction(AbstractAdminListConfigurator $configurator, $en
...
+    {
+        /* @var EntityManager $em */
+        $em = $this->getEntityManager();
+        if (is_null($request)) {
```
jockri: not needed if required

PR #1009, opened 2016-02-16
Fix FOSUserBundle extend

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

Admin login page doesn't seem to work on current dev-master with latest Symfony 2.8.x because of this extend.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Security/login.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Security/login.html.twig
@@ -1,4 +1,4 @@
-{% extends "@FOSUserBundle/Resources/views/layout.html.twig" %}
+{% extends "FOSUserBundle::layout.html.twig" %}
 
 {% block fos_user_content %}
     <form action="{{ path("fos_user_security_check") }}" method="post">


sandergo90:
Hi @wesleylancel 

We have tried to reproduce this problem but by using the latest version of the bundles we can not find a problem with the template. The admin login page is displaying correctly. 

The pull request you suggested will also be a step back because by following the best practices of symfony we have to add the "@" in the extend of a template.

wesleylancel:
Hi @sandergo90,

The fact that it's unreproducable might be because I'm using the development version of Symfony 2.8, because of a bug that is present in 2.8.2 related to FOSUser and the Doctrine alias thing (I know of the workaround, but it's fixed in the latest Symfony).

I'll switch back to 2.8.2 and implement the workaround for now, so I can continue without this, but it might be necessary to check if this extend still works when Symfony 2.8.3 is released, if it still does it's probably just a thing in the dev current dev version.

Regarding the syntax of the include, it's a copy straight from FOSUserBundle itsself. Symfony still refers to this syntax in their docs: http://symfony.com/doc/current/book/templating.html#referencing-templates-in-a-bundle http://symfony.com/doc/current/best_practices/templates.html#template-locations.

sandergo90:
Hi,

We will indeed have to check if the problem occurs in symfony 2.8.3 release.

The FOSUserBundle implemenation is indeed like you have done it, but when you look at the following page http://symfony.com/doc/current/cookbook/templating/namespaced_paths.html you will see that using namespaced paths is faster, therefore we have rewrited all our templates to use the @ syntax.

wesleylancel:
See #1022

PR #1038, opened 2016-02-23
Fix typo in SearchBundle (RenderContect -> RenderContext)

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    |  no
| Deprecations? | yes

Deprecated the wrong syntax getter and setter.

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Services/SearchService.php
+++ b/src/Kunstmaan/NodeSearchBundle/Services/SearchService.php
@@ -24,7 +24,7 @@ class SearchService
      * @var RenderContext
      *
      */
-    protected $renderContect;
+    protected $renderContext;
 
     /**
      * @var Container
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/NodeSearchBundle/Services/SearchService.php
```diff
@@ -67,17 +67,35 @@ public function setDefaultPerPage($defaultPerPage)
...
+    }
+
+    /**
+     * @return RenderContext
+     * @deprecated Use getRenderContext
+     */
     public function getRenderContect()
     {
-        return $this->renderContect;
+        return $this->renderContext;
```
sandergo90: Can you remove the old getters and setters?

jverdeyen: But then it's a BC.. we can deprecate them first and remove them in the nex major version.

sandergo90: @jockri what do you think?

jockri: Make it deprecated now, and we will remove all deprecations things before 4.0 gets released.

PR #1111, opened 2016-04-13
[MediaBundle] Added extensionless file upload

Caroga:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no

Users should be able to upload extension-less files in case when exporting a media file without extension.
As there is no visual error handling for this, we should just check for it inside the backend and allow it.

### DIFF
--- a/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
+++ b/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
@@ -167,9 +167,14 @@ class FileHandler extends AbstractMediaHandler
         }
         if ($content instanceof UploadedFile) {
             $pathInfo = pathinfo($content->getClientOriginalName());
-            $media->setOriginalFilename($this->slugifier->slugify($pathInfo['filename']).'.'.$pathInfo['extension']);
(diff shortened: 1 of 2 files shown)


jockri:
We discussed this internally, and think it is not best practice to upload extensionless files in the CMS. 

When you want to share things like this you can always use .txt or put the file in a zip folder.

Caroga:
In current situation you will get a 500 server error, which is not handled for. As for the users who would upload extension-less images: they presumable have no knowledge about the case, and are to be considered unknowingly about how to solve this. The only thing they will be served with is a perfectly preventable 500 server error page, which resolves them to run screaming in fear and awe.

As this is not a breaking change and adds a little bit of error handling and usability (as little as it is, it still does add some..) I really urge you to reconsider the pull request. 
Thank you very much.

jockri:
I agree that there should be a nice error displayed when someone uploads an extensionless file.
But the extensionless file should not arrive in the uploads folder or database.
If you want to add the error message, please open a new PR.
Thanks!

PR #1061, opened 2016-03-05
[WIP] [AdminBundle] Auto Google SignIn with Guard

Devolicious:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Thanks to the hosted_domain option with Google SignIn you can now automatically login as an Admin with your email domain. You'll have to configure the ```google_signin``` in your ```config.yml``` options like this:

```yml
kunstmaan_admin:
    google_signin:
        enabled: true
        client_id: some_client_id.apps.googleusercontent.com
        client_secret: some_secret
        hosted_domain: kunstmaan.be
```

After that you only have to configure the ```guard``` component in the ```security.yml``` like so:
```yml
security:
    ...
    firewalls:
        main:
            pattern: .*
            guard:
                authenticators:
                    - kunstmaan_admin.oauth_authenticator
            form_login:
                login_path: fos_user_security_login
                check_path: fos_user_security_check
                provider: fos_userbundle
            ...
```

And then the login will look like this:

screenshot 2016-03-05 13 45 12

PR #1119, opened 2016-04-20
[AdminBundle] Auto Google SignIn with Guard

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Finishes #1061 

Thanks to the hosted_domain option with Google SignIn you can now automatically login as an Admin with your email domain. You'll have to configure the google_signin in your config.yml options like this:

```
    kunstmaan_admin:
        google_signin:
           enabled: true
           client_id:  some_client_id.apps.googleusercontent
           client_secret: some_secret
           hosted_domains:
              - { domain_name: kunstmaan.be, access_levels: ['ROLE_SUPER_ADMIN'] }
              - { domain_name: mydomain.example, access_levels: ['ROLE_USER'] }
```

After that you only have to configure the guard component in the security.yml like so:

```
security:
    ...
    firewalls:
        main:
            pattern: .*
            guard:
                authenticators:
                    - kunstmaan_admin.oauth_authenticator
            form_login:
                login_path: fos_user_security_login
                check_path: fos_user_security_check
                provider: fos_userbundle
            ...
```

And then the login will look like this:

f1dc2d20-e2d8-11e5-8d6a-ecf8c62eb490


src/Kunstmaan/AdminBundle/Resources/views/Default/layout.html.twig
```diff
@@ -22,6 +22,8 @@
 
     <!-- JS -->
     {% include "@KunstmaanAdmin/Default/_js_header.html.twig" %}
+
+    <script src="https://apis.google.com/js/platform.js" async defer></script>
```
jockri: Enkel inladen wanner `{% if google_signin_enabled() %}` ?

diskwriter:
JS 👍

src/Kunstmaan/AdminBundle/Resources/config/services.yml
```diff
@@ -179,3 +179,23 @@ services:
...
+        class: Kunstmaan\AdminBundle\Security\OAuthAuthenticator
+        arguments:
+            - '@doctrine.orm.entity_manager'
+            - '@router'
+            - '@session'
+            - '@translator'
+            - '%kunstmaan_admin.google_signin.client_id%'
+            - '%kunstmaan_admin.google_signin.client_secret%'
+            - '%kunstmaan_admin.google_signin.hosted_domains%'
+            - '%fos_user.model.user.class%'
```
mlebkowski: think about splitting it into more services with fewer dependencies each.

Numkil:
@mlebkowski Good call, if have extracted the logic for creating the users in it's own helper class

Issue #1124, opened 2016-04-22
Undefined index: client_id


w3q:
Hi,

I found one problem and don't really know why it just stopped working.

I have my KunstmaanBundlesCMS working. I've been doing some customization, part of it was to composer install "jms/i18n-routing-bundle": "dev-master".

After composer update, I've started to get the error message:

<img width="1280" alt="snimka obrazovky 2016-04-22 o 11 51 47" src="https://cloud.githubusercontent.com/assets/1138670/14738203/a03541d2-0880-11e6-9f54-bcc5f0646625.png">

Now Im not able even to remove the bundle from composer, since composer remove will return the same error just in command line.

I have no idea why it did happened after installing the i18n-routing-bundle, cause it looks like it has something to do with OAuth for the admin?

Thank you a lot.

ronaldtb:
I just ran into the same problem. A workaround seems to be to add the `google_signin` configuration to your config.yml. This probleem seems to be related to #806 and #1119.

```yaml
kunstmaan_admin:
    admin_password: -adminpwd-
    dashboard_route: 'kunstmaan_dashboard'
    google_signin:
        enabled: false
        client_id: ''
        client_secret: ''
```

w3q:
Thank you a lot, Ill try that.

PR #1126, opened 2016-04-22
[AdminBundle] fixed undefined index client_id

slaci:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1124 

Parameters remain defined even in disabled status, because a service requires them...

### DIFF
--- a/src/Kunstmaan/AdminBundle/DependencyInjection/Configuration.php
+++ b/src/Kunstmaan/AdminBundle/DependencyInjection/Configuration.php
@@ -4,6 +4,7 @@ namespace Kunstmaan\AdminBundle\DependencyInjection;
 
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 
 /**
  * This is the class that validates and merges configuration from your app/config files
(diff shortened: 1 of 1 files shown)


Numkil:
+1 Smart fix, could you edit the cookbook section about google signin in the documentation to reflect the changes you made? Thanks!

fchris82:
+1

slaci:
Well, this only fixes the bug fast, but isn't smart. If you write `enabled: false` then it fails, because the isRequired() on the children (if you omit this then works fine ofc). The only way to fix this to check the requirements in the extension and throw exception (there is some configurationexception or something i think that fits here). when neccessary afaik.

slaci:
I updated the PR with the mentioned check. Now `enabled: false` works alone too.

About the [cookbook](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/docs/05-13-enabling-google-auth-login.md), i don't think its need any update cause this. All the same after this, just no fatal error by default. The only change is: now `google_signin: false` means the same as `enabled: false` cause the canBeEnabled() stuff, but thats very minor.

fchris82:
@slaci Maybe there is a better solution. Use the `beforeNormalization()` function, then throw exception if it is enabled but doesn't set the required values: http://php-and-symfony.matthiasnoback.nl/2011/11/symfony2-define-your-bundles-configuration-values-using-the-treebuilder/
See the `Before normalization - then what?` block.

slaci:
Ok now its 100% done! Refactored to 1 commit.

PR #1131, opened 2016-04-26
[UserManagementBundle] workaround for custom user entity class

ro0NL:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no 
| Deprecations? | no
| Fixed tickets | ~

Fixes

> The form's view data is expected to be an instance of class Kunstmaan\AdminBundle\Entity\User, but is an instance of class **Company\Entity\User**. You can avoid this error by setting the "data_class" option to null or by adding a view transformer that transforms an instance of class **Company\Entity\User** to an instance of Kunstmaan\AdminBundle\Entity\User.

at `/admin/settings/users/add` and  #`/admin/settings/users/xxx/edit`

as we configured `fos.user_class` in `config.yml`

### DIFF
--- a/src/Kunstmaan/UserManagementBundle/Controller/UsersController.php
+++ b/src/Kunstmaan/UserManagementBundle/Controller/UsersController.php
@@ -85,7 +85,7 @@ class UsersController extends BaseSettingsController
         $this->checkPermission();
         $user = $this->getUserClassInstance();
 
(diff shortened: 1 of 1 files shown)


ro0NL:
Btw, this still requires some custom ORM configuration, i.e. skipping the `Kunstmaan\AdminBundle\Entity\User` entity in `doctrine:schema:update` and such. Otherwise our custom entity conflicts (table already exists).

To clearify, we've added additional properties to our user class while we want to maintain the `kuma_users` table, we've altered the table now.

Issue #1141, opened 2016-04-28
[UserManagementBundle] User form rendered twice when using tabpanes


ro0NL:
Hi,

I've implemented a tabpane for the user form as described in #349. The form is rendered twice...

Snippet `KunstmaanUserManagement/Users/edit.html.twig`;
```
        <!-- Fields -->
        <fieldset class="form__fieldset--padded">
            {% if tabPane is defined %}
                {{ tabs_widget(tabPane) }}
            {% else %}
                {{ form_rest(form) }}
            {% endif %}
        </fieldset>
    {{ form_end(form) }}
```

First user form is rendered in a tab via `tabs_widget(tabPane)`. Second user form is rendered via `form_end`, i.e. the `FormView` instance is not marked as rendered by the first call. I guess this somehow happens due the fact we're using a `FormView` instance directly in `AdminBundle/TabsTwigExtension/widget.html.twig`... any thoughts?

PR #1142, opened 2016-04-28
[UserManagementBundle] User form rendered twice when using tabpanes

ro0NL:
| Q             | A
| ------------- | ---
| Bug fix?      | yes|
| New feature?  | no|
| BC breaks?    | no|
| Deprecations? | no|
| Fixed tickets | #1141|

Had to figure this one out :) the problem is we have 2 instances of `FormView`, one assigned by default to the template (https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/UserManagementBundle/Controller/UsersController.php#L197) the other one is generated from `TabPane::getFormView` when needed. Both track their own "am i rendered" state.

So we have to ignore one based on `tabPane` being set yes/no, as done with rendering the form itself;
https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/UserManagementBundle/Resources/views/Users/edit.html.twig#L49

Your welcome :) not sure this is also a low-level problem in e.g. the adminbundle itself... Also noticed the event is only triggered for *user* edit action, not add :( (ref #628)

### DIFF
--- a/src/Kunstmaan/UserManagementBundle/Resources/views/Users/edit.html.twig
+++ b/src/Kunstmaan/UserManagementBundle/Resources/views/Users/edit.html.twig
@@ -52,5 +52,10 @@
                 {{ form_rest(form) }}
             {% endif %}
         </fieldset>
-    {{ form_end(form) }}
+
+    {% if tabPane is defined %}
+        {{ form_end(tabPane.formView) }}
+    {% else %}
+        {{ form_end(form) }}
+    {% endif %}
 {% endblock %}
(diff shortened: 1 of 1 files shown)


ro0NL:
Now also fixes rendering of the form start tag :+1:

Our 2nd tab form contains a file upload. field The `enctype` attribute was missing due wrong form view instance, i.e. the one that symfony adds by default if your form contains a file upload field. It's merged into exsisting params, i.e. no BC break here.

Issue #1144, opened 2016-05-02
[AdminBundle] Google sign in not correct for existing users 


sandergo90:
I have activated the Google auth for a project. I already have an existing user for the e-mail address I would like to login with in my application.

In the OAuthUserCreator in the function "getOrCreateUser" there is a check to look for a user with username equal to the e-mail address provided. My username is not equal to my e-mail address. Therefor I think the check needs to be done on e-mail address and not on username.

Numkil:
@sandergo90 A fix has been merged that changes the default behaviour to something a bit more expected. Also included in this merge is an improvement to the code itself to allow users to override the default search for users with google login data functionality with their own or even to override the complete login code.

sandergo90:
Superb ;)

PR #1154, opened 2016-05-10
[AdminBundle] OAuthUserCreator Should query on username and email

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1144 

When someone signs in for the first time using google sign in, his user should be searched based on
email or username.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserCreatorInterface.php
+++ b/src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserCreatorInterface.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Kunstmaan\AdminBundle\Helper\Security\OAuth;
+
+/**
+ * Interface OAuthUserCreatorInterface
+ */
+interface OAuthUserCreatorInterface
+{
+    /**
+     * Returns an implementation of AbstractUser defined by the $userClass parameter.
+     * Checks if there already exists an account for the given googleId or email. If yes updates
+     * the access levels accordingly and returns that user. If no creates a new user with the
+     * configured access levels.
+     *
+     * Returns Null if email is not in configured domains
+     *
+     * @param string email
+     * @param string googleId
+     *
+     * @return mixed AbstractUser Implementation
+    */
+    public function getOrCreateUser($email, $googleId);
+}
--- a/src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserFinder.php
+++ b/src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserFinder.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Kunstmaan\AdminBundle\Helper\Security\OAuth;
+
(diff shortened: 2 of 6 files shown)


Numkil:
I just realised there could be still a problem with this code in the following use case. 
There are 2 users in the database. One with username == [the email we search for] and one with email == [the email we search for]. That way it would be impossible to decide which user should be logged in...

jockri:
Perhaps we should have logic that it first searches for matching email. If not found, also search username. What do you think?

Numkil:
I gave find by email a higher priority as find by username. That way if there are 2 accounts present with that email address the one with the email address in the email field will be chosen.

mlebkowski:
How about giving the developer a choice? I.e. extract some kind `UserFinderInterface`, create a default service implementing it with the above logic, but with the possibility for the developer to replace it with his own implementation (replacing a service in the container), hmm?

Numkil:
@mlebkowski following your advice I pulled everything apart and set everything behind interfaces. This should allow other people to override either only the code for finding a user based on the googleId and email address or the complete login procedure. 

@jockri I think this should be quite ready now

src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserCreatorInterface.php
```diff
@@ -0,0 +1,24 @@
...
+     * Checks if there already exists an account for the given googleId or email. If yes updates
+     * the access levels accordingly and returns that user. If no creates a new user with the
+     * configured access levels.
+     *
+     * Returns Null if email is not in configured domains
+     *
+     * @param string email
+     * @param string googleId
+     *
+     * @return AbstractUser Implementation
```
mlebkowski: `@return mixed AbstractUser Implementation` — this way IDE won’t expect to return `AbstractUser` instance. Maybe it should be a `UserInterface` instance from FOS?

src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserFinder.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /** @var string */
+    private $userClass;
+
+    /**
+     * OAuthUserCreator constructor.
+     * @param EntityManager $em
+     * @param $userClass
+     */
+    public function __construct(EntityManager $em, $userClass)
```
mlebkowski: `EntityManagerInterface`

src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserFinderInterface.php
```diff
@@ -0,0 +1,22 @@
...
+{
+
+    /**
+     * Tries to find a user in database based on email and googleId fields.
+     * Returns null when nothing has been found.
+     *
+     * @param string email
+     * @param string googleId
+     *
+     * @return AbstractUser Implementation
```
mlebkowski: same here, `mixed`

PR #1155, opened 2016-05-12
[AdminBundle]: admin token can also be PostAuthenticationGuardToken

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When logging in with the google sign in, we do not have a UsernamePassword token anymore. It will be a PostAuthenticationGuardToken. Therefore the admin locale you have chosen does not work anymore. When checking the admin token we should also check if the user's token is a PostAuthenticationGuardToken.


jockri:
can you do this for master branch? google sign in is also in master

Issue #1159, opened 2016-05-13
[NodeBundle] link select language problem


slaci:
The url chooser works only with multilanguage sites now: https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/NodeBundle/Resources/views/Widgets/selectLink.html.twig#L83

In this file: `data-base-url={{ "/" ~ app.request.locale }}` is the problematic piece, cause locale is not needed in the url on single language settings.

The fix should be simple, just `{{ multilanguage ? "/" ~ app.request.locale : '' }}`, but that multilanguage parameter is not accessible in twig (and cant access the container either).

The other parameters are defined in [config.yml](https://github.com/Kunstmaan/KunstmaanBundlesStandardEdition/blob/master/app/config/config.yml#L17), which is in a different package, and putting the `multilanguage: %multilanguage%` line there would be very easy but also a bc break.

So, what do you think, is it okey to fix this way or have a better idea?

---
Other approach would be a prod url generator, which just replaces the scriptname from the generated url (this is big hax and not recommended). A twig extension in NodeTwigExtension would be for this (with needs_context => true):

```php
public function getProdPathForNodeUrl($context, $urlOrNode, $locale = null, $routeParams = array(), $relative = false, $prodEnv = 'prod')
    {
        $request = $this->requestStack->getMasterRequest();

        if ($urlOrNode instanceof Node) {
            $urlOrNode->getNodeTranslation($locale);
        }

        if ($urlOrNode instanceof NodeTranslation) {
            $locale = $locale ?: $urlOrNode->getLang();
            $urlOrNode = $urlOrNode->getUrl();
        }

        $routeParams = array_merge($routeParams, array(
            'url'       => $urlOrNode,
            '_locale'   => $locale ?: $request->attributes->get('_locale'),
        ));

        $path = $this->generator->generate(
            '_slug',
            $routeParams,
            $relative ? UrlGeneratorInterface::RELATIVE_PATH : UrlGeneratorInterface::ABSOLUTE_PATH
        );

        if ($context['app']->getEnvironment() !== $prodEnv) {
            $path = str_replace($request->getScriptName(), '', $path);
        }

        return $path;
    }
```

jockri:
I think we can even do this for getting the base url:

```
{{ path('_slug', {'url': ''}) }}
```

slaci:
[...]

PR #1167, opened 2016-05-23
[NodeBundle] link select language problem

slaci:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1159 

Uses the "multilanguage" twig global variable to control the base url. This is not defined by default (and consistently can only be defined in the standard edition), so when the variable is missing, then the old behavior applies (generates locale aware base urls, like multilanguage == true).

Generating urls with routing would be much better, but its not that simple.

The multilanguage variable should come from this: https://github.com/Kunstmaan/KunstmaanBundlesStandardEdition/pull/208

### DIFF
--- a/src/Kunstmaan/NodeBundle/Resources/views/Widgets/selectLink.html.twig
+++ b/src/Kunstmaan/NodeBundle/Resources/views/Widgets/selectLink.html.twig
@@ -80,7 +80,8 @@
                 <hr>
             </div>
 
-            <nav role="navigation" id="app__sidebar__navigation" class="app__sidebar__module app__sidebar__navigation" data-reorder-url="{{ path('KunstmaanNodeBundle_nodes_reorder') }}" data-base-url={{ "/" ~ app.request.locale }}>
+            {% set base_url = multilanguage is not defined or multilanguage ? '/' ~ app.request.locale : '' %}
+            <nav role="navigation" id="app__sidebar__navigation" class="app__sidebar__module app__sidebar__navigation" data-reorder-url="{{ path('KunstmaanNodeBundle_nodes_reorder') }}" data-base-url={{ base_url }}>
                 <ul>
                     {% for item in tree.rootItems %}
                         {{ macros.selectLinkRecTreeView(tree, item) }}


mlebkowski:
> Generating urls with routing would be much better, but its not that simple.

Why? How about `path("_slug", { "_locale": lang, "url": "REPLACE_ME" })` and then use this as a placeholder instead of a prefix?

slaci:
Because if some1 selects a link in dev mode, then the link will contain app_dev.php, and that will be saved into the db. Its ofc not recommended to upload content in dev mode for prod usage, but this is not unlikely to happen I think...
There is no easy way to generate prod url in symfony from a different env afaik... only if you dualboot somehow a prod appkernel maybe... seems very big overhead, but the cleanest.
The other way is the hax, when just use str_replace and dontcare.. but I don't want to push something like that (I pasted an example into the linked issue for this) :P

mlebkowski:
wow, I didn’t realize that symfony would add `/app_dev.php` prefix if the rewrite is not configured.

on a side note, storing urls in the db is fragile either way — they won’t update when the `NodeTranslation` changes. This is why we created a [url to NodeTranslation transformer](https://github.com/syzygypl/kunstmaan-extra-bundle/blob/master/src/Form/Extension/DataTransformer/NodeTranslationToUrlTransformer.php) and used it to extend the behaviour of [`UrlChooser`](https://github.com/syzygypl/kunstmaan-extra-bundle/blob/master/src/Form/PageUrlType.php). This way the interface stays the same, but the database stores a reference to NodeTranslation instead of a string with url.

jockri:
I would use `path("_slug", { "_locale": lang, "url": "" })`. Then no replace is needed.

A couple of weeks ago, we moved the url generating outside of the recursive loop because it was becoming a bottleneck. In the previous bundle versions, the app_dev.php was also added in dev mode.

I agree that it would be better to store a reference in the database, and generate te link on the fly when needed. This was something we have discussed internally and it should be added to the roadmap.

mlebkowski:
> I would use `path("_slug", { "_locale": lang, "url": "" })`. Then no replace is needed.

You assume that the `_slug` route is like `/{_locale}/{url}`, but in fact it can be changed so that the `url` is not a simple prefix. This is why I suggested a "REPLACE_ME" approach.

mlebkowski:
And — yes — your solution would work out of the box now w/o any more changes needed, IIRC

jockri:
closed in favor of https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/1223

PR #1168, opened 2016-05-24
[AdminBundle] Command to update all nodes with given permission(s) for a desired role.

bijsterdee:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

**Description**:
With this command you can easily set permissions for a role on every existing node.

**Scenario 1 (new role)**:
After adding a new role in Kunstmaan, the role has no permissions on existing nodes.
The user wants to add "view" and "edit" permission for that user for every node.

**Scenario 2 (existing role)**:
When a role needs to have "publish" and "unpublish" permissions for every existing node.


Caroga:
+1!

PR #1169, opened 2016-05-24
[UserManagementBundle] Added event listeners for user management

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

Also added an check which checks of the user is found


jockri:
What is the use case for this events? Can you also do this with the Doctrine events? preRemove, preUpdate, ...

dannyvw:
We have different super admins but we don't want that other customers can edit or delete our super admin. It's a special use-case for us but maybe can someone else use this listeners too (with there own logic).

PR #1170, opened 2016-05-24
[AdminBundle] Remove ROLE_SUPER_ADMIN role for permissions when user is not an superadmin

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Remove ROLE_SUPER_ADMIN role for permissions when user is not an superadmin


jockri:
I don't think it is a good idea to change the `getAllRoles` function for this. Some other projects might depend on this function and is expecting all roles. Better to create separate function for your use case it think eg `getManagableRolesForPages`.

dannyvw:
Is this something you want in the core? If so.. then we can add a separate method like getManagableRolesForPages and edit the view to use this method.

jockri:
Yes, that can be added in the core. Just create a new helper function.

dannyvw:
@jockri Pull-request is updated

PR #1171, opened 2016-05-25
[NodeBundle] validate if node is not null

bakie:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

Validate if the node is found, otherwise we get a 500 error


mlebkowski:
This adds boilerplate to this controller, and it’s not one of the prettiest to begin with. How about reducing it instead by using [ParamConverters](http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/converters.html), see e16455ce4d2f2e975084f913e891f683c574d06b

bakie:
@mlebkowski that is indeed a much cleaner solution, but that does make it a BC break :). I will close this one.

PR #1180, opened 2016-05-31
[AdminBundle] Updated translations

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Resetting/request_content.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Resetting/request_content.html.twig
@@ -26,7 +26,7 @@
         {{ 'Reset password' |trans }}
     </button>
     <a href="{{ path('fos_user_security_login' ) }}" class="btn btn-default btn--raise-on-hover">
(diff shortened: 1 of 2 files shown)


jockri:
Why is this a bug? We prefer key based translation strings.

dannyvw:
It renders "form.cancel" instead of "Cancel"

jockri:
Then there is a translations missage missing. Can you check this?

dannyvw:
The AdminBundle has as parent FOSUserBundle and there is no translation for "form.cancel"

jockri:
You can pass an other domain in the template so it uses the messages domain.

dannyvw:
True, but KunstmaanAdminBundle is not a valid domain, so which domain you want to pass?

jockri:
You can use 'messages' as domain

dannyvw:
It is no longer a problem in master. Dutch translation was not available in 3.5.1

PR #1182, opened 2016-06-01
[AdminBundle] fix regex to check if admin preview

bakie:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

I extended the admin/preview check. The check to see if the url contains /admin/preview was not working if you have a preview route like /admin/project/preview


mlebkowski:
And what if you have a different route altogether? This seems like an ugly hack.

Wouldn’t it be better to mark admin routes with a request parameter for an instance?

PR #1187, opened 2016-06-03
[NodeBundle]: add custom form type for node translations

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

I have added a new form type to use instead of the EntityType when working with node translations.

When you want to have a entity fields that has a relationship to a node and you want to create a form for this entities relationship, you would normally use the EntityType. By using the NodeTranslationChoiceType, the title used as choice_label, will be automatically the title in the current language. You can also provide the "locale" parameter to set a default locale.

Example:

`   ->add('question', NodeTranslationChoiceType::class,
                array(
                    'page_class' => array(StartersVraagPage::class),
                )
            )`

### DIFF
--- a/src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php
+++ b/src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Kunstmaan\NodeBundle\Form;
+
+use Kunstmaan\NodeBundle\Repository\NodeRepository;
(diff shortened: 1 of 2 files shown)


jockri:
You are saving the node id as reference? Then it is better to call this NodeChoiceType.
Can you update the PR?

sandergo90:
Changed ;)

PR #1188, opened 2016-06-06
[NodeSearchBundle] Change title search + add configuration

fchris82:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

While we search with "állam" (= state) then get back the title "vállalat" ( = company), because find "álla" in the "vállalat". This is wrong, and we don't want to allow it in some project.


Numkil:
This is a good bugfix for now. 

However I want to mention that in the next release of our cms we have a big refactoring of the whole Search and NodeSearch bundles coming up. Included in these refactorings we are moving away from using custom analyzers with basic Ngrams to stem words and are going to use the build in language specific analyzers from elasticsearch itself. These will be more intelligent and weird matches like that should not happen again.

PR #1196, opened 2016-06-14
Prevent exception in node editor if search provider is unavailable

whitezo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Page save won't throw an exception if search provider (elasticsearch) is not running/available at the moment but saves page without indexing it.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/translations/messages.en.yml
+++ b/src/Kunstmaan/AdminBundle/Resources/translations/messages.en.yml
@@ -104,6 +104,7 @@ kuma_admin:
     edit:
         flash:
             success: The welcome page has been edited!
+            search_provider_offline: Search indexing unavailable!
     password_check:
         flash:
             error: Your password has not yet been changed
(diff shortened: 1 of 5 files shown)


mlebkowski:
I don’t think that silently discarding an error when indexing is a good solution. If you only wish to discard this error when saving node and you’re comfortable with it for some reason (I wouldn’t be):

 * Override just the search index listener, not the indexer itself. This way the `kuma:search:populate` command will work and throw an error (as expected) if there is a connection problem
 * Catch an error there and convert it to a flash message. I’d suggest to narrow down the type of the exception. Currently all of them are converted to `SearchProviderException`, so you’re not only catching connection problems, but every single thing that can go wrong during indexing, even permanent situations.
 * When converting exceptions it is a good practice to pass the original exception as a third argument to the new one (you won’t lose context this way)
 * Silently discarding an exception and then checking the service status may very well produce both false positives (the node won’t be indexed, but then the service will *just* recover and the status will report everything is ok) and false negatives (the node will be indexed but the service will go down before the status can be checked)

Anyway, the data integrity is important to me. I’d like to have an error if it cannot be enforced because of a service downtime. I wouldn’t be able to do that with this patch merged. OTOH, by catching exceptions in the search indexer listener, you have the ability to decide one way or another, according to your project needs. I think this shouldn’t be merged to the core.

jockri:
I agree with @mlebkowski 
Closing this.

Issue #1183, opened 2016-06-02
[sf3] Image upload crashes


aistis-:
Image upload crashes if filename does not contain extension. For example file `my-ribs.jpg` will be uploaded successfully, but `my-ribsjpg` will cause an exception:

```
[2016-06-02 12:18:34] request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "Notice: Undefined index: extension" at /var/www/html/vendor/kunstmaan/bundles-cms/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php line 175 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\ContextErrorException(code: 0): Notice: Undefined index: extension at /var/www/html/vendor/kunstmaan/bundles-cms/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php:175)"} {"user":{"username":"admin","roles":["ROLE_SUPER_ADMIN","ROLE_PERMISSIONMANAGER","ROLE_ADMIN","IS_AUTHENTICATED_ANONYMOUSLY","ROLE_USER"],"is_account_non_expired":true,"is_account_non_locked":true,"is_credentials_non_expired":true,"is_enabled":true},"file":"/var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/EventListener/ExceptionListener.php","line":89,"class":"Symfony\\Component\\HttpKernel\\EventListener\\ExceptionListener","function":"logException"}
```

The error occurs on `sf3` branch.

jockri:
Yes, it is not allowed to upload extensionless files. However, the end user should see a nice error message instead of a 500 error.

Numkil:
@aistis- 
I can confirm, I tested it myself with a clean project and I am getting the same error messages. I will create a bugfix shortly. Thank you for reporting it.

sandergo90:
@jockri close

aistis-:
Thanks.

PR #1208, opened 2016-06-20
[MediaBundle] Auto guess Extension when missing from file or throw validation error if not possible to guess.

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1183 

* [x] Implement extensionguesser in filehandler logic
* [x] User friendly error message when impossible to guess extension/mimetype
* [x] Fix tests

Allow files to be uploaded without an extension in the filename. Detect it based on Symfony's MimeTypeGuesser and ExtensionGuesser.

I was going to write a simple check check to avoid the internal server error first but I figured this small "feature" is pretty cool.

### DIFF
--- a/src/Kunstmaan/MediaBundle/Form/File/FileType.php
+++ b/src/Kunstmaan/MediaBundle/Form/File/FileType.php
@@ -3,6 +3,7 @@
 namespace Kunstmaan\MediaBundle\Form\File;
 
 use Kunstmaan\MediaBundle\Repository\FolderRepository;
+use Kunstmaan\MediaBundle\Validator\Constraints\HasGuessableExtension;
 use Symfony\Bridge\Doctrine\Form\Type\EntityType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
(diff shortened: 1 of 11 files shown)


Caroga:
Just gonna leave this here.
https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1111

Numkil:
@Caroga crap, I should be more thorough and read closed PR's before I get my hands dirty. Thanks for letting me know.

mlebkowski:
Guys, I’m so disappointed. I just updated Bundles from `3.6.0` to `3.6.1` just to discover that there is a breaking change. Adding a required constructor dependency **is a breaking change**. If you don’t consider it a major breaking change, just try not to smuggle it with a patch release. 

Please pay more attention to this. I have a „legacy” project on version 3.3 AFAIR and I’m afraid to upgrade it because I’m scared what I can encounter, **even after reading the changelog and monitoring the development here**.

Devolicious:
@mlebkowski, we're really sorry about this. It was definitely not our intention to "smuggle" it into this patch release. It has slipped through the net, nothing more I can say about that.
On the bright side, we have been busy to improve our release process as you might have noticed. 

With this **new release process** we are trying to release more often, be more strict on PR's and carefully review them to avoid this kind of obstacles and **improve** the **quality** even more of our KunstmaanCMSBundles.

PR #1215, opened 2016-06-22
[MediaBundle] Rename soft deleted media

ro0NL:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes?
| Deprecations? | no

This renames a physical local file if the associated media entity is soft-deleted. I dont think its OK to keep the physical file publicly available for soft-deleted media **under the same URL**.

It restores the orginal filename when you _readd_ a soft deleted media entity, ie;
`$media->setDeleted(false); $em->persist($media);`

Any thoughts?

image

### DIFF
--- a/src/Kunstmaan/MediaBundle/Command/MigrateSoftDeletedCommand.php
+++ b/src/Kunstmaan/MediaBundle/Command/MigrateSoftDeletedCommand.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace Kunstmaan\MediaBundle\Command;
+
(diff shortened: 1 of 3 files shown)


ro0NL:
I guess it's a feature.. but you can keep re-using this command to generate new URL's whenever needed.

ro0NL:
The original filename now can also be restored via `kuma:media:rename-soft-deleted --original` :+1:

ro0NL:
Here for future reference. I dont have access to this repo anymore, so i cant work on it anyway.

PR #1225, opened 2016-07-01
[AllBundles] URLChooser major changes

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |

## URL Chooser

The URLChooserType had been rewritten.

You now have the option to add the following type of links:

- Internal
- External (external links or for example anchors)
- Email (mailto: will be automatically added)

We have added this rewrite especially for the MultiDomainBundle but it's completely integrated in the KunstmaanBundlesCMS.

When you choose for an internal link type, you will need to choose a link in the menu tree. On a multidomain website, you can switch you're domain on the right site of the URL chooser.

When clicking a node in the left node tree, a token is generated and saved for the URL. 

##For example: in a single domain website##

Selecting the node in the tree with node translation id 50, will generate a token like "[NT50]".

##For example: in a multi domain website##

Selecting the node in the tree with node translation id 50, on domain "my_domain_en" will generate a token like "[my_domain_en:NT50]".

##MAJOR CHANGE##

For the new URL chooser to work, we have added a new twig filter "replace_url". This filter needs to be used to generate the correct url automatically.

When creating a new pagePart with an url field, the twig filter is automatically being added. 

On all the exiting pageParts with an URLChooser field, you need to change the twig files of the pagePart and pipe all URL's to the replace_url filter.

When adding a rich text field or a wysiwyg field, you will need to pipe the output of this field also to the replace_url filter. Hereby, all URL's chosen in the wysiwyg will also be replaced correctly.


mlebkowski:
You’re modifying an interface. It’s as clear as it gets a breaking change.

jockri:
Can you rebase?

sandergo90:
@jockri done

src/Kunstmaan/AdminBundle/Resources/config/services.yml
```diff
@@ -57,6 +56,12 @@ services:
         tags:
             -  { name: 'twig.extension' }
 
+    kunstmaan_admin.multidomain.twig.extension:
```
jockri: Better to place this in MultiDomainBundle?

src/Kunstmaan/AdminBundle/Resources/views/MultiDomainAdminTwigExtension/widget.html.twig
```diff
@@ -0,0 +1,17 @@
+{% set switchedHost = get_switched_host() %}
```
jockri: Better to place this in MultiDomainBundle?

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
```diff
@@ -0,0 +1,102 @@
+<?php
```
jockri: Better to place this in MultiDomainBundle?

src/Kunstmaan/AdminBundle/Resources/config/services.yml
sandergo90: We can not put this in the MultiDomainBundle because the twig function needs to be available also when the MultiDomainBundle is not enabled.

src/Kunstmaan/AdminBundle/Resources/views/MultiDomainAdminTwigExtension/widget.html.twig
sandergo90: We can not put this in the MultiDomainBundle because the twig function needs to be available also when the MultiDomainBundle is not enabled.

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
sandergo90: I think the name should stay this way. It's an extension for the AdminBundle with functionality for the multidomain.

Issue #1226, opened 2016-07-04
No clear error message when slug is invalid


jockri:
When you modify the slug of a page manually, and you enter invalid input (eg add spaces), you get a general message that you have form errors, but no specific error on the slug field.

sandergo90:
@jockri close

PR #1245, opened 2016-07-23
[NodeBundle] fix bug error label not correct slugchooser

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1226 

This fixes the issue where when trying to update the slug of a page the error messages would not be displayed correctly. The only problem for me is that I do not understand why this fix works and the original code does not. It should give the exact same results? Maybe a bug in the Assert/Regex code?


src/Kunstmaan/NodeBundle/Entity/NodeTranslation.php
```diff
@@ -56,7 +56,6 @@ class NodeTranslation extends AbstractEntity
      * @var string
      *
      * @ORM\Column(type="string", nullable=true)
-     * @Assert\Regex("/^[a-zA-Z0-9\-_\/]+$/")
```
Devolicious: Laten staan :)

PR #1252, opened 2016-08-02
[SearchBundle] Allow easier switching between language and ngram word analyzers.

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 

Minor BC break: method name in analysisfactory interface changed. addnGramFilter has been changed to addnGramTokenizer because nGrams are supposed to be used as tokenizers and not as a filter. This change actually reflects how elasticsearch is implemented.


sandergo90:
Looking good! @krispypen can you merge?

PR #1256, opened 2016-08-04
[SearchBundle] check if index exists before deleting

bakie:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

Check if the index exists before trying to delete it, otherwise it throws an exception.


Devolicious:
👍

PR #1257, opened 2016-08-08
[FixtureBundle] Check allowed_children and getPossibleChildTypes in fixture generator

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no (TBD)
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no

This will make fixtures more _production ready_ :)

It's a BC break because projects with _wrong_ fixtures will have to adjust their code.

### DIFF
--- a/src/Kunstmaan/FixturesBundle/Builder/PageBuilder.php
+++ b/src/Kunstmaan/FixturesBundle/Builder/PageBuilder.php
@@ -10,10 +10,11 @@ use Kunstmaan\NodeBundle\Entity\Node;
 use Kunstmaan\NodeBundle\Entity\NodeTranslation;
 use Kunstmaan\NodeBundle\Entity\NodeVersion;
 use Kunstmaan\NodeBundle\Entity\StructureNode;
+use Kunstmaan\NodeBundle\Helper\PagesConfiguration;
 use Kunstmaan\NodeBundle\Helper\Services\ACLPermissionCreatorService;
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/FixturesBundle/Builder/PageBuilder.php
```diff
@@ -229,4 +248,25 @@ private static function incrementString($string, $append = '-v')
...
+     * @param PageInterface $parentPage
+     * @param string $childPageClass
+     * @return bool
+     */
+    private function canHaveChild(PageInterface $parentPage, $childPageClass)
+    {
+        $list = array_merge(
+            $this->pagesConfiguration->getPossibleChildTypes(get_class($parentPage)),
+            $parentPage->getPossibleChildTypes()
+        );
```
mlebkowski: `PagesConfiguration` will call `Page::getPossibleChildTypes` if nothing is set in the configuration. You cannot add/override using this method if a config is present (this is the behaviour when adding new pages via Admin Panel). This way you could skip the `$parentPageMock` altogether.

jverdeyen: Good catch! Thx!

jverdeyen:
Fixed!

Devolicious:
@jockri 👍

Issue #1175, opened 2016-05-26
[ConfigBundle] Admin type as fully qualified class name raises exception


aistis-:
`ServiceNotFoundException` is thrown when using fully qualified class name in Config entity:

```
    public function getDefaultAdminType()
    {
        return MyCustomConfigType::class; // extens AbstractConfig
    }
```

```
// ConfigController.php

// If the formType is a service, get it from the container.
if (!is_object($formType) && is_string($formType)) {
    $formType = $this->container->get($formType);
}

$formFqn = get_class($formType);

$form = $this->formFactory->create(
    $formFqn,
    $config
);
```

jockri:
I think we need to add a check that both configurations are possible (instance and classname). Can you create a PR to fix this? Thanks!

delboy1978uk:
I guess this can be closed?

PR #1262, opened 2016-08-10
[NodeBundle][AdminListBundle][PagePartBundle] Check if admin type is fqcn

dannyvw:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | #1175 |

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
+++ b/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
@@ -121,7 +121,11 @@ abstract class AdminListController extends Controller
 
         $formType = $configurator->getAdminType($helper);
         if (!is_object($formType) && is_string($formType)) {
-            $formType = $this->container->get($formType);
+            if (class_exists($formType)) {
+                $formType = new $formType;
+            } else {
+                $formType = $this->container->get($formType);
+            }
         }
         $formFqn = get_class($formType);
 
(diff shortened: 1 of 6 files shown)


mlebkowski:
Form Type could have its dependencies, so `new $formType` can fail with a fatal error.

dannyvw:
True, but all form types are changed because passing parameters is deprecated. See also PR https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/950

mlebkowski:
@dannyvw, I don’t see how’s that related. I’m just staying that instead of:

```
$formType; /// class name
$form = new $formType;
/// …
$fqcn = get_class($form);
```

you could simply:

```
$formType; /// class name
/// ...
$fqcn = $formType;
```

dannyvw:
@mlebkowski, I think that i know what you mean, but $formType can have different values like: 
```
// FQCN
AdminFormType::class
// Object
new AdminFormType() 
// service
'service.admin_form_type'
```
and yes ```$formType = new $formType;``` is then unnecessary because it's already an FQCN.

We can also do something like this
```
        $formType = $configurator->getAdminType($helper);
        $formFqn = is_object($formType) ? get_class($formType) : $formType;
        if (!is_object($formType) && is_string($formType)) {
            if ($this->container->has($formType)) {
                $formFqn = get_class($this->container->get($formType));
            }
        }
```

mlebkowski:
Yes, that’s what I was talking about.

dannyvw:
I have updated the pr's

aistis-:
Looks like there is some duplicated code. Could you extract it to one place and reuse it?

dannyvw:
@aistis- I added a helper for getting the fqcn and merged this branch into the sf3 branch, see #1263

aistis-:
I think it is a good fix for 3.x Kunstmaan version, but I suggest to drop BC for the old strategy on 4.x as it is for Symfony 3+ version only.

Devolicious:
This was covered in #1348 so I'll close this.

PR #1263, opened 2016-08-10
[ConfigBundle] Check if admin type is fqcn

dannyvw:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | #1175 |

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
+++ b/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
@@ -120,16 +120,13 @@ abstract class AdminListController extends Controller
         $helper = $configurator->decorateNewEntity($helper);
 
         $formType = $configurator->getAdminType($helper);
-        if (!is_object($formType) && is_string($formType)) {
-            $formType = $this->container->get($formType);
-        }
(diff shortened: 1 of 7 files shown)


aistis-:
Need this fix! :)

relates to #1262

aistis-:
@dannyvw @jockri What if refactor all forms to use fqcn instead of keeping BC when Type object is passed and then trying to compute fqcn by `get_class($typeObject)`? This branch is for Symfony 3 and it dropped a BC for deprecated strategy. I suggest do the same, as the code will come cleaner, more convenient. This legacy workaround causes some issue for the form with injected parameters (see https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1310) via constructor instead of using *form options*.

What do you think guys?

dannyvw:
@aistis- +1 for dropping BC for the old strategy. But before we can drop this, all implementations must be changed.

aistis-:
Yes, indeed :)

Devolicious:
This was covered in #1348 so I'll close this.

PR #1270, opened 2016-08-18
[AdminBundle] Added post-install script to setup single lang site

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 1086

[AdminBundle] Added post-install script to setup configs for single language file
[GeneratorBundle] Use multilanguage parameter to detect multi/single language, when generating default-site

### DIFF
--- a/src/Kunstmaan/AdminBundle/Composer/ScriptHandler.php
+++ b/src/Kunstmaan/AdminBundle/Composer/ScriptHandler.php
@@ -1,0 +1,84 @@
+<?php
+namespace Kunstmaan\AdminBundle\Composer;
+
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Yaml\Parser;
+use Composer\Script\Event;
+
+class ScriptHandler
+{
+    protected static $options = array(
+        'symfony-app-dir' => 'app',
+        'multi-language-option' => 'multilanguage'
+    );
+
+    /**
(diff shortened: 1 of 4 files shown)


mlebkowski:
I don’t think that this `post-install-cmd will` fire, [look at this node](https://getcomposer.org/doc/articles/scripts.md#what-is-a-script-)

treeleaf:
@mlebkowski I have removed it from the AdminBundle composer.json and added it in the Kunstmaan/KunstmaanBundlesStandardEdition in the composer.json

denbatte:
I would prefer it the other way around and have multi language as a default and use 'singlelanguage' as a flag.

treeleaf:
@denbatte multilanguage is still the default, but if you choose to use singlelanguage, then this script will replace the routing and security configs as in https://bundles.kunstmaan.be/documentation/cookbook/single-language-websites

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 
 ## 4.0.0-RC1 / 2016-06-21
 
+* [GeneratorBundle] Use multilanguage parameter to detect multi/single language site, when generating default-site [#1086](https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/1086) ([@treeleaf](https://github.com/treeleaf))
```
jockri: Can you remove the changes form the CHANGELOG, it will be generated automatically

Issue #1296, opened 2016-09-15
Linkchooser site switcher issue


Devolicious:
When you are on a multidomain site, switched from site and then try to add a link with the link chooser it fails. Outputs an illegal offset because the `selectLink.html.twig` implements the following on line 74:

```
{{ localeswitcher_widget(get_backend_locales(is_multidomain_site() ? get_switched_host()), route) }}
```

The `get_backend_locales(...)`is causing the issue. The `get_switched_host()` is returning an array whereas the `get_backend_locales(...)` is expecting a string (hostname)

PR #1302, opened 2016-09-18
[AllBundles] Fixed switching domain

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #1296

### DIFF
--- a/src/Kunstmaan/AdminBundle/Helper/DomainConfiguration.php
+++ b/src/Kunstmaan/AdminBundle/Helper/DomainConfiguration.php
@@ -26,6 +26,9 @@ class DomainConfiguration implements DomainConfigurationInterface
     /** @var string */
     protected $defaultLocale;
 
+    /**
+     * @param ContainerInterface $container
+     */
     public function __construct(ContainerInterface $container)
     {
         $this->container = $container;
@@ -69,6 +72,8 @@ class DomainConfiguration implements DomainConfigurationInterface
     }
 
     /**
+     * @param string|null $host
+     *
      * @return bool
      */
     public function isMultiLanguage($host = null)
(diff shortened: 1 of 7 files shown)


mlebkowski:
Guys, please. Changing an interface is a BC break.

sandergo90:
@dannyvw Our application breaks with your pull request. In the selectLink.html.twig template you have removed the switched_host variable.

`{% if is_multidomain_site() %}
        {% if switched_host_is_current() %}
            {% set slug = "[%s]" | format("NT" ~ item.nt_id) %}
        {% else %}
            {% set slug = "[%s:%s]" | format(switched_host.id, "NT" ~ item.nt_id) %}
        {% endif %}
    {% else %}
        {% set slug = "[%s]" | format("NT" ~ item.nt_id) %}
    {% endif %}`

When you open the URL Chooser now, you will get a undefined variable switched_host

sandergo90:
@jockri can you revert this? This is not the correct solution..

dannyvw:
@sandergo90 @jockri If you revert this pull request then you can not switch from domain.

I suggest to change only the changes in selectLink.html.twig or fix the URL Chooser.

sandergo90:
@dannyvw I've created a new pull request to fix the issue. 

https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1346

dannyvw:
@sandergo90 @jockri Looks good, but the PR is now completely reverted which is not good.

As example the getFullHostConfig method is now missing.. this bug is introduced in https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1225. So please revert this pull request and add the fix from #1346

PR #1321, opened 2016-10-07
[Composer]: revert twig to 1.25

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

As of twig 1.26 they removed the twig extension getName() function. 

See https://github.com/twigphp/Twig/issues/2147

Symfony has also released a new tag but we do not install dev releases of symfony. Therefore we need to use twig 1.25 for now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,7 @@
         "twig/extensions": "~1.0",
         "egulias/email-validator": "~1.2",
         "box/spout": "^2.5",
+        "twig/twig": "1.25",
         "ruflin/elastica": "^3.2"
     },
     "require-dev": {


aistis-:
Could you specify a particular issue of Twig v1.26+? How does it break an application?

sandergo90:
Our travis build was not running correctly anymore for the bundles. 

See https://travis-ci.org/Kunstmaan/KunstmaanBundlesCMS/jobs/165451873

"Referencing the "Twig_Extension_Core" extension by its name (defined by getName()) is deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully Qualified Extension Class Name instead: 2x"

aistis-:
I suggest simply remove `getName()` method from all extensions as we do not have stable release anyways :)

sandergo90:
@mattiasdelang has already done this and will create a pull request for this. This does not solve the issue because in the class EnvironmentConfiguration class the following function is used:

$environment->getExtension('core')->setDateFormat($this->dateFormat, $this->intervalFormat);

The 'core' can not be used anymore and this has been changed in a dev release of symfony that we are and will not use :)

aistis-:
This merge request breaks an application:

```
An error occurred when executing the "'cache:clear --no-warmup'" command:                                                                                                                                  
  PHP Fatal error:  Class Kunstmaan\NodeBundle\Twig\PagesConfigurationTwigExtension contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Twig_ExtensionIn  
  terface::getName) in /home/wut/Projects/crowdhouse/www/vendor/kunstmaan/bundles-cms/src/Kunstmaan/NodeBundle/Twig/PagesConfigurationTwigExtension.php on line 57
```

jockri:
Reverted the removal, can you check again?

aistis-:
@jockri Yeah, it works. Thanks

PR #1324, opened 2016-10-10
[SeoBundle] Catch exception when OG image no longer exists

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

When the selected ogImage is no longer present on disk, `get_image_dimensions` will throw an exception.

### DIFF
--- a/src/Kunstmaan/SeoBundle/Twig/SeoTwigExtension.php
+++ b/src/Kunstmaan/SeoBundle/Twig/SeoTwigExtension.php
@@ -244,11 +244,16 @@ class SeoTwigExtension extends Twig_Extension
     /**
      * @param $src
      *
-     * @return Seo
+     * @return array|null
      */
     public function getImageDimensions($src)
     {
-        list($width, $height) = getimagesize($src);
+        try {
+            list($width, $height) = getimagesize($src);
+        } catch (\Exception $e) {
+            return null;
+        }
+
         return array('width' => $width, 'height' => $height);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/SeoBundle/Twig/SeoTwigExtension.php
```diff
@@ -244,11 +244,16 @@ public function setWebsiteTitle($websiteTitle)
...
-     * @return Seo
+     * @return array|bool
      */
     public function getImageDimensions($src)
     {
-        list($width, $height) = getimagesize($src);
+        try {
+            list($width, $height) = getimagesize($src);
+        } catch (\Exception $e) {
+            return false;
```
aistis-: I think `return null` is more convenient :)

wesleylancel: Changed.

src/Kunstmaan/SeoBundle/Twig/SeoTwigExtension.php
```diff
@@ -244,11 +244,16 @@ public function setWebsiteTitle($websiteTitle)
...
      */
     public function getImageDimensions($src)
     {
-        list($width, $height) = getimagesize($src);
+        try {
+            list($width, $height) = getimagesize($src);
+        } catch (\Exception $e) {
+            return null;
+        }
+
```
bakie: why not do a check with file_exists ?
```
if (file_exists($src)) {
    list($width, $height) = getimagesize($src);
            
    return array('width' => $width, 'height' => $height);
}
        
return null;
```

wesleylancel: @bakie Sure, that's possible. Not sure about the exact performance of those two functions, but wouldn't that just hit the disk a second time for the same thing `getimagesize` would already do?

mlebkowski: bare in mind that the file can be stored on a remote server. I turned this feature altogether since my media files are stored in a S3 bucket

PR #1337, opened 2016-10-12
[TranslatorBundle]: Scale textarea to content

mattiasdelang:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

Scale textareas to the content that is being edited.

screen shot 2016-10-12 at 14 30 30

Textarea was too small for the content.

screen shot 2016-10-12 at 14 31 22

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/Resources/ui/js/_inline-edit.js
+++ b/src/Kunstmaan/TranslatorBundle/Resources/ui/js/_inline-edit.js
@@ -9,6 +9,11 @@ kunstmaanTranslatorBundle.inlineEdit = (function(window, undefined) {
 
         $('.js-editable').each(function() {
             var $field = $(this);
+            var width = $field.width();
+
+            $field.on('shown', function(e, editable) {
+                editable.input.$input.css('width', width + 50);
+            });
 
             $field.editable({
                 showbuttons: 'bottom',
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/TranslatorBundle/Resources/ui/js/_inline-edit.js
```diff
@@ -1,32 +1,36 @@
...
     var init;
 
-    init = function() {
+    init = function () {
         $.fn.editable.defaults.mode = 'inline';
 
-        $('.js-editable').each(function() {
+        $('.js-editable').each(function () {
             var $field = $(this);
+            var $width = $field.width();
```
Devolicious: vars only start with $ when it is a jQuery object

Devolicious:
indentation fix is correct, space between function and () needs to be reverted.

src/Kunstmaan/TranslatorBundle/Resources/ui/js/_inline-edit.js
```diff
@@ -9,6 +9,11 @@ kunstmaanTranslatorBundle.inlineEdit = (function(window, undefined) {
 
         $('.js-editable').each(function() {
             var $field = $(this);
+            var width = $field.width();
+
+            $($field).on('shown', function(e, editable) {
```
diskwriter: No need to double wrap $field here. $field is already a jQuery object

PR #1344, opened 2016-10-17
[ConfigBundle] Fixed: non-abstract classes in constructor, code style

aistis-:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |


aistis-:
@jockri could you check the fix? We want to deploy our custom router to production and it fails because of the controller signature: (

Issue #1323, opened 2016-10-10
[AllBundles] Hardcoded admin path pattern


aistis-:
Hello, at the moment it is not possible to customize admin path - for example rename it to `/{locale}/vip` as the current `admin` is hardcoded.

See more: https://github.com/Kunstmaan/KunstmaanBundlesCMS/search?utf8=%E2%9C%93&q=preg_match+admin&type=Code

PR #1349, opened 2016-10-17
[All bundles] Allow /admin/ url to be configurable in parameters.yml

Numkil:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | #1323 |
- [x] Still need to test everything
- [x] write tests for new helper service
- [x] better check for previewurl not via regex
- [x] explore better option configuration
- [x] documententation on configuration and usage

[All bundles] Allow /admin/ url to be configurable in parameters.yml, refactor code to not have duplicate isAdminroute method spread around the codebase

More than likely oversight which breaks the site: If you decide to configure your admin url you have to double check your security.yml file to see if there are no hardcoded /admin/'s left in the firewalls or access_control list.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Helper/AdminRouteHelper.php
+++ b/src/Kunstmaan/AdminBundle/Helper/AdminRouteHelper.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Kunstmaan\AdminBundle\Helper;
+
+use Kunstmaan\NodeBundle\Router\SlugRouter;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Class AdminRouteHelper
+ */
+class AdminRouteHelper
+{
+    protected static $ADMIN_MATCH_REGEX = '/^\/(app_[a-zA-Z]+\.php\/)?([a-zA-Z_-]{2,5}\/)?%s\/(.*)/';
+
+    /**
+     * @var string $adminKey
+     */
+    protected $adminKey;
+
+    /**
+     * @var RequestStack $requestStack
+     */
+    protected $requestStack;
+
+    /**
+     * @param string $adminKey
+     * @param RequestStack $requestStack
+     */
+    public function __construct($adminKey, RequestStack $requestStack)
+    {
+        $this->adminKey = $adminKey;
+        $this->requestStack = $requestStack;
+    }
+
+    /**
+     * Checks wether the given url points to an admin route
+     *
+     * @param string $url
+     *
+     * @return bool
+     */
+    public function isAdminRoute($url)
+    {
+        if ($this->matchesPreviewRoute($url)) {
+            return false;
+        }
+
+        preg_match(sprintf(self::$ADMIN_MATCH_REGEX, $this->adminKey), $url, $matches);
+
+        // Check if path is part of admin area
+        if (count($matches) === 0) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Checks the current request if it's route is equal to SlugRouter::$SLUG_PREVIEW
+     *
+     * @param string $url
+     *
+     * @return boolean
+     */
+    protected function matchesPreviewRoute($url)
+    {
+        $routeName = $this->requestStack->getCurrentRequest()->get('_route');
+
+        return $routeName == SlugRouter::$SLUG_PREVIEW;
+    }
+}
--- a/src/Kunstmaan/AdminBundle/Resources/config/routing.yml
+++ b/src/Kunstmaan/AdminBundle/Resources/config/routing.yml
@@ -1,21 +1,21 @@
 KunstmaanAdminBundle_default:
     resource: '@KunstmaanAdminBundle/Controller/DefaultController.php'
     type:     annotation
-    prefix:   /admin  
+    prefix:   /%kunstmaan_admin.admin_prefix%
 
 KunstmaanAdminBundle_modules:
     resource: '@KunstmaanAdminBundle/Controller/ModulesController.php'
     type:     annotation
-    prefix:   /admin/modules
-        
+    prefix:   /%kunstmaan_admin.admin_prefix%/modules
+
 KunstmaanAdminBundle_settings:
     resource: '@KunstmaanAdminBundle/Controller/SettingsController.php'
     type:     annotation
-    prefix:   /admin/settings      
+    prefix:   /%kunstmaan_admin.admin_prefix%/settings
 
 # Change user password route
 KunstmaanAdminBundle_user_change_password:
-    path: /admin/settings/users/{id}/edit
+    path: /%kunstmaan_admin.admin_prefix%/settings/users/{id}/edit
     defaults: { _controller: FOSUserBundle:ChangePassword:changePassword }
     methods: [GET, POST]
 
@@ -24,27 +24,27 @@ KunstmaanAdminBundle_user_change_password:
 ###########################
 
 fos_user_security_login:
-    path: /admin/login
+    path: /%kunstmaan_admin.admin_prefix%/login
     defaults: { _controller: FOSUserBundle:Security:login }
 
 fos_user_security_check:
-    path: /admin/login_check
+    path: /%kunstmaan_admin.admin_prefix%/login_check
     defaults: { _controller: FOSUserBundle:Security:check }
-        
+
 fos_user_security_logout:
-    path: /admin/logout
+    path: /%kunstmaan_admin.admin_prefix%/logout
     defaults: { _controller: FOSUserBundle:Security:logout }
 
 fos_user_profile:
     resource: '@FOSUserBundle/Resources/config/routing/profile.xml'
-    prefix: /admin/profile
+    prefix: /%kunstmaan_admin.admin_prefix%/profile
 
 fos_user_resetting:
     resource: '@FOSUserBundle/Resources/config/routing/resetting.xml'
-    prefix: /admin/resetting
+    prefix: /%kunstmaan_admin.admin_prefix%/resetting
 
 fos_user_change_password:
-    path: /admin/profile/change-password
+    path: /%kunstmaan_admin.admin_prefix%/profile/change-password
     defaults: { _controller: FOSUserBundle:ChangePassword:changePassword }
     methods: [GET, POST]
 
@@ -52,5 +52,5 @@ fos_user_change_password:
 ## Google OAuth Sign In ##
 ##########################
 KunstmaanAdminBundle_oauth_signin:
-    path: /admin/google_signin_check
+    path: /%kunstmaan_admin.admin_prefix%/google_signin_check
     defaults: { _controller: KunstmaanAdminBundle:OAuth:check }
--- a/src/Kunstmaan/AdminBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/AdminBundle/Resources/config/services.yml
@@ -13,6 +13,8 @@ parameters:
     kunstmaan_admin.firewall.provider_key: 'main'
     kunstmaan_admin.domain_configuration.class: 'Kunstmaan\AdminBundle\Helper\DomainConfiguration'
     kunstmaan_admin.validator.password_restrictions.class: 'Kunstmaan\AdminBundle\Validator\Constraints\PasswordRestrictionsValidator'
+    kunstmaan_admin.adminroute.helper.class: 'Kunstmaan\AdminBundle\Helper\AdminRouteHelper'
+    kunstmaan_admin.adminroute.twig.class: 'Kunstmaan\AdminBundle\Twig\AdminRouteHelperTwigExtension'
 
 services:
     kunstmaan_admin.menubuilder:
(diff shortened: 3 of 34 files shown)


src/Kunstmaan/AdminBundle/Resources/config/routing.yml
```diff
@@ -1,21 +1,21 @@
 KunstmaanAdminBundle_default:
     resource: '@KunstmaanAdminBundle/Controller/DefaultController.php'
     type:     annotation
-    prefix:   /admin  
+    prefix:   /%admin_url%
```
aistis-: I suggest `%admin_prefix%`

src/Kunstmaan/AdminBundle/Resources/config/services.yml
```diff
@@ -88,6 +88,10 @@ services:
         class: Kunstmaan\AdminBundle\Helper\CloneHelper
         arguments: ['@doctrine.orm.entity_manager', '@event_dispatcher']
 
+    kunstmaan_admin.isadminroute.helper:
+        class: Kunstmaan\AdminBundle\Helper\IsAdminRouteHelper
+        arguments: ['%admin_url%']
```
aistis-: How `'%admin_url%` is set? What if we configure it via [bundle configuration](http://symfony.com/doc/current/bundles/configuration.html)?

src/Kunstmaan/AdminBundle/Helper/IsAdminRouteHelper.php
```diff
@@ -0,0 +1,48 @@
...
+        $this->adminKey = $adminKey;
+    }
+
+    /**
+     * Checks wether the given url poins to an admin route
+     * @param string $url
+     *
+     * @return bool
+     */
+    public function isAdminRoute($url)
```
aistis-: Would be great to have a test for this method/class... :)

src/Kunstmaan/AdminBundle/Helper/IsAdminRouteHelper.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        //If the url contains an admin part and a preview part then it is not an admin route
+        preg_match(sprintf('/^(\/app_[a-zA-Z]+\.php)?\/([a-zA-Z_-]{2,5}\/)?%s(\/.*)?\/preview/', $adminKey), $url, $matches);
```
aistis-: Put regular expressions to `array` and use `foreach` to iterate over by matching each of expression

src/Kunstmaan/AdminBundle/Helper/IsAdminRouteHelper.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Kunstmaan\AdminBundle\Helper;
+
+/**
+ * Class IsAdminRouteHelper
+ */
+class IsAdminRouteHelper
```
aistis-: I suggest simply `AdminRouteHelper` as in the future it could have not only `isAdminRoute` method.

src/Kunstmaan/AdminBundle/Resources/config/services.yml
dannyvw: Also set the default to "admin"

src/Kunstmaan/AdminBundle/Helper/IsAdminRouteHelper.php
```diff
@@ -0,0 +1,48 @@
...
+    public function isAdminRoute($url)
+    {
+        //If the url contains an admin part and a preview part then it is not an admin route
+        preg_match(sprintf('/^(\/app_[a-zA-Z]+\.php)?\/([a-zA-Z_-]{2,5}\/)?%s(\/.*)?\/preview/', $adminKey), $url, $matches);
+
+        if (count($matches) > 0) {
+            return false;
+        }
+
+        preg_match(sprintf('/^\/(app_[a-zA-Z]+\.php\/)?([a-zA-Z_-]{2,5}\/)?%s\/(.*)/', $adminKey), $url, $matches);
```
mlebkowski: `$adminKey` is unsafe in the regex context (for example, it could contain a forward slash), should be escaped. Now `foo/admin` would trigger an error. 

It would be nice too if the value were normalized, i.e. leading and trailing slashes would be handled. Should one provide `admin` or `/admin` or `admin/`…?

src/Kunstmaan/AdminBundle/Helper/IsAdminRouteHelper.php
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Checks wether the given url poins to an admin route
+     * @param string $url
+     *
+     * @return bool
+     */
+    public function isAdminRoute($url)
+    {
+        //If the url contains an admin part and a preview part then it is not an admin route
```
mlebkowski: How about actually checking the request if `_slug_preview` route matched?

src/Kunstmaan/AdminBundle/Resources/config/services.yml
```diff
@@ -88,6 +88,10 @@ services:
         class: Kunstmaan\AdminBundle\Helper\CloneHelper
         arguments: ['@doctrine.orm.entity_manager', '@event_dispatcher']
 
+    kunstmaan_admin.isadminroute.helper:
+        class: Kunstmaan\AdminBundle\Helper\IsAdminRouteHelper
```
mlebkowski: unquoted value will trigger a notice in recent symfony versions

src/Kunstmaan/AdminBundle/Helper/IsAdminRouteHelper.php
aistis-: Yeap, would be a better approach than hardcoded path regexp :+1:

Numkil:
@mlebkowski @aistis- @dannyvw 
thx for all the feedback guys. I will process everything this evening and update the PR

src/Kunstmaan/AdminBundle/Resources/config/services.yml
Numkil: @aistis- 
I've been puzzled about the best way to attack how I should actually make this parameter configurable. For now I settled on just relying on the application parameters.yml. Bundle configuration did not seem like the right tool for the job because of a few reasons. 

Configuration for this parameter should be duplicated across all the bundles included in this PR or advanced code in the AdminBundle is needed to check if any of the other depended bundles are active and pass the parameter on. Having this parameter defined for each bundle also makes the configuration a lot more error prone, one could override the admin_prefix for one bundle but fail to adjust it in the adminbundle.

A custom route loader would be necessary I believe to allow for this amount of configuration without using the symfony parameters.yml. If possible I would like to avoid that and keep it simple.

It also makes sense to put this configuration option in the application parameters.yml in my opinion. It's never a choice you make for each bundle on its own but rather for the whole application.

If you have other opinions on this topic or have more deep knowledge on how to workaround this issue, please let me know. That is exactly the reason why I made this PR public so soon before finishing it. 

@dannyvw I was thinking about putting admin_prefix: 'admin' as a default in the parameters.yml.dist of our kunstmaanbundlesstandardedition. That way new projects would have the option set by default.

src/Kunstmaan/AdminBundle/Helper/IsAdminRouteHelper.php
Numkil: Great remarks! I will escape and normalize the value.

src/Kunstmaan/AdminBundle/Resources/config/services.yml
aistis-: Agree with @mlebkowski :+1:

src/Kunstmaan/AdminBundle/Helper/AdminRouteHelper.php
```diff
@@ -17,18 +19,18 @@ class AdminRouteHelper
...
-     * @param SlugRouter $slugRouter
+     * @param RequestStack $requestStack
      */
-    public function __construct($adminKey, SlugRouter $slugRouter)
+    public function __construct($adminKey, RequestStack $requestStack)
     {
-        $this->adminKey = $this->normalizeUrlSlice($adminKey);
-        $this->slugRouter = $slugRouter;
+        $this->adminKey = $adminKey;
+        $this->request = $requestStack->getCurrentRequest();
```
mlebkowski: There could be more than one request during the lifetime of this service (subrequests, forwards, etc). It’s better to just store the `requestStack` and fetch the request each time.

src/Kunstmaan/AdminBundle/Tests/Helper/AdminRouteHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $this->assertFalse($result);
+    }
+
+    /**
+     * @covers Kunstmaan\AdminBundle\Helper\AdminRouteHelper::isAdminRoute
+     */
+    public function testIsAdminRouteReturnsFalseWhenPreviewUrl()
+    {
+        $requestStack = $this->getMock('Symfony\Component\HttpFoundation\RequestStack');
```
aistis-: RequestStack::class

src/Kunstmaan/AdminBundle/Tests/Helper/AdminRouteHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+    protected function setUp()
+    {
+        $this->adminRouteHelper = $this->getAdminRouteHelper();
+    }
+
+    /**
+     * Tears down the fixture, for example, closes a network connection.
+     * This method is called after a test is executed.
+     */
+    protected function tearDown()
```
aistis-: Not necessary method.

src/Kunstmaan/AdminBundle/Tests/Helper/AdminRouteHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+    protected static $ADMIN_URL = "/admin/blog/page/1";
+    protected static $PREVIEW_ADMIN_URL = "/admin/preview/blog/page/1";
+
+    /**
+     * @var AdminRouteHelper
+     */
+    protected $adminRouteHelper;
+
+    /**
+     * Sets up the fixture, for example, opens a network connection.
```
aistis-: Please use `{@inheritdoc}`

src/Kunstmaan/AdminBundle/Tests/Helper/AdminRouteHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+    private function getAdminRouteHelper()
+    {
+        $adminRouteHelper = new AdminRouteHelper(self::$ADMIN_KEY, $this->getRequestStack());
+
+        return $adminRouteHelper;
+    }
+
+    private function getRequestStack()
+    {
+        $requestStack = $this->getMock('Symfony\Component\HttpFoundation\RequestStack');
```
aistis-: `RequestStack::class`

src/Kunstmaan/AdminBundle/Tests/Helper/AdminRouteHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+namespace Kunstmaan\AdminBundle\Tests\Helper;
+
+use Kunstmaan\AdminBundle\Helper\AdminRouteHelper;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Kunstmaan\NodeBundle\Router\SlugRouter;
+
+class AdminRouteHelperTest extends \PHPUnit_Framework_TestCase
+{
+    protected static $ADMIN_KEY = "admin";
```
aistis-: Please test with a wider test data set. It would be nice if you use `@dataProvider` annotation with various cases. Not just with a default :)

Please test at least these cases:

When prefix is `admin`:
* domain.tld/admin
* domain.tld/en/admin
* domain.tld/en/some_path/admin

When prefix is `vip`:
* domain.tld/vip
* domain.tld/en/vip
* domain.tld/en/some_path/vip

aistis-: + use single quotes instead of double if it is not necessary :)

src/Kunstmaan/AdminBundle/Tests/Helper/AdminRouteHelperTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $result = $adminRouteHelper->isAdminRoute(self::$PREVIEW_ADMIN_URL);
+
+        $this->assertFalse($result);
+    }
+
+    private function getAdminRouteHelper()
+    {
+        $adminRouteHelper = new AdminRouteHelper(self::$ADMIN_KEY, $this->getRequestStack());
+
+        return $adminRouteHelper;
```
aistis-: Why not?

`return new AdminRouteHelper(self::$ADMIN_KEY, $this->getRequestStack());`

src/Kunstmaan/AdminBundle/Tests/Helper/AdminRouteHelperTest.php
mlebkowski: well, in fact you don’t even need to mock the `RequestStack`, just use it directly. 

```php
$requestStack = new RequestStack;
$requestStack->push($this->getPreviewRequest());
```

src/Kunstmaan/MultiDomainBundle/Tests/Helper/DomainConfigurationTest.php
```diff
@@ -318,6 +317,24 @@ private function getEntityManager()
         return $em;
     }
 
+    private function getAdminRouteHelper()
```
aistis-: @Numkil this test is useless as you are supposed to test `AdminRouteHelper`, but not mock it :(

Numkil: There is no need for testing the isAdminRoute method from the AdminRouteHelper again in the DomainConfigurationTest in my opinion. We moved the logic into a service for this exact reason, I tested the AdminRouteHelper on it's own so we can assume that service works as intended. Therefore we can mock it here and not have to worry about anything else than the functionality of the DomainConfiguration.

aistis-:
I guess `src/Kunstmaan/TranslatorBundle/Tests/app/cache/phpunit/test.db` is not necessary

aistis-:
I guess `src/Kunstmaan/TranslatorBundle/Tests/app/cache/phpunit/test.db` is not necessary

aistis-:
I guess `src/Kunstmaan/TranslatorBundle/Tests/app/cache/phpunit/test.db` is not necessary

src/Kunstmaan/MultiDomainBundle/Tests/Helper/DomainConfigurationTest.php
aistis-: Oh sorry, you are right. I missed `AdminRouteHelperTest` class

Numkil:
@aistis- I removed the bin file, and put it in the .gitignore immediately so we don't keep accidentally committing these.

Devolicious:
@Numkil could you please rebase to 1 commit.

PR #1368, opened 2016-11-07
[NodeBundle]: check if we have a request

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php
+++ b/src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php
@@ -71,6 +71,10 @@ class NodeChoiceType extends AbstractType
 
     private function getCurrentLocale()
     {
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php
```diff
@@ -71,6 +71,10 @@ public function getParent()
 
     private function getCurrentLocale()
     {
+        if (is_null($this->requestStack->getCurrentRequest())) {
```
aistis-: `=== null` is faster :D

sandergo90:
Especially for you @aistis- I have changed it 😅

PR #1372, opened 2016-11-09
[AdminList] Allow custom xlsx template on string values

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Allow custom xlsx template on string values, check on template existence not on being an object.


jverdeyen:
Checks fail:
```
The Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\DebugHandlerPass class is deprecated since version 2.12 and will be removed in 3.0. Use AddDebugLogProcessorPass in FrameworkBundle instead: 1x
    1x in TestListener::startTestSuite
```

Issue #1332, opened 2016-10-11
Nested regions in page template config


jockri:
In the page template configuration (yml) it should be possible the nest regions.
In this way a sidebar can span multiple rows.

mlebkowski:
You’re talking about this piece?  ;)

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/PagePartBundle/DependencyInjection/Configuration.php#L51-L54

jockri:
If you want a sidebar that spans on multiple regions, this is not possible in the admin.
The first scheme is the current situation, the second scheme is the would be situation where the "sidebar region" spans multiple other regions on the right.

![untitled diagram](https://cloud.githubusercontent.com/assets/517574/19303162/e3afbd6a-9067-11e6-8b7b-04ac02412ab5.png)

Perni1984:
I realised something like that already, writing a separate PageTemplateConfigurationReader, that reads a dynamic page templated configuration with nested rows / columns out of the db. I am not sure how you would like to implement this, but I imagine I could eventually extract some of the code to a separate bundle.

sandergo90:
HI @Perni1984,

I'm currently working on a pull request for nested regions. Almost finished! 

Cheers

Perni1984:
@sandergo90: Ok, I see - thanks for the info - I guess I must have missed that on the roadmap somewhere. BTW if you are interested in the dynamic page templates I can support on that aswell - see some real-life example of that here:
![image](https://cloud.githubusercontent.com/assets/9039956/19756843/af2e8b8e-9c1f-11e6-96ab-4e95e894ba57.png)

sandergo90:
Looking good @Perni1984 

Here are some screens of what I am doing.

https://cl.ly/3q2E0P1l2H3U

https://cl.ly/1A1S3n1H2S0f

Perni1984:
@sandergo90: perfect - looks similar ;-) Eventually there would have been some synergies before, when I knew that you guys needed such a feature aswell.

sandergo90:
Indeed ;)

And for the configuration, this looks like the following:

![screen shot 2016-10-27 at 08 37 14](https://cloud.githubusercontent.com/assets/1516413/19756997/9ca18f2e-9c20-11e6-9544-0b9310e39428.png)

wesleylancel:
[...]

PR #1379, opened 2016-11-22
[NodeBundle]: Nested regions

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1332


src/Kunstmaan/NodeBundle/Resources/translations/messages.en.yml
```diff
@@ -167,3 +168,6 @@ kuma_node:
             button:
                 copy:   Copy
                 cancel: Cancel
+        export_pagetemplate:
+            button:
+                title:   Export pagetemplate
```
aistis-: Not necessary spacing

src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/Modals/_pagetemplate.html.twig
```diff
@@ -0,0 +1,34 @@
+{% import _self as macros %}
+
+{% for row in pageTemplate.rows %}
+<div class="row">
```
aistis-: Hardly readable indentation

src/Kunstmaan/PagePartBundle/PageTemplate/PageTemplateConfigurationParser.php
```diff
@@ -65,20 +64,45 @@ public function parse($name)
...
         $children = [];
+        $rows = [];
 
         if (isset($rawRegion['regions']) && count($rawRegion['regions'])) {
             foreach ($rawRegion['regions'] as $child) {
                 $children[] = $this->buildRegion($child);
             }
         }
 
+        if (isset($rawRegion['rows']) && count($rawRegion['rows'])) {
```
mlebkowski: instead of `if (isset() && count())` you can just set a default values:

```
$rawRegion = array_replace(['regions' => [], 'rows' => [], $rawRegion);
```

src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/Modals/_pagetemplate.html.twig
aistis-: You should use [whitespace control modifier](http://vvv.tobiassjosten.net/symfony/trim-and-remove-twig-whitespace/).

sandergo90: Hi @aistis- 

This will never give back to correct representation of the grid system.

My twig now returns:

`<div class="row">
    <div class="col-md-3"></div>
    <div class="col-md-9">
    <div class="row">
        <div class="col-md-6"></div>
        <div class="col-md-6"></div>
    </div>
    <div class="row">
        <div class="col-md-12"></div>
    </div>
</div>`

How do you think this will do the same with the whitespace control modifier?

aistis-: Hmmmm, and what is the original problem? Is it that the generated source is not aligned nicely?

sandergo90: Yes indeed, when i format the twig file the output will contain all the spaces of the indentation. But that's not how we want the alignment, it should be like the real html structure.

aistis-: But is it a difference? I mean, browser will still render the same regarding to indentation of generated source received from the server.

sandergo90: Yes it's a difference. Because we want the html structure to be rendered as text so you can copy the template.

aistis-: Ok. Got it :+1:

sandergo90: Perfect, sorry for the misunderstanding ;)

sandergo90:
@todo add "Export pagetemplate" only for super admins

sandergo90:
@Devolicious updated this PR to base branch 4.0

PR #1380, opened 2016-11-25
[AdminListBundle]: don't use query cache when gedmo annotations are used

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

When we use for example the Softdeleteable annotation of Gedmo, our admin lists don't work properly. When you go to your admin list which has a Softdeleteable entity and delete an entity from the list, the entity will still remain in the admin list but the entity is already adapted in the database. The query cache causes this problem. Therefor this check is needed. If any Gedmo annotations are used, this can also be the Translatable, drop the query cache.

### DIFF
--- a/src/Kunstmaan/AdminListBundle/AdminList/Configurator/AbstractDoctrineORMAdminListConfigurator.php
+++ b/src/Kunstmaan/AdminListBundle/AdminList/Configurator/AbstractDoctrineORMAdminListConfigurator.php
@@ -2,6 +2,7 @@
 
 namespace Kunstmaan\AdminListBundle\AdminList\Configurator;
 
+use Doctrine\Common\Annotations\AnnotationReader;
 use Kunstmaan\AdminListBundle\AdminList\SortableInterface;
 use Traversable;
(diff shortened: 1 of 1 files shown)


aistis-:
I suggest to have an out of the box caching disabling and do not couple it with `Gedmo` annotations. It even could be not annotations, but the mapping could be done via yml or xml as well.

sandergo90:
I think we should use this implementation for Gedmo because it's a common case. If you want to disable the query cache for something else then Gedmo, you can just override the getQuery() method for your admin list configuration and disable the caching there.

mlebkowski:
This shouldn’t be in the core. It covers only one specific use case, it doesn’t fix a problem the core introduces (i.e. the softdeletable is not used in the core), etc. The fix itself does a comparison using a class name and a magic value ("Gedmo"), so produces false positives.

sandergo90:
Yes indeed, it's maybe a to specific use case. I'll just add it to my project, it does not have to be included in the bundles then.

PR #1417, opened 2016-12-20
[NodeSearchBundle] strip all javascript including the script tags

bijsterdee:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

When you have javascript in a pagepart, the search provider will index a node with the javascript (without the script tags). When you search it will return the javascript code.

This fix will strip all the javascript inside the script tag.

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
+++ b/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
@@ -640,8 +640,11 @@ class NodePagesConfiguration implements SearchConfigurationInterface
      */
     protected function removeHtml($text)
     {
+        // Strip all javascript including the script tags
+        $result = preg_replace('/<script\b[^>]*>(.*?)<\/script>/is', "", $text);
+        
         // Remove HTML markup
-        $result = strip_tags($text);
+        $result = strip_tags($result);
 
         // Decode HTML entities
         $result = trim(html_entity_decode($result, ENT_QUOTES));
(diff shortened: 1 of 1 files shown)


mlebkowski:
I had the same issue with HTML content and `strip_tags`, but not with Kunstmaan Search! I used a similar solution:

```php
// first, remove all blacklisted nodes incl. their content (strip tags would leave the content):
/** @var \Symfony\Component\DomCrawler\Crawler $crawler */
$crawler->filter('style, script')->each(function (Crawler $blacklisted) {
    /** @var \DOMElement $node */
    foreach ($blacklisted as $node) {
        $node->parentNode->removeChild($node);
    }
});
```


You may want to add `<style>` tags as well.

Devolicious:
Personally I like @mlebkowski 's approach more than the regex. It opens up an easy hook to define your own nodes you want to filter out more easily than with regex.

Defining blacklisted nodes in an array configuration would be extremely useful when dealing with custom webcomponent html tags. So by default this config could be empty and in case of @bijsterdee it would just contain `script`. But you could as easily add `style` or some custom html tag.

bijsterdee:
Thank you all for the reply. In our case, it was just a quick fix to strip the javascript (instead of just strip_tags). I agree with @mlebkowski that his approach is more dynamic. Also I added styling support.

Devolicious:
@bijsterdee Could you please refactor to @mlebkowski 's approach?

bijsterdee:
@Devolicious not at this time

sandergo90:
@bijsterdee when you have the time, can you continue this PR ? Thanks!

cv65kr:
@sandergo90 i fix this in #1630

sandergo90:
Superb, i will close this PR then.

PR #1273, opened 2016-08-22
[GeneratorBundle] Changed the entity generator to use the same fields as pagepart generator

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #670

PR #1422, opened 2016-12-23
[GeneratorBundle]: set default values

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

When generating a new page, an error will popup that 'minHeight' does not exists. This is related to #1273 where the generator was adapted. MinHeight etc. is not set when the type of field is different from image of media.


Devolicious:
@sandergo90 could you please add brackets to the else statement for a better code styling.

sandergo90:
@Devolicious I've updated the else statement and formatted the other code also properly.

PR #1424, opened 2017-01-02
[PagePartBundle]: make page variable available inside pagepart admin

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

We are using the page variable very frequently inside our pagepart twig templates. The page variable is not available when editing a node. Therefor you need to create a admin view template for this issue. By adding the page variable always, we can use it also in the admin views.


Numkil:
Is it possible to also make the variable "node" which is the current node available? Seems like we use it often as well.

sandergo90:
@Numkil that's not possible for now. The pageparts are rendered with a twig extensions with no knowledge of the current node at that moment.

PR #1426, opened 2017-01-05
[PagePartBundle]: add post persist event

sandergo90:
[PagePartBundle]: add post persist event

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

When using the TaggingBundle and adding tags on a pagePart, the tags are not persisted in the database when saving the node. Therefore I've added a post persist event for the PageParts that can be used in the TaggingListener for each pagePart.

### DIFF
--- a/src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
+++ b/src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Kunstmaan\PagePartBundle\Event;
+
+use Kunstmaan\NodeBundle\Entity\HasNodeInterface;
+use Kunstmaan\NodeBundle\Entity\Node;
+use Kunstmaan\NodeBundle\Entity\NodeTranslation;
+use Kunstmaan\NodeBundle\Entity\NodeVersion;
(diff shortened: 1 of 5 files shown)


src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public function getPagePart()
+    {
+        return $this->pagePart;
+    }
+
+    /**
+     * @param PagePartInterface $pagePart
+     */
+    public function setPagePart($pagePart)
```
aistis-: setPagePart(PagePartInterface $pagePart)

src/Kunstmaan/PagePartBundle/PagePartAdmin/PagePartAdmin.php
```diff
@@ -75,15 +77,16 @@ public function __construct(
         HasPagePartsInterface $page,
         $context = null,
         ContainerInterface $container = null
-    ) {
+    )
```
aistis-: Lets use `) {` when constructor arguments are multilined

PR #1429, opened 2017-01-09
[TaggingBundle]: make tag names translatable

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


sandergo90:
@jockri are there any benefits of doing this ?

jockri:
I would add the Translatable interface, even if there are no methods present in the interface, it can be used later on to identify the entities that are translatable in listeners for example. 

I would also add the `locale` property and the `setTranslatableLocale` function so that you are more flexible in manipulating the translatable behaviour. Useful in tests etc.

sandergo90:
@jockri Alright, done ;)

Issue #1066, opened 2016-03-08
[AdminList] It is not possible to sort entites by group


mtnorthrop:
It's currently not possible to sort entities within a group.

When a new entity is added to an admin list through the `AdminListController::doAddAction()` method, `AdminListController::getMaxSortableField()` is called and returns the highest sorting value, without any option for grouping the entities according to a category or a foreign key.

In my case, I need to sort entities, grouped by foreign key. Since the built-in sortable functionality can not do this, I'm using the Gedmo Sortable Doctrine extension instead.

Overriding the default `moveUp()` and `moveDown()` methods to work with Gedmo Sortable wasn't a problem, but the `doAddAction()` method calls `AdminListController::getMaxSortableField()`, which is a private.

In order to get around this limitation I've overridden the whole `AdminListController::doAdminAction()` method, but I don't feel like this is an ideal solution.

Would it be possible to allow developers to implement their own method for setting the weight value for new entities? Maybe by moving the following code to its own protected method that can be overridden?

```
[From AdminListController::doAddAction()]

// Check if Sortable interface is implemented
if ($configurator instanceof SortableInterface) {
    $sort = $configurator->getSortableField();
    $weight = $this->getMaxSortableField($repo, $sort);
    $setter = "set".ucfirst($sort);
    $helper->$setter($weight + 1);
}
```

This way, if a developer wants to use the Gedmo Sortable extension, they can simply override the parent method with a blank method.

There's probably a much better and more systematic solution, but this would solve the immediate problem.

sandergo90:
Hi,

Do you just want to sort the admin list ? Because that way you can just adapt the query builder and change the orderby clause.

Sander

mtnorthrop:
This is actually a problem when trying to add a new entity to an admin list, not so much when fetching them for the view.

For example, let's say I have a Car entity, but I want to have a separate admin list for each Brand. It's no problem to create an admin list for this by adapting the query builder; **however**, when I create a new Car of a certain Brand using the `AdminListController::doAddAction()` method, it uses the private `AdminListController::getMaxSortableField()` method, which gets the highest sortable value from the database and sets the new entity's sortable value as one more. 

I think that, when getting the highest sortable value, it should be possible add conditions. For example, to get the highest sortable value from the Cars table, only counting rows with a brand_id of "5". 

[...]

PR #1432, opened 2017-01-11
[AdminListBundle] Moved setting of sort weight on new items to a protected method

mtnorthrop:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1066 

This change makes it easier to use other sorting strategies besides the default one. For example, if you want to use the Gedmo Sortable behavior the sort weight is set automatically, meaning that the code that sets the sort weight in the AdminListController can be skipped.


src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
```diff
@@ -105,7 +105,6 @@ protected function doAddAction(AbstractAdminListConfigurator $configurator, $typ
 
         /* @var EntityManager $em */
         $em = $this->getEntityManager();
-        $repo = $em->getRepository($configurator->getRepositoryName());
```
mtnorthrop: Setting the $repo variable was moved to the new method, as it was not used anywhere else.

src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
```diff
@@ -456,6 +450,25 @@ private function getMaxSortableField($repo, $sort)
...
+    /**
+     * Sets the sort weight on a new item. Can be overridden if a non-default sorting implementation is being used.
+     *
+     * @param AbstractAdminListConfigurator $configurator The adminlist configurator
+     * @param $item
+     *
+     * @return mixed
+     */
+    protected function setSortWeightOnNewItem(AbstractAdminListConfigurator $configurator, $item) {
+        if ($configurator instanceof SortableInterface) {
```
mtnorthrop: I moved the check for the SortableInterface here, since people using other sorting implementations shouldn't be required to implement this interface if their preferred sorting method does not require it.

mtnorthrop:
Hello @Devolicious , I've changed the base branch to Kunstmaan:4.0. Is this correct now?

Devolicious:
Hi @mtnorthrop,

Sorry for the late response but yes, this is a good start. Except that you should rebase your PR starting from 4.0 and drop the commits except yours. Because you started from master this will cause a back merge from master to 4.0 (see the extra commits) and that's something we don't want to happen :)

mtnorthrop:
Hello @Devolicious,

Thanks for the response. I'll try to get this set up properly for you. I apologize for the complications, but, as a solo developer, I mainly use Git for version control and don't have that much experience with its more collaborative aspects.

mtnorthrop:
I'm going to close this PR and make another clean attempt at doing this correctly.

PR #1433, opened 2017-01-12
[AdminBundle] Also use replacedUrl in CKEDITOR instance

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Otherwise you will get something like `[NT8]`.


jverdeyen:
Related to #1434

PR #1436, opened 2017-01-18
[ALL] updated twig includes to correct convention

Devolicious:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Default/_ckeditor_configs.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Default/_ckeditor_configs.html.twig
@@ -46,4 +46,4 @@
     };
 </script>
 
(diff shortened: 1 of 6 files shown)


wesleylancel:
@Devolicious: I think @ notation is recommended by Symfony

Devolicious:
@wesleylancel I know they support both syntaxes by default, but I just read that the namespaced syntax is faster. http://symfony.com/doc/current/templating/namespaced_paths.html

So changes are on it's way :)

wesleylancel:
@Devolicious Yeah, both are supported, but @ is recommended. There's even talk about deprecating the other syntax: https://github.com/symfony/symfony/pull/21035

PR #1442, opened 2017-01-24
[NodeBundle] Remove default for is_homepage

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | n/a

If a default value is applied the PagesConfiguration will never fallback to checking wether or node `HomePageInterface` is implemented. See https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/NodeBundle/Helper/PagesConfiguration.php#L110


Devolicious:
Hi @wesleylancel,

I've updated your PR and we will merge it to the 5.0 branch because this change implies a BC break when you're using the configuration but didn't set the `is_homepage` because it's a default.

PR #1450, opened 2017-02-02
[NodeBundle] ignore slug of structurenode when changing slug of page

bakie:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

### DIFF
--- a/src/Kunstmaan/NodeBundle/EventListener/NodeTranslationListener.php
+++ b/src/Kunstmaan/NodeBundle/EventListener/NodeTranslationListener.php
@@ -12,6 +12,7 @@ use Kunstmaan\NodeBundle\Entity\HasNodeInterface;
 use Kunstmaan\NodeBundle\Entity\Node;
 use Kunstmaan\NodeBundle\Entity\NodeTranslation;
 use Kunstmaan\NodeBundle\Entity\NodeVersion;
+use Kunstmaan\NodeBundle\Helper\PagesConfiguration;
 use Kunstmaan\NodeBundle\Repository\NodeTranslationRepository;
 use Kunstmaan\UtilitiesBundle\Helper\SlugifierInterface;
 use Symfony\Bridge\Monolog\Logger;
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/NodeBundle/EventListener/NodeTranslationListener.php
```diff
@@ -324,7 +325,18 @@ private function ensureUniqueUrl(
...
         );
 
-        if (count($translations) > 0) {
+        $translationsWithSameUrl = array();
+
+        /** @var NodeTranslation $trans */
+        foreach ($translations as $trans) {
+            $refEntityName = $trans->getPublicNodeVersion()->getRefEntityName();
+            $object = new $refEntityName();
+            if (!$object instanceof StructureNode) {
```
jockri: Also call the method of StructureNode to check the boolean?

mlebkowski: How is even `StructureNode` any special here? This is an implementation detail. I have structure nodes/pages that do not extend this class. `instanceof StructureNode` does not determine if a page is a structure node. `HasNodeInterface::isStructureNode()` does.

Also in this commit:

 * Creating a page assuming it has no constructor arguments. Poor design
 * Checking if instance implements interface / extends class, instead of checking if class extends / implements (using `class_extends` / `class_implements`). This would fix the first issue.
 * Reimplementing business logic. Please use `PagesConfiguration::isStructureNode($refNameOrInstance)` for this

src/Kunstmaan/NodeBundle/EventListener/NodeTranslationListener.php
```diff
@@ -324,7 +332,16 @@ private function ensureUniqueUrl(
             ).' node(s) that match url \''.$translation->getUrl().'\''
         );
 
-        if (count($translations) > 0) {
+        $translationsWithSameUrl = array();
+
+        /** @var NodeTranslation $trans */
+        foreach ($translations as $trans) {
+            if ($this->pagesConfiguration->isStructureNode($trans->getPublicNodeVersion()->getRefEntityName())) {
```
Devolicious: Is this not supposed to be negatory? As I'm reading this you're only including structured nodes instead of ignoring them.

PR #1451, opened 2017-02-03
[NodeSearchBundle] Make NodeSearcher properties protected for extending

b-franco:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

Make the properties of NodeSearcher protected so it's easier/cleaner to extend it.

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Search/NodeSearcher.php
+++ b/src/Kunstmaan/NodeSearchBundle/Search/NodeSearcher.php
@@ -18,22 +18,22 @@ class NodeSearcher extends AbstractElasticaSearcher
     /**
      * @var TokenStorageInterface
      */
-    private $tokenStorage = null;
+    protected $tokenStorage = null;
 
     /**
      * @var DomainConfigurationInterface
      */
-    private $domainConfiguration;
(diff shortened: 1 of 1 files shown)


mlebkowski:
You don’t need this for extending. Create your own private fields for those dependencies you need, and overrider the constructor to get them (or in this case, override setters):

```php
class ExtendedNodeSearcher extends NodeSearcher {
    private $em;

    public function setEntityManager(EntityManager $em) {
        $this->em = $em;
        parent::setEntityManager($em);
    }

    public function foo() {
        $this->em…;
    }
}

b-franco:
Well, I never said I "need" it. The solution you suggest is how we currently implement it. But this means redefining 4 properties and 4 setters with calls to their parents. It would be better if the properties would be accessible in extending classes.

Devolicious:
Hi @b-franco,

I like your changes but they implicate a BC break since you can't define properties with a more strict access level than it's parent. So I would like to ask you to rebase and submit this PR to the 5.0 branch and add some documentation to the upgrade guide.

Thanks and keep up the good work!

PR #1458, opened 2017-02-14
Add pagepart_index to pagepart templates

tarjei:
I know you can access the loop.index0 variable directly within the PagePart template, but I think it's worth adding it as an explicit API point.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no


Devolicious:
Hi @tarjei,

We talked it over and decided not to merge this PR because we don't have any particular use cases for it and therefor find it unnecessary variable setting/usage. Also because you have the option to simply overwrite the twig file in app/Resources since it's a small and uncomplicated template. To our opinion that's a more suited approach for an edge case like this.

PR #1459, opened 2017-02-14
[Composer] Upgrade to latest LiipImagine bundle

tarjei:
There has happened quite a lot since the last upgrade - it would be nice to have access to that functionality.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | unknown
| Deprecations? | no


Devolicious:
Hi @tarjei,

We have locked to that specific version because we rely on the feature to convert from one format to another format. 

We will look into it to see how we are going to handle this.

PR #1464, opened 2017-02-25
[AdminListBundle] Moved setting of sort weight on new items to a protected method (revised)

mtnorthrop:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1066 

This change makes it easier to use other sorting strategies besides the default one. For example, if you want to use the Gedmo Sortable behavior the sort weight is set automatically, meaning that the code that sets the sort weight in the AdminListController can be skipped.

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
+++ b/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
@@ -105,7 +105,6 @@ abstract class AdminListController extends Controller
 
         /* @var EntityManager $em */
         $em = $this->getEntityManager();
-        $repo = $em->getRepository($configurator->getRepositoryName());
         $entityName = null;
         if (isset($type)) {
             $entityName = $type;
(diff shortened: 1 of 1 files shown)


mtnorthrop:
@Devolicious I've created a completely new pull request to try to clean up the one I submitted before. I think the directions on https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/docs/07-02-pull-requests.md#select-the-correct-branch-for-your-pull-request were throwing me off, because all the examples are for the `master` branch and and was trying to follow those.

Could you have a look at this PR and let me know if I've submitted it properly this time?

Thanks!

Devolicious:
Hi @mtnorthrop, 

This PR is good. 
Regarding the selection of the correct branch, only when it's a new feature or a deprecation you need to select master, bugfixes must be submitted to the latest minor branch (in this case 4.0). Security fixes have to be submitted to the lowest minor branch which in this case is also 4.0 and when you are developing a new feature that generates a BC break you need to submit it to the next major (5.0). 
Does this explanation makes it clearer how to submit your PR? Should we add more guidance or do you have any recommendations?

mtnorthrop:
Hello @Devolicious,

Thanks for your feedback. The problem was mainly on my side, but it would probably help developers like myself, who are less experienced with submitting PRs, if the docs on https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/docs/07-02-pull-requests.md#rebase-your-pull-request were updated to make it clear that `master` is just being used as an example and should be replaced with the appropriate branch name for bug fixes.

This would probably be obvious to developers with more GitHub experience, but updating the docs might help us noobs. :-)

Devolicious:
@mtnorthrop noted :) We'll try to improve our docs or if you want you can submit a PR for that as well ;)

Thanks for the contribution!

Issue #1469, opened 2017-03-02
Password strength configuration


Devolicious:
Add optional configuration to require better passwords

Should be able to assign:

- min length
- max length (optional)
- min capitals
- min numbers
- min symbols

PR #1473, opened 2017-03-06
[AdminBundle] Added optional password restrictions

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1469

To set the optional parameters for the password restrictions, in app/config.yml:
```
kunstmaan_admin:
    password_restrictions:
        min_digits: 2
        min_uppercase: 2
        min_special_characters: 1
        min_length: 8
        max_length: 15
```
**TODO:**
- [ ] Submit changes to the documentation

### DIFF
--- a/src/Kunstmaan/AdminBundle/DependencyInjection/Configuration.php
+++ b/src/Kunstmaan/AdminBundle/DependencyInjection/Configuration.php
@@ -88,6 +88,15 @@ class Configuration implements ConfigurationInterface
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('password_restrictions')
+                ->addDefaultsIfNotSet()
+                    ->children()
+                        ->integerNode('min_digits')->defaultNull()->end()
+                        ->integerNode('min_uppercase')->defaultNull()->end()
+                        ->integerNode('min_special_characters')->defaultNull()->end()
+                        ->integerNode('min_length')->defaultNull()->end()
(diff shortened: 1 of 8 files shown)


src/Kunstmaan/AdminBundle/DependencyInjection/Configuration.php
```diff
@@ -88,6 +88,15 @@ public function getConfigTreeBuilder()
...
                     ->end()
                 ->end()
+                ->arrayNode('password_restrictions')
+                ->addDefaultsIfNotSet()
+                    ->children()
+                        ->integerNode('min_digits')->defaultNull()->end()
+                        ->integerNode('min_uppercase')->defaultNull()->end()
+                        ->integerNode('min_special_characters')->defaultNull()->end()
+                        ->integerNode('min_length')->defaultNull()->end()
+                        ->integerNode('max_length')->defaultNull()->end()
```
jockri: I don't think we need to limit the password lenght

deZinc: The limit of the the password length is optional, so it's not necessary to set it.
(All restrictions are optional.)

mlebkowski: Why would you encourage anyone to set a maximum password length (by providing this option)? Please don’t! @jockri made a great point!

Issue #1411, opened 2016-12-13
Feature request: notes for redirects


aistis-:
Hello!

It would be nice to be able add some notes for created redirects via CMS. It would be just a text field for notes, which would make easier to remember why the redirect is created or whatever is the reason :)

In one of our projects we have more than 30 redirects and it became hard to remember why the old redirects still exists.

What do you think?

sandergo90:
Hi @aistis- 

Sounds like a good use case. You can make a PR for this ;)

Cheers

Numkil:
Merging because feature implemented.

PR #1490, opened 2017-03-30
[RedirectBundle]: add note to redirect

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1411

### DIFF
--- a/UPGRADE-4.1.md
+++ b/UPGRADE-4.1.md
@@ -1,0 +1,9 @@
+# UPGRADE FROM 4.0 to 4.1
+
+## RedirectBundle
+
+Added note to the RedirectBundle:
+
+https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1490
+
+Please be sure to execute database migrations after upgrading.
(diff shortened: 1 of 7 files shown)


mlebkowski:
Is this a breaking change? it requires database migration

sandergo90:
@mlebkowski a new field in an entity is not a breaking change I think? It requires to execute a diff and a migrated indeed. But is it therefore a breaking change?

wesleylancel:
@sandergo90 @mlebkowski: I think it is, because it requires a manual action (generating the migration) if you don't want your code to break...

mlebkowski:
I honestly don’t know. For me it’s a minor change as long as it’s clearly stated in the changelog/migration guide.  It won’t break any existing application code or libraries (since the field is nullable), and it’s a trivial fix

Devolicious:
Imho it's not breaking. There is no code that needs fixing. Doing a migration after installing new features is rather normal I think.

sandergo90:
@Devolicious Upgrade guide added

PR #1492, opened 2017-04-03
[AllBundles]: fix ckeditor media tokens

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

When we introduced the new url chooser, we had some problems with the ckeditor and the media chooser. When you choose a media item in the ckeditor, the URL of it will be a token. This is our wished behaviour but the ckeditor does not render this token. 

With this PR I've added a Transformer that will do this correctly.

I've also changed all the old TextAreaTypes to the new WysiwygType because the transformer only works on this form type.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
+++ b/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Kunstmaan\AdminBundle\Form;
+
+use Symfony\Component\DomCrawler\Crawler;
+use Symfony\Component\Form\DataTransformerInterface;
+
+/**
+ * Class MediaTokenTransformer.
+ */
+class MediaTokenTransformer implements DataTransformerInterface
+{
+    /**
+     * @param mixed $content
+     *
+     * @return string
+     */
+    public function transform($content)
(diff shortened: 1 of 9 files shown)


sandergo90:
@Devolicious done

src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
```diff
@@ -0,0 +1,72 @@
...
+class MediaTokenTransformer implements DataTransformerInterface
+{
+    /**
+     * @param mixed $content
+     *
+     * @return string
+     */
+    public function transform($content)
+    {
+        $html = utf8_encode("<!DOCTYPE html>
```
mlebkowski: does this really work? the native utf8 support for PHP is rather shady. 

When using the crawler, I always use `addHTMLContent`, and provide the charset myself, or add a `<meta charset=utf8>` to the HTML string as in your case.

sandergo90: @mlebkowski it does work ;) But it's maybe a good solution to adapt it. I will do this later.

PR #1493, opened 2017-04-04
[KunstmaanCacheBundle]: add new caching bundle

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 1331

#  Using the Cache bundle

## Installation

This bundle is compatible with all Symfony 3.* releases. More information about installing can be found in this line by line walkthrough of installing Symfony and all our bundles, please refer to the [Getting Started guide](http://bundles.kunstmaan.be/getting-started) and enjoy the full blown experience.

## Usage

This bundle allows you to do stuff with caching. For now it's a starting bundle with the possibility
to ban stuff from varnish.

### Configure bundle in config.yml

This bundle works with the fos http cache bundle. Therefore you need to add the following configuration, of course with your own varnish path.
```YAML
fos_http_cache:
    proxy_client:
        varnish:
            servers:
                - 127.0.0.1:6081
    cache_manager:
        enabled: true
    invalidation:
        enabled: true
```

### Add the kunstmaan_cache routing to your routing.yml

```YAML
# KunstmaanCacheBundle
KunstmaanCacheBundle:
    resource: "@KunstmaanCacheBundle/Resources/config/routing.yml"
    prefix:   /{_locale}/
    requirements:
        _locale: "%requiredlocales%"
```
    
### Result

When you browse to "Settings" in the main menu, you will see that there is a new menu item available with the label of "Varnish ban".
There you can add a path that you would like to ban from varnish. When you check the all domains option and you are using a multidomain website,
the path will be banned from all hosts of your multidomain.

**On nodes you have a new menu actions to clear the cache for that node.**

If your varnish configuration does not support purging or banning yet, please take a look at the following configuration:

https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/doc/varnish-configuration.rst

### DIFF
--- a/UPGRADE-4.1.md
+++ b/UPGRADE-4.1.md
@@ -7,3 +7,11 @@ Added note to the RedirectBundle:
 https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1490
 
 Please be sure to execute database migrations after upgrading.
+
+## KunstmaanCacheBundle
+
+We added a new caching bundle.
+
+https://github.com/Kunstmaan/KunstmaanBundlesCMS/tree/master/docs/05-14-using-the-cache-bundle.md
+
+This bundle allows you to do stuff with caching. For now it's a starting bundle with the possibility to ban stuff from varnish.
--- a/docs/05-14-using-the-cache-bundle.md
+++ b/docs/05-14-using-the-cache-bundle.md
@@ -1,0 +1,45 @@
+#  Using the Cache bundle
+
+
+## Installation
+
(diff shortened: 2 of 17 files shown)


jockri:
What do you think about adding a "ban button" when editing a page (on same level als Publish, Delete, ...). In this way you can ban the url of the page you are editing without copy/pasting the url.

sandergo90:
That would be a nice expansion to this bundle. I think we should leave this functionality but indeed add a new varnish ban button in the menu.

Excellent idea!

sandergo90:
@jockry Done

sandergo90:
@Devolicious upgrade guide added

PR #1494, opened 2017-04-05
[KunstmaanNodeSearchBundle]: upgrade elastica to 5.1

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets |

# UPGRADE FROM 4.* to 5.0

## Elastica

The ruflin/elastica bundle has been upgraded to the latest 5.1 version.
This version is compatible with the latest elasticsearch version.

The only change that should be made is when you override the NodeSearcher that comes by default from the bundles:

**The setMinimumNumberShouldMatch function is now replaced by the setMinimumShouldMatch function**

When you have created some extra extensions on elastica you should read the changelog:

https://github.com/ruflin/Elastica/blob/master/CHANGELOG.md

### DIFF
--- a/UPGRADE-5.0.md
+++ b/UPGRADE-5.0.md
@@ -1,0 +1,14 @@
+# UPGRADE FROM 4.* to 5.0
+
+## Elastica
+
+The ruflin/elastica bundle has been upgraded to the latest 5.1 version.
+This version is compatible with the latest elasticsearch version.
+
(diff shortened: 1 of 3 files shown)


sandergo90:
It's a new feature so shouldn't it be pushed to master like now ?

sandergo90:
@Devolicious done

PR #1500, opened 2017-04-07
[MediaBundle] Create runtime config hash using the original image path

b-franco:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

There is an inconsistency when generating the hash in image paths when using a runtime config and the filter extension is not the same as the original image extension. This results in a hash mismatch in Liip\ImagineBundle\Controller\ImagineController. The hash needs to be made with the original image path to be consistent.

I don't think this solution is ideal, but it sketches the problem a bit. Any opinions/suggestions?

### DIFF
--- a/src/Kunstmaan/MediaBundle/Helper/Imagine/CacheManager.php
+++ b/src/Kunstmaan/MediaBundle/Helper/Imagine/CacheManager.php
@@ -2,6 +2,8 @@
 
 namespace Kunstmaan\MediaBundle\Helper\Imagine;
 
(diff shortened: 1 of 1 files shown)


sandergo90:
@b-franco I think it's indeed a good solution. @wesleylancel do you have an opinion on this?

wesleylancel:
@sandergo90 we've been using this for quite a while by overruling that class, works as intended :)

sandergo90:
Alright, then let's get this merged!

Issue #676, opened 2015-09-01
Migrate to Gulp instead of Assetic


jverdeyen:
Is it an option to build all the assets with a custom Gulpfile?
A main project Gulpfile can include all Gulpfile from specific bundles.

I'm currently using this custom Guplfile to pre-build the assets, as they will almost never change during developement.

https://gist.github.com/jverdeyen/ba8471307f2589a3d85b

Overruling templates an adding the static assets it trivial for example in `app/Resources/KunstmaanAdminBundle/views/Default/_css.html.twig`

```
<link rel="stylesheet" href="{{ asset('/css/admin.min.css') }}">
```

Can I open this discussion? Any suggestions? Any remarks?

vizo:
:+1:

roderik:
This is fine, we tried to keep the backend without extra gulp dependency, but in reality you always need it.

jverdeyen:
In order to remove the Assetic bundle, how do you guys see this implemented?
A gulpfile with settings/tasks per bundle and add css/js in `Resources/public`.
Pre-pack these files in the repo can speed up the CMS install process and lower the number of requirements.

wesleylancel:
Bumping this. Any suggestions or remarks from the Kunstmaan team? /cc @jockri @diskwriter

diskwriter:
@wesleylancel I discussed this with @jockri yesterday. It's on our roadmap. I'll keep you posted on the progress!

jverdeyen:
Nice! 🚀

Perni1984:
@diskwriter @jockri:
as this thread is fairly old, but has now been rescheduled to 4.1 - can we help anyhow?

I am using Kunstmaan as underlying base for a project of mine, which I needed to rewrite also on the backend quite a bit. The current assetic architecture makes this kind of hard (especially in conjunction with the bower dependencies in the different bundles) - I eventually need to rewrite the asset management part anyway to stay clean in my architecture. Any input from you how you would like to have that implemented? (e.g. how to manage the bower dependencies, etc.)

mlebkowski:
For an end user, KunstmaanBundles is just a library. They shouldn’t have to compile its assets, the lib should be shipped with them compiled already, in my opinion. BTW, the first thing I do in a new kunstmaan project is `assetic:dump`, commit the generated files to the repository and never do it again unless I update one of the bundles.

diskwriter:
[...]

PR #1510, opened 2017-05-16
[ALL] use gulp for cms assets & add compiled assets to public folders

dbeerten:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #676 

CMS assets can be compiled by gulp/groundcontrol now. The compiled assets are placed inside the public folder of the Bundle.

It's no longer required to run the `bin/console assetic:dump` command as the assets are already compiled. Simply running the `bin/console assets:install` command will copy the assets from each Bundle to the `web/` folder


mlebkowski:
How does this change the end-user workflow?

iainvdw:
A bit late to discuss this maybe, but seeing the release notes for v5.x I noticed the changes to the FE workflow and Gulp tasks. In our Kunstmaan v4.x setup we replaced the default Kunstmaan Gulp tasks with our customized Gulp setup which we use for a variety of different project, not just Kunstmaan. The v4.x didn't compile the CMS assets as they were installed via `assets:install` and `assetic:dump` Symfony CLI commands.

As there's not a lot of discussion in this ticket about how this change affects the root Gulp setup in Kunstmaan v5.x, I wonder how this change to the Kunstmaan Gulp setup affect our ability to use our own Gulp setup with custom tasks and directory structure?

PR #1517, opened 2017-06-12
[AdminBundle]: fix #1514

### LINKED ISSUE #1514: [AdminBundle] MediaTokenTransformer issue with TextPagePart
What is the logic behind the ```MediaTokenTransformer```?

Right now it is not working as expected with TextPagePart.

Scenario:

1, We select an image, the data-src is /uploads/media/592d25e1709db/big-jpg.jpg?token=[M144]
2, In the editor window it is 

``````

3, After save the ```MediaTokenTransformer``` set the src to be the token, so the img is going to be:

``````

Happens in https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php#L56

@sandergo90 @Devolicious do we missed something?

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1514

### DIFF
--- a/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
+++ b/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
@@ -25,8 +25,8 @@ class MediaTokenTransformer implements DataTransformerInterface
         $crawler->addHtmlContent($content);
 
         $crawler->filter('img')->each(
-            function (Crawler $node, $i) {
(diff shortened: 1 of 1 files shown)


wesleylancel:
@Devolicious @sandergo90 can this be merged into the 5.0 branch?

PR #1519, opened 2017-06-16
[SeoBundle]: change meta length

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

https://www.seoinc.com/seo-blog/google-increases-title-meta-description-length/

### DIFF
--- a/src/Kunstmaan/SeoBundle/Entity/Seo.php
+++ b/src/Kunstmaan/SeoBundle/Entity/Seo.php
@@ -8,7 +8,6 @@ use Kunstmaan\AdminBundle\Entity\AbstractEntity;
 use Kunstmaan\MediaBundle\Entity\Media;
 use Kunstmaan\SeoBundle\Form\SeoType;
 use Kunstmaan\UtilitiesBundle\Helper\ClassLookup;
-use Symfony\Component\Validator\Constraints as Assert;
 
 /**
  * Seo metadata for entities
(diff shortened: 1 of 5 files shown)


mlebkowski:
It feels shortsighted to limit the field to what google thinks is useful. It’s called `meta_title`, not `google_title`, there are other search engines and robots out there… I suggest a length of 255.

sandergo90:
That's even better indeed. Maybe we should add a question mark next to the input field with some more information ?

sandergo90:
@mlebkowski i've checked some resources and all common search engines limit their meta title around 70 characters.

https://www.sagerock.com/blog/title-tag-meta-description-length/

So I would suggest to set the limit to 70 characters. 255 is to much for a meta title.

mlebkowski:
Why should a CMS limit this? If I wanted to use 71 characters, because I wanted to end a sentence, I could not do it?

sandergo90:
Real clients who are using the CMS want to have a limitation on it. The people who are doing the content input do not know how many characters they can enter. I think by limiting this, it will be clearer.

mlebkowski:
Your argument suggests that I am not a real client. ;) Wouldn’t not limiting the field in a database and adding a `max_length` config value with a sane default be a good compromise?

sandergo90:
Yes I call us Developer ;) So we remove the assert on the entity and add a constraint in the form type that has a max_length value coming from the config ?

mlebkowski:
yes, sandergo, that would be the idea. but as much as I think about it, the more I’m convinced it isn’t worth the effort :)

sandergo90:
@mlebkowski I was just thinking the same 😆 . Maybe we can remove the assert on the entity and leave it at 70 on the form type and add a constraint there. When someone wants to override this, they can override the form type.

Issue #1523, opened 2017-06-20
Translating missing strings


mablae:
Hello Team, 

I would love to help translating missing strings into german. 
Is there a standard procedure to follow or even an online tool like weblate? 

How can I contribute? Ping @jockri

mablae:
Okay, I have now started translating missing strings to german language. 

Should I provide a PR on that?

Devolicious:
Hi @mablae,

Yes you are very welcome to do so. To submit a PR you can just follow our [contribution guide](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/docs/99-02-pull-requests.md)

mablae:
Hey @Devolicious , 

I am already half done. 😄  I am using `jms/translation-bundle` to extract the strings. 

Quick question: Which branch should I use as a base? It's kind of a bugfix, isn't it? 

I am currently using the `v4.0.2`

Devolicious:
Hi @mablae,

It's fair to say that missing translations are a bugfix :) So for that you should use the `4.0` branch as the base for your PR.

sandergo90:
As PR #1651 says, the translation will be soon available with crowdin. If you want to add more german translations, they can be added there.

Issue #1102, opened 2016-04-07
Translations not complete in admin media


wouterv55:
Some parts stay untranslated when using the admin media folders in another language. (e.g. in the 'add subfolder' feature)

PR #1532, opened 2017-07-04
[WIP] Translation Fixes DE

mablae:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #1523 #1102 

Please do not merge yet - Work in Progress

### DIFF
--- a/src/Kunstmaan/MediaBundle/Resources/translations/messages.de.yml
+++ b/src/Kunstmaan/MediaBundle/Resources/translations/messages.de.yml
@@ -1,84 +1,153 @@
-# Media
 media:
-  aviary:
-    title: Aviary
-    nothing: Hier gibt es nichts zu sehen.
-
-  media:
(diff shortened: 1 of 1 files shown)


mablae:
@Devolicious 

I used the jms/translation-bundle to extract missing strings. However I noticed a lot of strings that are not translations keys but "english Words" like, "View", "Delete" missing any semantics. 

How should we address this issue?

Devolicious:
@mablae in that case I think that we need some new generic translations that exist in the AdminBundle.

sandergo90:
@mablae Can you check this again ? Because I think the english words are already replaced by translation keys

sandergo90:
@mablae As PR https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1651 says, the translation will be soon available with crowdin. If you want to add more german translations, they can be added there.

PR #1534, opened 2017-07-05
[NodeBundle, AdminListBundle]: fix new introduced abstractPageAdminList

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


mlebkowski:
Adding an abstract method looks like a breaking change

sandergo90:
True but this abstract class is a new feature and does not work without this modification.

PR #1540, opened 2017-07-07
[MultiDomainBundle] Add option to disable host override notice

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | n/a

BC: any class that extends the current HostOverrideListener will need the extra parameter injected.


mlebkowski:
You can instead make a compiler pass to remove the listener altogether if the switch is set to disabled. This would mean there is no code change, and no BC break

b-franco:
Resolved by removing the definition by using a compilerpass. This can be closed. @Devolicious @jockri

Issue #1547, opened 2017-07-10
[Node][NodeSearch] Child nodes not reindexed after changing parent url


dannyvw:
After changing the parent node url, all childs have still the old url indexed (only affected on search page).

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php#L47

Possible solution, change `$nodeSearchConfiguration->indexNodeTranslation($event->getNodeTranslation(), true);` to `$nodeSearchConfiguration->indexNode($event->getNode(), $event->getNodeTranslation()->getLang());`

treeleaf:
I have modified this in an pull request: https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1550

Numkil:
Closing because referenced PR is merged

Issue #1509, opened 2017-05-11
StructureNode isOnline=false breaks search indexing


mwoynarski:
As the search bundle is configured to ignore any offline page, and its children, the default `isOnline=false` in [StructureNode](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/NodeBundle/Entity/StructureNode.php#L17) prevents the search bundle from indexing any page that is a child of a structure node.

This can be remedied easily enough by changing the return value for isOnline() to `true` in either StructureNode itself or any of your own classes that extend it, but I'm hesitant to globally make that change in a PR because it may have repercussions beyond my own use cases.

I do consider this a bug, though, because the default behaviour renders the search bundle completely unusable.

Devolicious:
Hi @mwoynarski,

Thanks for reporting this bug. We'll look into it.

treeleaf:
I have created a pull request : https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1550

PR #1550, opened 2017-07-13
[NodeSearchBundle] Bugfix for indexing StructuredNodes and children

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1509, #1547, #1606

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
+++ b/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
@@ -15,7 +15,6 @@ use Kunstmaan\NodeBundle\Entity\PageInterface;
 use Kunstmaan\NodeBundle\Helper\RenderContext;
 use Kunstmaan\NodeSearchBundle\Event\IndexNodeEvent;
 use Kunstmaan\NodeSearchBundle\Helper\IndexablePagePartsService;
-use Kunstmaan\NodeSearchBundle\Helper\SearchBoostInterface;
 use Kunstmaan\NodeSearchBundle\Helper\SearchViewTemplateInterface;
 use Kunstmaan\PagePartBundle\Helper\HasPagePartsInterface;
 use Kunstmaan\SearchBundle\Configuration\SearchConfigurationInterface;
@@ -45,7 +44,7 @@ class NodePagesConfiguration implements SearchConfigurationInterface
     protected $searchProvider;
 
     /** @var array */
-    protected $locales = array();
+    protected $locales = [];
 
     /** @var array */
     protected $analyzerLanguages;
@@ -54,7 +53,7 @@ class NodePagesConfiguration implements SearchConfigurationInterface
     protected $em;
 
     /** @var array */
-    protected $documents = array();
+    protected $documents = [];
 
     /** @var ContainerInterface */
     protected $container;
@@ -71,10 +70,14 @@ class NodePagesConfiguration implements SearchConfigurationInterface
     /** @var DomainConfigurationInterface */
     protected $domainConfiguration;
 
-    private $properties = [];
+    /** @var array */
+    protected $properties = [];
 
     /** @var Node */
-    private $currentTopNode = null;
+    protected $currentTopNode = null;
+
+    /** @var array */
+    protected $nodeRefs = [];
 
     /**
      * @param ContainerInterface      $container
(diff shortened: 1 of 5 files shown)


src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
```diff
@@ -245,12 +246,15 @@ public function indexNodeTranslation(NodeTranslation $nodeTranslation, $add = fa
             return false;
         }
 
+        // Retrieve the referenced entity from the public NodeVersion
+        $page = $publicNodeVersion->getRef($this->em);
+
         $node = $nodeTranslation->getNode();
         if ($this->isIndexable($page)) {
```
sandergo90: @treeleaf Maybe use the ReflectionClass here to ? And only when it's indexable you retrieve the ref ?

treeleaf: Ok, I did oversee that part. I have modified it.

src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
```diff
@@ -683,4 +687,21 @@ protected function getAclPermissions($object)
...
+     * @param $publicNodeVersion
+     *
+     * @return mixed
+     */
+    private function getNodeRefPage(NodeVersion $publicNodeVersion)
+    {
+        $refEntityName = $publicNodeVersion->getRefEntityName();
+
+        if (!isset($this->nodeRefs[$refEntityName])) {
+            $refEntityClass = new \ReflectionClass($refEntityName);
```
Devolicious: Reflection class seems to be obsolete here. Just `new $refEntityName()` should do the trick in this case.

src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php
```diff
@@ -44,7 +44,9 @@ public function onPostPersist(NodeEvent $event)
     private function index(NodeEvent $event)
     {
         $nodeSearchConfiguration = $this->container->get('kunstmaan_node_search.search_configuration.node');
-        $nodeSearchConfiguration->indexNodeTranslation($event->getNodeTranslation(), true);
+        $nodeTranslation = $event->getNodeTranslation();
+        $nodeSearchConfiguration->indexNodeTranslation($nodeTranslation, true);
+        $nodeSearchConfiguration->indexChildren($event->getNode(), $nodeTranslation->getLang());
```
mlebkowski: what is the purpose of this addition?

treeleaf: This is to address issue #1547; to reindex children when the parent url changes for instance.

mlebkowski: Wouldn’t this reindex the whole site if you save the home page?

dannyvw: Yes that's right

mlebkowski: Are we sure we want to do this? It takes over a dozen of seconds to do this on my production machine, not to mention on the dev server and unnecessary load on elasticsearch, etc.

Devolicious: I think that we can solve this with a quick check to see if the URL has changed and if it has, than reindex it's children.

mlebkowski: Can we please think about abstracting it into some kind of `ChildrenNeedReindexingDecisionMakerInterface`? I have nodes that index elements of their parents, so this kind of behaviour would apply there.

mlebkowski: Disregard my last. It’s on an event either way, so any additional logic can be attached in client bundles.

PR #1552, opened 2017-07-20
[Docs] Update list of generated classes

svenluijten:
This list wasn't up to date with what classes are generated, and where they're generated. This PR fixes that.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

PS: This may concern more versions than just 4.0 (which this PR is for), so `master` might be a better fit.


svenluijten:
Failing CI doesn't appear to be this PR's fault, unless I'm missing something obvious?

Numkil:
@svenluijten No worries it is not your fault. Next release will fix this issue. (deprecated vendor not installing)

PR #1556, opened 2017-07-21
[AdminListBundle] refactor and modernize exportservice

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 

Replace legacy method of returning array with available outputs
Remove template for rendering with csv's since box/spout can do it. Since we have a a way of rendering templates for specific columns in excel we can use these for csv and ods as well. No need for seperate implementations.
1 method 1 way for each output
refactoring legacy style coding

- [x] refactor code
- [x] rewrite tests
- [x] icon for ods export

BC breaks: Since new option has been added to excelservice (ods) it is necessary to fix all adminlistcontrollers that have the old format requirement regex. ods is not in that list yet.

old:
```php
    /**
     * @Route("/export.{_format}", requirements={"_format" = "csv|xlsx"}, name="{{ bundle_name|lower }}_admin_bike_export")
     * @Method({"GET", "POST"})
     * @return array
     */
    public function exportAction(Request $request, $_format)
    {
	return parent::doExportAction($this->getAdminListConfigurator(), $_format, $request);
    }
```

new:
```php
    /**
     * @Route("/export.{_format}", requirements={"_format" = "csv|xlsx|ods"}, name="{{ bundle_name|lower }}_admin_bike_export")
     * @Method({"GET", "POST"})
     * @return array
     */
    public function exportAction(Request $request, $_format)
    {
	return parent::doExportAction($this->getAdminListConfigurator(), $_format, $request);
    }
```


Devolicious:
@Numkil This is PHP7 code only. So basically this will be something for version 6.0

sandergo90:
@Devolicious can you dubble review ?

PR #1563, opened 2017-07-28
[RedirectBundle] Prevent redirect loop with trailing slash

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When removing trailing slash (https://symfony.com/doc/current/routing/redirect_trailing_slash.html) and a customer fills in an url with a trailing slash you will get a redirect loop.

### DIFF
--- a/src/Kunstmaan/RedirectBundle/Router/RedirectRouter.php
+++ b/src/Kunstmaan/RedirectBundle/Router/RedirectRouter.php
@@ -119,7 +119,7 @@ class RedirectRouter implements RouterInterface
             if ($redirect->getDomain() == $domain || !$redirect->getDomain()) {
                 $this->routeCollection->add(
                     '_redirect_route_' . $redirect->getId(),
-                    new Route($redirect->getOrigin(), array(
+                    new Route(rtrim($redirect->getOrigin(), '/'), array(
                         '_controller' => 'FrameworkBundle:Redirect:urlRedirect',
                         'path' => $redirect->getTarget(),
                         'permanent' => $redirect->isPermanent(),


sandergo90:
@dannyvw can you start from 4.0 branch ? Thanks!

sandergo90:
@dannyvw The goal of your PR is clear and it's indeed a good fix but shouldn't we also create the redirect controller for the trailing slashes then as described in the symfony documentation?

dannyvw:
@sandergo90 You can add the controller in your own project because it's possible that not everyone want this.

sandergo90:
I understand, but if we merge this PR, and we add a redirect from /nl/ to /nl/foo, it will not work because the slash is trimmed ?

dannyvw:
Not sure if this is an problem. I need to test that. But i can add the controller if you want.

sandergo90:
@dannyvw If you can create the controller then this PR can certainly be merged. Thanks!

dannyvw:
@sandergo90 Rebase is done. Controller is not added yet, because of this "Make sure to include this route in your routing configuration at the very end of your route listing. Otherwise, you risk redirecting real routes (including Symfony core routes) that actually do have a trailing slash in their path.". How do you want to solve this?

sandergo90:
@dannyvw probably the best way to do this is to create a new router for this. The router can be given a very low priority so it will be executed at the very end. You can take a look at the SlugRouter and maybe start from there.

dannyvw:
@sandergo90 Yes that is possible, but i don't have the time for that now. Maybe someone else can build this?

sandergo90:
@dannyvw Yes we will check if someone has the time to do this. Thanks!

sandergo90:
Added to the roadmap https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/1648

PR #1587, opened 2017-09-08
[AdminBundle] Allow easier form theming for tabs widget

mandalornl:
This way you'll be able override/extend the form theme by extending the twig file and adding a `form_theme` tag inside the `block`. Instead of just copy+pasting the whole template.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


sandergo90:
Hi @mandalornl,

Can you give a specific use case for this ? Because you would also have to copy the whole file and override the block content.  We don't see the advantage of this right now. 

Cheers

Devolicious:
Hi @mandalornl,

If you have a specific use case where this is needed please let us know, as for now, I'm gonna close this PR. If needed, we can reopen this later.

Thanks for the contribution!

PR #1588, opened 2017-09-10
[WIP][GeneratorBundle] use plyr as the video player for the VideoPagePart

NindroidX:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

Use [plyr](https://plyr.io/) as the video player for the VideoPagePart. This makes the player consistent for each supported video type, and makes the controls themeable. Playing self hosted videos is also possible now.

**Important**: Dailymotion videos are not supported by plyr. Is Dailymotion support still needed?

**Update**: Dailymotion is still needed.

**Update 2** Plyr seems to be having some bugs which make implementing dataLayers a chore. Combined with the lack of Dailymotion support and plyr not being very easy to extend, it's probably a better idea to use Video.js instead (which has a plugin system). 

- [x] Replace the current VideoPagePart code with plyr support
- [x] Add Dailymotion support
- [ ] Replace Plyr with Video.js


sandergo90:
@NindroidX Can you finish this feature ?

PR #1593, opened 2017-09-14
[GeneratorBundle] Replace all rem values with px

NindroidX:
It was decided to stop using rem as a unit and replace it with px as the main unit (and use the other units where it makes sense).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none


FVKVN:
👍

FVKVN:
@Devolicious

PR #1612, opened 2017-10-04
[GeneratorBundle] Use livingcss to build styleguide (remove hologram & gems)

dbeerten:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

- Removed hologram & added [livingcss](https://github.com/straker/livingcss)
- Removed gems
- Removed babel-preset-es2015, use env instead ([http://babeljs.io/env](url))

With livingcss we can write our styleguide html elements in seperate files, so we don't need to code html in comments of the scss file.

Theme of the styleguide got an update:
image


FVKVN:
👍 @Devolicious

janb87:
@dbeerten Build is broken. I've recently updated the travis file to build also the frontend.

Should be fixed by removing line 26 and 27 in the travis.yml
https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/5.0/.travis.yml#L26

dbeerten:
@janb87 fixed, thanks!

PR #1646, opened 2017-11-14
[NodeSearchBundle] shards and replicas configurable

bakie:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | /

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
+++ b/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
@@ -73,6 +73,12 @@ class NodePagesConfiguration implements SearchConfigurationInterface
 
     /** @var array */
     protected $properties = [];
+    
+    /** @var integer */
+    protected $numberOfShards;
+
+    /** @var integer */
+    protected $numberOfReplicas;
 
     /** @var Node */
     protected $currentTopNode = null;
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
```diff
@@ -71,6 +71,12 @@ class NodePagesConfiguration implements SearchConfigurationInterface
     /** @var DomainConfigurationInterface */
     protected $domainConfiguration;
 
+    /** @var integer */
+    protected $numberOfShards;
```
treeleaf: initialize values to 4 and 1 (to maintain BC, when the config is not changed)?

bakie: This is not the BC. The BC is when you extend this class and call the __construct function. There are no default values for the shards and replicas.
The values 4 and 1 are in a single node setup unnecessary as there is only one node available.

Numkil: Can we not prevent this BC break by writing the constructor for this class in this way? Then we add the posibility for adding them when extending this class but older files do not need to be updated because of this change. 
```php
public function __construct($container, $searchProvider, $name, $type, $numberOfShards = 1, $numberOfReplicas = 0)
{
}
```

bakie: @Numkil I was also thinking about that, but then if you override the parameters but forget to change it in your class you get the value 1 and 0.

@Devolicious what do you think. Should I add the default values to prevent the BC break or not ?

Devolicious: IMHO I would rewrite the whole thing and get rid of the container aswell :) But if we want it in the master then we have to set default values. We can do a 2 stage PR and submit this with some deprecations and merge it into 5.0 and there we fix it how we want it.

PR #1666, opened 2017-11-20
[UtilitiesBundle] Feature transliterator

cv65kr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | improvment
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | no

I think we should use `Behat/Transliterator` is more accurate. It's made some BC breaks, but the second argument in slugify function never been used.

### DIFF
--- a/src/Kunstmaan/UtilitiesBundle/Helper/Slugifier.php
+++ b/src/Kunstmaan/UtilitiesBundle/Helper/Slugifier.php
@@ -13,11 +13,12 @@ final class Slugifier implements SlugifierInterface
      * Slugify a string
      *
      * @param string $text
+     * @param string $delimiter
      *
      * @return string
      */
-    public function slugify($text)
+    public function slugify($text, $delimiter = '-')
     {
-        return Transliterator::transliterate($text);
+        return Transliterator::transliterate($text, $delimiter);
     }
 }
--- a/src/Kunstmaan/UtilitiesBundle/Helper/Slugifier.php
+++ b/src/Kunstmaan/UtilitiesBundle/Helper/Slugifier.php
@@ -2,41 +2,22 @@
 
 namespace Kunstmaan\UtilitiesBundle\Helper;
 
+use Behat\Transliterator\Transliterator;
+
 /**
  * Sulgifier is a helper to slugify a certain string
  */
-class Slugifier implements SlugifierInterface
+final class Slugifier implements SlugifierInterface
 {
     /**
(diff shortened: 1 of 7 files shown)


src/Kunstmaan/UtilitiesBundle/Helper/Slugifier.php
```diff
@@ -2,41 +2,22 @@
...
      * Slugify a string
      *
-     * @param string $text    Text to slugify
-     * @param string $default Default return value (override when slugify would return an empty string)
+     * @param string $text
      *
      * @return string
      */
-    public function slugify($text, $default = '', $replace = array("'"), $delimiter = '-')
+    public function slugify($text)
```
sandergo90: We can keep the delimiter here. The transliterate function https://github.com/Behat/Transliterator/blob/master/src/Behat/Transliterator/Transliterator.php#L431 also has a seperator as argument.

src/Kunstmaan/UtilitiesBundle/Helper/SlugifierInterface.php
```diff
@@ -12,11 +12,8 @@
 
     /**
      * @param $text
-     * @param $default
-     * @param $replace
-     * @param $delimiter
      * @return mixed
      */
-    public function slugify($text, $default = 'n-a', $replace = array("'"), $delimiter = '-');
+    public function slugify($text);
```
sandergo90: Keep delimiter here to

cv65kr:
@sandergo90 done.

krewetka:
I just wanted to let you know that replacing previous slugifer using https://www.php.net/manual/en/transliterator.transliterate.php with behat caused arabic transliteration to change.

From what I see previous one was correct and new one is not.

Example:

الستائر - earlier transliterated to `alstayr`  not is transliterated to `lstyr` and in general all texts with this character https://www.fileformat.info/info/unicode/char/0627/index.htm

https://translate.google.com translated it to `alsatayir` ( even ore more `a`) but in general this character I mentioned (0627) should be transliterated to A -> https://en.wikipedia.org/wiki/Romanization_of_Arabic

and behat transliterate replace it with empty character

acrobat:
Hi @krewetka,  can you create a new issue for this bug and reference this PR? It makes it easier to keep track of the bug!

Can you also report this issue to behat/transliterator so it possibly can be fixed upstream? Thanks!

PR #1674, opened 2017-11-22
[SensiolabsInsight] Throw AccessDeniedException instead of AccessDeniedHttpException

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no-ish (different exception class as before)
| Deprecations? | no
| Fixed tickets | 

Fixes the sensiolabs insight error "Symfony applications should not throw AccessDeniedHttpException". Not sure if this is a "real" BC break, if so I can rebase this against 5.0!


deZinc:
@Devolicious I'd consider this a BC break as it's possible that users are catching the exceptions or using them in unittests. If so, this should go into the 5.0 branch. What's your opinion on this?

acrobat:
You're right, let's just put this in 5.0.

Issue #1683, opened 2017-11-22
[Insight] Unused method, property, variable or parameter - in src/Kunstmaan/…/Helper/URLHelper.php, line 47


delboy1978uk:
**[in src/Kunstmaan/NodeBundle/Helper/URLHelper.php, line 47](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/f0922510bb34c0da89f4534a3788ddb5ff17847c/src/Kunstmaan/NodeBundle/Helper/URLHelper.php#L47)**

> This `requestStack` attribute is declared but never used. You should remove it.

```php
    private $mediaMap = null;

    /**
     * @var RequestStack
     */
    private $requestStack;

    /**
     * @var DomainConfigurationInterface
     */
    private $domainConfiguration;
```

---------------------

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/34aece18-3640-47ea-8e9f-96b79919905b/analyses/373)_

PR #1684, opened 2017-11-22
[NodeBundle] unused class property

delboy1978uk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/1683

### DIFF
--- a/src/Kunstmaan/NodeBundle/Helper/URLHelper.php
+++ b/src/Kunstmaan/NodeBundle/Helper/URLHelper.php
@@ -42,11 +42,6 @@ class URLHelper
     private $mediaMap = null;
 
     /**
-     * @var RequestStack
-     */
-    private $requestStack;
-
-    /**
      * @var DomainConfigurationInterface
      */
     private $domainConfiguration;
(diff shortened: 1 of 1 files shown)


delboy1978uk:
If anyone instantiates this manually though, would that not create a BC break? I realise autowiring will handle it, but are we sure we always get that the magic Symfony way? Other people using bundles might have said `$x = new UrlHelper($etc, $etc, $oopsBrokeIt, $etc);`

deZinc:
@delboy1978uk Yup, that makes it a BC break indeed. So the change has to be added to 5 branch and also added to the upgrade guide.

delboy1978uk:
In it's current state it can be merged into master with no BC break, and insight will stop complaining. 
I can make another ticket and break BC by removing it fully and pull request on 5.0 😀

deZinc:
Let's keep it clean though. :-p
Either we remove everything, or nothing... I wouldn't be surprised if we merge it like this that insight will complain about an unused param in the configurator.

delboy1978uk:
haha you could be right! ok i'll rebase and push again

delboy1978uk:
rebased to 5.0 😀

Issue #1685, opened 2017-11-22
[Insight] User specific files should not appear in .gitignore - in .gitignore, line 6


delboy1978uk:
**[in .gitignore, line 6](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/f0922510bb34c0da89f4534a3788ddb5ff17847c/.gitignore#L6)**

> `.idea` is user-specific and should not appear in a project `.gitignore`. Consider adding it to the user global `.gitignore` instead.

```
phpunit.xml
composer.lock
composer.phar
autoload.php
/vendor/
.idea
.DS_Store
npm-debug.log
*~

```

---------------------

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/34aece18-3640-47ea-8e9f-96b79919905b/analyses/373)_

delboy1978uk:
In order to start using Global .gitignore, we need to go through these steps:
```
    touch ~/.gitignore_global - this will create global .gitignore file in your home directory.
    git config --global core.excludesfile ~/.gitignore_global .
```

PR #1686, opened 2017-11-22
[SensiolabsInsight] - remove user .gitignore files

delboy1978uk:
Should be in global gitignore. Create and use it like so:
```
touch ~/.gitignore_global
git config --global core.excludesfile ~/.gitignore_global .
```
Then add .idea to the file

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/1685

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ composer.lock
 composer.phar
 autoload.php
 /vendor/
-.idea
 .DS_Store
 npm-debug.log
 *~
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,5 @@ composer.lock
 composer.phar
 autoload.php
 /vendor/
-.DS_Store
 npm-debug.log
 *~


delboy1978uk:
ah I didn't see that one! Ok give me two minutes!

delboy1978uk:
😑 lol, ok, i'll rebase this too!

mlebkowski:
So SensioInsights is ignoring the fact that about half of the developers ever checking out this repo will use either PHPStorm or a mac, hence having both of those files in the repo? So instead making their life easier by having those „user specific” entries, you instead want every dev to have a custom setup step? In my opinion this insight is stupid. It doesn’t hurt anyone to have those two lines there.

OTOH, every mac user should have `.DS_Store` in their global ignore, as well as every PHPStorm user probably has `.idea` there… ;) 

As for your snippet, I don’t see a reason to rename the global ignore file. You can just do this instead:

```
grep ".idea" ~/.gitignore >/dev/nul || echo .idea | tee -a ~/.gitignore 
grep ".DS_Store" ~/.gitignore >/dev/nul || echo .DS_Store | tee -a ~/.gitignore 
```

So if a user already have their `~/.gitignore` (the default path), you can just append your entries if they don’t exist.

delboy1978uk:
I like being verbose about my files and classnames etc, hence the _global. Just a suggestion! 
I guess most people here are devs smart enough to realise that ~/.gitignore would be user specific and global. :-)

sandergo90:
@mlebkowski welcome back ! 😉 It's maybe a stupid suggestion of Insights but I can see why. It's a user specific think so it shouldn't be in our project IMHO.

mlebkowski:
hi @sandergo90, I never went anywhere, I were just lurking ;)

@delboy1978uk, ok I agree that the devs should be smart enough to interpret. point taken.

PR #1692, opened 2017-11-22
[AllBundles] Code style fixes

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Some general code style fixes as detected by phpstorm. Each commit is a specific code inspection fixed

### DIFF
--- a/src/Kunstmaan/FixturesBundle/Builder/PageBuilder.php
+++ b/src/Kunstmaan/FixturesBundle/Builder/PageBuilder.php
@@ -220,7 +220,7 @@ class PageBuilder implements BuilderInterface
         $translationWithSameUrl = $this->nodeTranslationRepo->getNodeTranslationForUrl($translation->getUrl(), $translation->getLang(), false, $translation);
 
         if ($translationWithSameUrl instanceof NodeTranslation) {
-            $translation->setSlug($this->slugifier->slugify($this->incrementString($translation->getSlug())));
+            $translation->setSlug($this->slugifier->slugify(static::incrementString($translation->getSlug())));
             $this->ensureUniqueUrl($translation, $page);
         }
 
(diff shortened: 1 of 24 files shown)


src/Kunstmaan/FixturesBundle/Builder/PageBuilder.php
```diff
@@ -220,7 +220,7 @@ private function ensureUniqueUrl(NodeTranslation $translation, HasNodeInterface
         $translationWithSameUrl = $this->nodeTranslationRepo->getNodeTranslationForUrl($translation->getUrl(), $translation->getLang(), false, $translation);
 
         if ($translationWithSameUrl instanceof NodeTranslation) {
-            $translation->setSlug($this->slugifier->slugify($this->incrementString($translation->getSlug())));
+            $translation->setSlug($this->slugifier->slugify(static::incrementString($translation->getSlug())));
```
Numkil: This method has no reason to be static in the first place. I would suggest changing the method definition instead of this line. It is private anyway so you always need an instantiated object of this class so the static adjective is useless.

src/Kunstmaan/NodeBundle/EventListener/NodeTranslationListener.php
```diff
@@ -345,7 +345,7 @@ private function ensureUniqueUrl(
             $oldUrl = $translation->getFullSlug();
             $translation->setSlug(
                 $this->slugifier->slugify(
-                    $this->incrementString($translation->getSlug())
+                    static::incrementString($translation->getSlug())
```
Numkil: This method has no reason to be static in the first place. I would suggest changing the method definition instead of this line. It is private anyway so you always need an instantiated object of this class so the static adjective is useless.

acrobat:
Fixed the comments!

Issue #1714, opened 2017-11-29
MultiDomainAdminTwigExtension::switchedHostIsCurrent() always false


delboy1978uk:
```php
return $this->domainConfiguration->getHost() == $this->getSwitchedHost()['host'];
```

`getSwitchedHost()` returns a string:
```php
return $this->domainConfiguration->getHostSwitched();
```

The domain configuration interface returns a string. All we need to do is remove the array part from the end for this to start working.

Numkil:
Closing because PR merged.

PR #1715, opened 2017-11-29
[AdminBundle] - Fix MultiDomainAdminTwigExtension::switchedHostIsCurr…

delboy1978uk:
…ent()

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature? (Improvement) | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/1714

### DIFF
--- a/src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
+++ b/src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
@@ -3,8 +3,11 @@
 namespace Kunstmaan\AdminBundle\Twig;
 
 use Kunstmaan\AdminBundle\Helper\DomainConfigurationInterface;
+use Twig_Environment;
+use Twig_Extension;
+use Twig_SimpleFunction;
 
-class MultiDomainAdminTwigExtension extends \Twig_Extension
+class MultiDomainAdminTwigExtension extends Twig_Extension
 {
     /**
      * @var DomainConfigurationInterface
@@ -24,37 +27,35 @@ class MultiDomainAdminTwigExtension extends \Twig_Extension
      */
     public function getFunctions()
     {
-        return array(
-            new \Twig_SimpleFunction('multidomain_widget', array($this, 'renderWidget'), array('needs_environment' => true, 'is_safe' => array('html'))),
-            new \Twig_SimpleFunction('is_multidomain_site', array($this, 'isMultiDomainSite')),
-            new \Twig_SimpleFunction('get_switched_host', array($this, 'getSwitchedHost')),
-            new \Twig_SimpleFunction('switched_host_is_current', array($this, 'switchedHostIsCurrent')),
-        );
+        return [
+            new Twig_SimpleFunction('multidomain_widget', [$this, 'renderWidget'], ['needs_environment' => true, 'is_safe' => ['html']]),
+            new Twig_SimpleFunction('is_multidomain_site', [$this, 'isMultiDomainSite']),
+            new Twig_SimpleFunction('get_switched_host', [$this, 'getSwitchedHost']),
+            new Twig_SimpleFunction('switched_host_is_current', [$this, 'switchedHostIsCurrent']),
+        ];
     }
 
     /**
      * Render multidomain switcher widget.
      *
-     * @param \Twig_Environment $env
-     * @param array             $locales    The locales
+     * @param Twig_Environment $env
      * @param string            $route      The route
      * @param array             $parameters The route parameters
      *
      * @return string
      */
-    public function renderWidget(\Twig_Environment $env, $route, array $parameters = array())
+    public function renderWidget(Twig_Environment $env, $route, array $parameters = [])
     {
         $template = $env->loadTemplate(
             "@KunstmaanAdmin/MultiDomainAdminTwigExtension/widget.html.twig"
         );
 
         return $template->render(
-            array_merge(
-                $parameters,
-                array(
+            \array_merge(
+                $parameters, [
                     'hosts' => $this->getAdminDomainHosts(),
                     'route' => $route
-                )
+                ]
             )
         );
     }
@@ -70,7 +71,7 @@ class MultiDomainAdminTwigExtension extends \Twig_Extension
     }
 
     /**
-     * @return string
+     * @return array|null
      */
     public function getSwitchedHost()
     {
@@ -78,15 +79,21 @@ class MultiDomainAdminTwigExtension extends \Twig_Extension
     }
 
     /**
-     * @return string
+     * @return bool
      */
     public function switchedHostIsCurrent()
     {
-        return $this->domainConfiguration->getHost() == $this->getSwitchedHost()['host'];
+        $hostInfo = $this->getSwitchedHost();
+        return $this->domainConfiguration->getHost() === $hostInfo['host'];
     }
 
+    /**
+     * @return array
+     */
     public function getAdminDomainHosts()
     {
         return $this->domainConfiguration->getHosts();
     }
+    
 }
+
--- a/src/Kunstmaan/MultiDomainBundle/Helper/DomainConfiguration.php
+++ b/src/Kunstmaan/MultiDomainBundle/Helper/DomainConfiguration.php
@@ -233,7 +233,7 @@ class DomainConfiguration extends BaseDomainConfiguration
     }
 
     /**
-     * @return string|null
+     * @return array
      */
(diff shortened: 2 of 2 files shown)


src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
```diff
@@ -25,10 +28,10 @@ public function __construct(DomainConfigurationInterface $domainConfiguration)
     public function getFunctions()
     {
         return array(
```
cv65kr: ```array``` => ```[]``` :D

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
```diff
@@ -42,7 +45,7 @@ public function getFunctions()
      *
      * @return string
      */
-    public function renderWidget(\Twig_Environment $env, $route, array $parameters = array())
+    public function renderWidget(Twig_Environment $env, $route, array $parameters = array())
```
cv65kr: here

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
```diff
@@ -78,13 +81,16 @@ public function getSwitchedHost()
     }
 
     /**
-     * @return string
+     * @return bool
      */
     public function switchedHostIsCurrent()
     {
-        return $this->domainConfiguration->getHost() == $this->getSwitchedHost()['host'];
+        return $this->domainConfiguration->getHost() == $this->getSwitchedHost();
```
cv65kr: ```==``` => ```===```

delboy1978uk: good idea! give me a minute :-)

cv65kr:
Also You forgot about extra line for Insight

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
cv65kr: The minimal version of branch 4.0 is PHP5.6 so PHP5.3 array notation is not necessary.

delboy1978uk: replaced!

delboy1978uk:
Bizarre - I definitely have the extra line at the bottom of my file locally, yet when I push, it's gone! I'll add yet another line and hopefully that should fix it!

cv65kr:
You forgot about others array notations ;) 
```
    public function getFunctions()
    {
        return [
            new Twig_SimpleFunction('multidomain_widget', [$this, 'renderWidget'], ['needs_environment' => true, 'is_safe' => ['html']]),
            new Twig_SimpleFunction('is_multidomain_site', [$this, 'isMultiDomainSite']),
            new Twig_SimpleFunction('get_switched_host', [$this, 'getSwitchedHost']),
            new Twig_SimpleFunction('switched_host_is_current', [$this, 'switchedHostIsCurrent']),
        ];
    }
```

If I good remember the Insight required also extra line before closing class tag ```}```

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
cv65kr: You forgot about this ```array```.

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
```diff
@@ -42,19 +45,18 @@ public function getFunctions()
...
      */
-    public function renderWidget(\Twig_Environment $env, $route, array $parameters = array())
+    public function renderWidget(Twig_Environment $env, $route, array $parameters = [])
     {
         $template = $env->loadTemplate(
             "@KunstmaanAdmin/MultiDomainAdminTwigExtension/widget.html.twig"
         );
 
         return $template->render(
             array_merge(
```
cv65kr: Maybe also use function with namespace is better choice?
[https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/](https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/)

delboy1978uk: I think that's unnecessary in this case as we have explicitly declared which class it is in the use statements at the top.  I think that just applies to functions rather than classes.

delboy1978uk: Just spoke with Ocramius (my friend Marco, the guy whose tweet was in the link you sent me). He always does it the same way as I have, and confirmed it's for core function calls and not classes :-)

cv65kr: I don't know if you understood me correctly i check only ```array_merge``` function but git showed huge part of function ;)

delboy1978uk: Ah! Yes I see now. I thought you highlighted the \Twig_Environment! :-P

delboy1978uk: Fixing now! 👍😃

cv65kr: :+1:

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
```diff
@@ -24,37 +27,35 @@ public function __construct(DomainConfigurationInterface $domainConfiguration)
...
+            new Twig_SimpleFunction('is_multidomain_site', [$this, 'isMultiDomainSite']),
+            new Twig_SimpleFunction('get_switched_host', [$this, 'getSwitchedHost']),
+            new Twig_SimpleFunction('switched_host_is_current', [$this, 'switchedHostIsCurrent']),
+        ];
     }
 
     /**
      * Render multidomain switcher widget.
      *
      * @param \Twig_Environment $env
```
deZinc: Nitpicking here: can you remove the \

delboy1978uk: haha, good! I hate fully qualified stuff mid code! Defining it all with use at the top, and never doing something like `new \DateTime()` is the way forward! But sure, in a docblock, you are indeed just nitpicking! :-P

Numkil:
@sandergo90 Do you maybe know if there are situations where the switchedhost could be an array instead like it is being used before this fix? I don't want to fix one usecase while breaking another.

delboy1978uk:
As far as I could see there are only two DomainConfiguration classes, one returns null, the other a string

src/Kunstmaan/AdminBundle/Twig/MultiDomainAdminTwigExtension.php
```diff
@@ -78,15 +79,20 @@ public function getSwitchedHost()
     }
 
     /**
-     * @return string
+     * @return bool
      */
     public function switchedHostIsCurrent()
     {
-        return $this->domainConfiguration->getHost() == $this->getSwitchedHost()['host'];
+        return $this->domainConfiguration->getHost() === $this->getSwitchedHost();
```
Devolicious: Yes it does return an array when you are using the Multidomain feature. So this will break.

delboy1978uk: Which class returns an array? The multi domain configuration class returns `$this->hosts[$host]`, which is a string, and the normal DomainConfiguration returns null.
I'll test this on one of our multi domain sites to make sure the behaviour is the same though!

delboy1978uk: Ok having checked on a multi domain site it does indeed return an array, so I'll fix the docblock, as it is obviously what was wrong, put the array part back in , and we should merge this anyway, since code has also been tidied up!

sandergo90:
@delboy1978uk can you rebase against master because it's not a bugfix anymore.

Issue #1713, opened 2017-11-28
Issues with new toolbar helpers


wesleylancel:
I just tried turning on the new toolbar helpers (on the 5.0 branch) using

```
kunstmaan_admin:
    enable_toolbar_helper: true
```

but all I'm getting is: `An error occurred while loading the web debug toolbar. Open the web profiler.`.

Also, it seems that the translator datacollector overrides Symfony's one, causing that part of the original profiler to be disabled.

Devolicious:
@wesleylancel, that's strange because that should be the only thing that you have to do. Did you get this on a new project or on an existing one?

wesleylancel:
@Devolicious tried two existing projects. They don't have anything special regarding to the toolbar though.

cv65kr:
Check logs ```var/logs/dev.log``` and try establish what is wrong or give our more informations.

wesleylancel:
Found this in my logs:

```
[2017-11-30 09:35:45] request.CRITICAL: Uncaught PHP Exception Twig_Error_Runtime: "Variable "data" does not exist." at /vendor/kunstmaan/bundles-cms/src/Kunstmaan/AdminBundle/Resources/views/Toolbar/bundles_version.html.twig line 10 {"exception":"[object] (Twig_Error_Runtime(code: 0): Variable \"data\" does not exist. at /vendor/kunstmaan/bundles-cms/src/Kunstmaan/AdminBundle/Resources/views/Toolbar/bundles_version.html.twig:10)"} {"file":"/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/EventListener/ExceptionListener.php","line":89,"class":"Symfony\\Component\\HttpKernel\\EventListener\\ExceptionListener","function":"logException"}
```

cv65kr:
@wesleylancel, do You tried clear cache?

wesleylancel:
Yes, but it doesn't seem to help.

We do have the version checker disabled, does the toolbar check this?

sandergo90:
@wesleylancel In the isEnabled function of the BundleVersionDataCollector there is a check for this.

cv65kr:
@sandergo90 yes, indeed

wesleylancel:
Okay, so that shouldn't be an issue then. Not sure what's happening then.

cv65kr:
@wesleylancel, @sandergo90 i sent a fix for this.

PR #1721, opened 2017-11-30
[AdminBundle] Fix toolbar version checker

cv65kr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1713

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Toolbar/bundles_version.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Toolbar/bundles_version.html.twig
@@ -3,44 +3,45 @@
         {% set data = collector.getTemplateData().data %}
     {% endif %}
 
-    {% set status_color = '#00529B' %}
-    {% set txt = 'toolbar.bundle_version.unknown'|trans %}
-    {% set version = null %}
+    {% if data is defined %}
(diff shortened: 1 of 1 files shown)


sandergo90:
@cv65kr I think we should also check why this template is still rendered when the version checker is disabled. This data check is already a good thing, but it shouldn't be rendered at all.

cv65kr:
@sandergo90, It's weird
If i change this part in ```Kunstmaan/AdminBundle/Resources/views/Toolbar/toolbar.html.twig```:
```
{% for collector in collectors %}
        {{ block_render(collector.template, 'toolbar', collector.collectData()) }}
    {% endfor %}
```
to
```
    {% for collector in collectors %}
        {% if collector.isEnabled() %}
            {{ block_render(collector.template, 'toolbar', collector.collectData()) }}
        {% endif %}
    {% endfor %}
```
that should be correct. But ignore these part, what is wrong?

sandergo90:
@cv65kr after looking further in it, your PR seems like the correct solution. When the isEnabled() function returns false, it will pass an empty array as data. So the check for empty date in the twig file is enough.

Numkil:
@cv65kr thanks!

Issue #1722, opened 2017-11-30
[Insight] Form types should be in Form/Type folders - in src/Kunstmaan/AdminBundle/Form, line 0


delboy1978uk:
**[in src/Kunstmaan/AdminBundle/Form](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/918f344a11cb270dd559ffc6574ce6ad750a701b/src/Kunstmaan/AdminBundle/Form#L0)**

> A form type has been found outside Form/Type folder

```

src/Kunstmaan/AdminBundle/Form/
├── ColorType.php
├── DashboardConfigurationType.php
├── GroupType.php
├── MediaTokenTransformer.php
├── RangeType.php
├── RoleDependentUserFormInterface.php
├── RoleType.php
├── UserType.php
└── WysiwygType.php
```

---------------------

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/34aece18-3640-47ea-8e9f-96b79919905b/analyses/493)_

delboy1978uk:
Moving the classes into a Type folder would break BC. However, what we can do as a non-BC breaking solution is to create copies of the classes into the Type subdirectory, fix the namespaces on them, take all the logic out of the original classes, make them extend the new ones, and put a `@deprecated` doc block on the class.

delboy1978uk:
Hi Sander, how are you? I take it shifting everything into a folder just to shut up Sensio Insights isn't happening now? :-P Also, say hello to everyone from me please!

sandergo90:
Hi @delboy1978uk, indeed 😄 Everything fine here! I will say hello to the team 😉 ! Cheers!

PR #1723, opened 2017-11-30
[AdminBundle] Deprecate Form Types in favour of Symfony namespacing c…

delboy1978uk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #1722


delboy1978uk:
All form types have now been deprecated, we can start using the new ones now!

delboy1978uk:
Commits have been squashed :-)

sandergo90:
@delboy1978uk did you also change the generators ? Because they will need to use the new form type.

delboy1978uk:
No, I just created the new form types. The old ones are still there, so nothing has broken. 

The difference is the old classes now extend the new ones, all the methods moved into the new ones, and @deprecated tags have been added to the old ones. So code will run normally, and PHPStorm will score through any usage of the old ones, and it's as simple as adjusting the namespace to upgrade.

I guess I could do a site wide search for uses within the bundles and replace them too then?

sandergo90:
@delboy1978uk I understand that you deprecated the old ones. But you now created the new ones, so the generators will need to use the new ones. Otherwise new projects will still use the deprecated forms. So maybe site wide search will do the trick ;-)

acrobat:
on second thought I'm 👎on this change as it is deprecating for the sake of deprecating and changing our code (which is perfectly fine) to keep a tool happy. It will also generate more work for our users to upgrade with "real" benefits.

Devolicious:
I'm with @acrobat on this. We just need to fix that particular rule in sensiolabs insight I think. So for now I'm gonna close this PR. Sorry @delboy1978uk.

PR #1729, opened 2017-12-04
[POC] Final commands and commands as services

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes(ish)
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

As discussed last week I would look into ways of closing the class api a bit more (Final classes, private properties, etc).
This is a POC with multiple fixes combined:

```
```

This is just a POC so please give some feedback on this proposal. In a first phase we could start with commands and then move to another part of the bundles.

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/Resources/config/commands.yml
+++ b/src/Kunstmaan/TranslatorBundle/Resources/config/commands.yml
@@ -1,0 +1,6 @@
+services:
+    Kunstmaan\TranslatorBundle\Command\ImportTranslationsCommand:
+        class: Kunstmaan\TranslatorBundle\Command\ImportTranslationsCommand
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/TranslatorBundle/Resources/config/commands.yml
```diff
@@ -0,0 +1,6 @@
+services:
+    Kunstmaan\TranslatorBundle\Command\ImportTranslationsCommand:
+        class: Kunstmaan\TranslatorBundle\Command\ImportTranslationsCommand
```
jockri: parameter?

acrobat:
POC was discussed so closing this PR and I will create separate PR's for each bundle.

Issue #931, opened 2016-01-21
Shell helper doesn't work on Mac OSX


henrypenny:
I'm on OSX 10.10.5 and the Shell helper isn't working as expected.

I've removed `nohup` from:

`vendor/kunstmaan/bundles-cms/src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php:22`

and this solves it for my setup. Is there a good way to detect if its running on Mac and put a switch in place?

mlebkowski:
You can override this utility in your application.

```
class OsxShellHelper extends Shell {
    public function runInBackground($command, $priority = 0) {
       // custom logic
    }
}
```

And in your `config.yml`:

```
parameters:
   kunstmaan_utilities.shell.class: OsxShellHelper
```

whitewhidow:
perhaps we can check if nohup is present on the system, by dong something like this:

        $try = shell_exec("nohup date");
        if ($try !== null) {
            //nohup is working 
        } else {
            //nohup is not working
        }

henrypenny:
@whitewhidow 
Feature detection might be a little more robust?

whitewhidow:
@henrypenny what exactly do you mean by 'Feature detection' ?

henrypenny:
As per your suggested solution: testing whether a given feature (`nohup`) is available as opposed to assuming that OSX doesn't have it. Might there be other OSs without `nohup`?

wesleylancel:
It seems my OSX does have nohup. Maybe just test with `which nohup`?

whitewhidow:
yep that seems like a nice solution ..

kimausloos:
The use of ```which``` for this kind of thing should be avoided for various reasons, there are alternatives though: http://stackoverflow.com/a/677212

delboy1978uk:
Maybe use `is_executable()` ? http://jp.php.net/manual/en/function.is-executable.php

delboy1978uk:
We don't need `nohup`. If we use `popen()` or `proc_open()` to run the command, then the process is *already* backgrounded :-)
http://php.net/manual/en/function.popen.php 
http://php.net/manual/en/function.proc-open.php

Numkil:
The problem should be resolved in master branch. Instead of using shell_exec we are now using the Symfony Process object. Process does use proc_open as mentioned in the above comment. This completely negates the necessity for nohup and removes any platform specific code.

PR #1748, opened 2017-12-07
[UtilitiesBundle] Using Symfony Process in Shell Class

cv65kr:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes (nohup)
| New feature?  | Improvment
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | #931

### DIFF
--- a/src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
+++ b/src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
@@ -2,6 +2,8 @@
 
 namespace Kunstmaan\UtilitiesBundle\Helper\Shell;
 
+use Symfony\Component\Process\Process;
+
 /**
  * A wrapper class which makes it possible to execute shell commands in the background.
  */
@@ -9,32 +11,41 @@ class Shell implements ShellInterface
 {
 
     /**
-     * @param string $command  The command
-     * @param int    $priority The priority
+     * @param string $command The command
+     * @deprecated int $priority
      *
-     * @return string The process id
+     * @return int The process id
+     *
+     * @throws \Symfony\Component\Process\Exception\RuntimeException
+     * @throws \Symfony\Component\Process\Exception\LogicException
      */
     public function runInBackground($command, $priority = 0)
     {
-        if ($priority) {
-            $pid = shell_exec("nohup nice -n $priority $command > /dev/null & echo $!");
-        } else {
-            $pid = shell_exec("nohup $command > /dev/null & echo $!");
-        }
+        $process = new Process($command);
+        $process->disableOutput();
+        $process->start();
 
-        return $pid;
+        return $process->getPid();
     }
 
     /**
      * @param int $pid
      *
      * @return boolean
+     * @throws \Symfony\Component\Process\Exception\LogicException
+     * @throws \Symfony\Component\Process\Exception\RuntimeException
      */
     public function isRunning($pid)
     {
-        exec("ps $pid", $processState);
+        $process = new Process(
+            sprintf('ps -p %s -o pid', $pid)
+        );
+        $process->run();
 
-        return (count($processState) >= 2);
+        $output = trim($process->getOutput());
+        $processState = explode("\n", $output);
+
+        return (2 >= count($processState));
     }
 
     /**
@@ -43,12 +54,16 @@ class Shell implements ShellInterface
      * @param int $pid
      *
      * @return boolean true when the process was successfully killed, false when the process wasn't running.
+     * @throws \Symfony\Component\Process\Exception\RuntimeException
+     * @throws \Symfony\Component\Process\Exception\LogicException
      */
     public function kill($pid)
     {
         if ($this->isRunning($pid)) {
-            exec("kill -KILL $pid");
-
+            $process = new Process(
+                sprintf('kill -KILL %d', $pid)
+            );
+            $process->run();
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


Numkil:
@cv65kr 

If you are going to be refacturing this class could you also take this issue in account?
#931 
The problem is that nohup is not always installed on each computer and then it will crash.
@delboy1978uk provided a solution in the last comment of this thread.

Thanks!

src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
```diff
@@ -32,9 +36,14 @@ public function runInBackground($command, $priority = 0)
      */
     public function isRunning($pid)
```
Numkil: I see that the symfony Process object has a method called isRunning. Can you not use that method instead of this old convoluted way?

src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
```diff
@@ -10,19 +12,21 @@ class Shell implements ShellInterface
...
      * @return string The process id
      */
     public function runInBackground($command, $priority = 0)
     {
-        if ($priority) {
-            $pid = shell_exec("nohup nice -n $priority $command > /dev/null & echo $!");
-        } else {
-            $pid = shell_exec("nohup $command > /dev/null & echo $!");
-        }
+        $command = null !== $priority ?
```
Numkil: As said in my other comment we should be able to drop the nohup from here especially if we are using Process->start() then it is already running in background. 
Also we can also disable output using the Process object, that way we don't have to write to /dev/null ourselves either.

Numkil:
@cv65kr also please rebase to master for this PR. Since it is not a necessary bugfix I would prefer merging into master.

cv65kr:
@Numkil, Symfony process as default using ```proc_open``` so i must just remove nohup [https://github.com/symfony/process/blob/eb4b2f8503461a5f18a0967f7a3a5a3d988564ca/Process.php#L301](https://github.com/symfony/process/blob/eb4b2f8503461a5f18a0967f7a3a5a3d988564ca/Process.php#L301)

src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
cv65kr: ```isRunning``` waiting for process to finish, but we need check output command ```ps -p PID -o pid```

Numkil:
@cv65kr That is great news. Please do so.

src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
Numkil: isRunning does not wait for a process to finish i think. That is what the Process->wait() function is for. The isRunning method just returns true or false.

cv65kr: Tested on ```ls -l``` and always return false.

Numkil: ls -l is not a continuous process so there is no way the process could return true. Maybe try tail -f ~/.bashrc. This is not a process that will instantly quit.

Numkil: Ofcourse replace ~/.bashrc if you do not have that file.

cv65kr: Also tested, it doesn't change anything.

Numkil: Ok if it does not work like that then we will use the existing way like you made it.

Devolicious:
@cv65kr This PR needs a lot of clean up. There is even an unresolved merge conflict `<<<<<HEAD` in there somewhere. Please let us know if you are going to clean up or if we can just close this one.

Numkil:
@Devolicious thats because I changed the base branch from 4.0 to master. Since its not a bugfix but an improvement. Only needs a rebase.

cv65kr:
@Numkil, @Devolicious done.

src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
```diff
@@ -2,39 +2,54 @@
...
+            sprintf('%d %s', (int) $priority, $command) :
```
Numkil: Does the priority still do anything? I think this syntax will crash because you are just sending a number to the terminal now and not to nohup. We can probably ignore the priority for now and remove it from function definition in 5.0 branch

cv65kr: @Numkil, you have right, fixed. Also added BC Breaks to description.

src/Kunstmaan/UtilitiesBundle/Helper/Shell/Shell.php
```diff
@@ -2,39 +2,48 @@
...
-     * @param int    $priority The priority
+     * @param string $command The command
      *
-     * @return string The process id
+     * @return int The process id
+     * @throws \Symfony\Component\Process\Exception\LogicException
+     * @throws \Symfony\Component\Process\Exception\RuntimeException
      */
-    public function runInBackground($command, $priority = 0)
+    public function runInBackground($command)
```
Numkil: @cv65kr 
Ok you removed the argument to it completely now. This would make it a BC break and would need to go in 5.0 branch.

Please choose one of both options.

- rebase branch to 5.0
OR
- add the argument back in the function definiton but not use it in the implementation and add Deprecated annotation to that argument.

cv65kr: @Numkil done, if you merge i will sent next PR for 5.0 version.

PR #1754, opened 2017-12-10
Bump knpBundles dependency

tarjei:
Bump knpBundles dependency to get rid of deprecation warnings. Tested with my local installation of 5.0.0-RC1

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes


Devolicious:
@tarjei This is not necessary. `~2.0` allows you to install verions `2.2`. You just need to run `composer update` and it should be fine.

PR #1759, opened 2017-12-13
[AllBundles] entity tests 100% covered

delboy1978uk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | improvement for development
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none - just getting code coverage up!


delboy1978uk:
I'll be adding more commits in here with more tests, but I guess we can merge this whenever since it's only tests being added

delboy1978uk:
I'll find and replace all of the KunstmaanTesting stuff once all the existing tests have been migrated over :-)

cv65kr:
@delboy1978uk can You squash commits?

delboy1978uk:
Yes I'll squash the commits once each bundle has been moved across and I've changed the namespace

delboy1978uk:
All tests moved, all `@covers` annotations removed, entity tests added, and all tests now live in the `Tests\Kunstmaan` namespace.

delboy1978uk:
@Numkil  ^ :-)

acrobat:
Why do we move the tests to a separate folder/namespace? It seems more logical to me to keep the tests for each bundle in each bundle dir/namespace. It is no problem for phpunit to run tests in subfolders and if we enable the subtree split again to have the separate packages available, the tests need to be in the packages itself.

delboy1978uk:
Hi @Devolicious there's a fake entity i was using in the test suite so i couljd test an abstract class, looks like there are extra methods in the master branch, you probably just need to add an empty method to it I'm guessing

Devolicious:
Hi @delboy1978uk,

Yes I'm currently fixing it but there were some other changes as well because of the switch from 4.0 to master branch. I'm gonna open a new PR that starts from your branch as base and then all the necessary patches on top of that.

Devolicious:
Continued in #1924

Devolicious:
@delboy1978uk You should subcribe on https://kumathon.kunstmaan.be/

delboy1978uk:
@Devolicious i think i already did

PR #1769, opened 2017-12-17
[AdminListBundle][5.1] Deprecate direct container access in controllers

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

I left the upgrade guide for now, let's do that after 5.0 is released and before the PR gets merged.

This PR prepares the code to move away from direct container access in controllers. All `$this->get()` usages are replaced by `$this->container->get()` to avoid triggering deprecations in non-user code. In 6.0 we should:
- remove the `get` and `getParameter` overrides
- extend from the symfony `AbstractController`
- override the `getSubscribedServices` method to extend the list of services to make sure we have all the "default" symfony services and the ones the kuma adminstlist controller needs in the service locator. Users extending from our base controller should just register their controllers as services.
  - I think for 6.0 we should support sf4.1+ so we can use the newly introduced "parameter bag" to access the parameters-as-a-service. See [symfony/symfony#25288](https://github.com/symfony/symfony/pull/25288)

This is just for the adminlist "base" controller. I will open a separate PR to do the changes for the other controllers just extending from the symfony controller class.

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
+++ b/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
@@ -29,6 +29,31 @@ use Symfony\Component\PropertyAccess\PropertyAccess;
 abstract class AdminListController extends Controller
 {
     /**
+     * {@inheritdoc}
+     *
+     * @deprecated
+     */
+    protected function get($id)
+    {
+        @trigger_error('Getting services directly from the container is deprecated in KunstmaanAdminListBundle 5.1 and will be removed in KunstmaanAdminListBundle 6.0. Register your controllers as services and inject the necessary dependencies.', E_USER_DEPRECATED);
+
+        return parent::get($id);
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @deprecated
+     */
+    protected function getParameter($name)
+    {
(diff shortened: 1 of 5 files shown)


src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
```diff
@@ -28,6 +28,27 @@
  */
 abstract class AdminListController extends Controller
 {
+    /**
+     * {@inheritdoc}
```
jockri: deprecated annotation

acrobat: Fixed!

src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
```diff
@@ -28,6 +28,31 @@
...
+     *
+     * @deprecated
+     */
+    protected function get($id)
+    {
+        @trigger_error('Getting services directly from the container is deprecated in 5.1 and will be removed in 6.0. Register your controllers as services and inject the necessary dependencies.', E_USER_DEPRECATED);
```
nicolas-grekas: Context needed: in *KunstmaanBundle* 5.1

acrobat: Good catch, fixed!

sandergo90:
Rebase to master

acrobat:
Done

PR #1778, opened 2018-01-05
Postgres compatibility fixes

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

I've started a new project with a postgres database and I got a lot of errors to start with. These fixes are "needed" to be able to use postgres as the backend database. I have tested these fixes with a clean install on mysql and postgres.

Summary of the fixes:

- Postgres strict compares fields so varying char = bigint with work in a where. So cast the field to the correct type (acl query)
- In the sql standard single quotes are used for strings
- In the sql standard boolean fields should be compared to `false/true` instead of `0/1`
- Postgres does not have an `IF` function, instead it has an `COALESCE` function


acrobat:
@Devolicious any chance this can still make it in 5.0? Or will it be moved to the first bugfix on 5.0?

Devolicious:
@acrobat I think that since this is a bugfix that there are 2 options. First option is like you said, wait until V5 has been release and than apply it as a bugfix for 5.0.1. Second option is that we rebase this and submit it to the 4.1 branch. From a safety point of view I like the first option better since we are moving away from V4 and that would mean that if we later found out that something is broken, that 4.1 stays broken since we don't support it anymore than. On the other hand you tested it, tests succeeded, so it would be nice if we could add this to 4.1 as well.

acrobat:
@Devolicious I've rebased the PR against 4.1, surprisingly no merge conflicts 🎉 💪  So this should be good to go for 4.1. I don't have any bundles 4 projects so only tested on a 5 project but as there were no merge conflicts this should work on all versions.

PR #1786, opened 2018-01-16
[DashboardBundle] fix javascript error

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

A previously commit added the block extrajavascript. This block is not needed because we can embed the js_footer and add our extra javascript here. TranslatorBundle also used this method.


wesleylancel:
@sandergo90 This wasn't tagged and documented as a BC but removing a block clearly should be one. This breaks any templates that extend from it and fill up that block.

sandergo90:
Hmm indeed @wesleylancel I thought that block was recently added. Should we add it again ?

wesleylancel:
Seems better to add it back, less existing stuff that will break. Also the `js_footer` block already has some content, so having a block dedicated to adding some extra Javascript seems useful.

PR #1788, opened 2018-01-18
[TranslatorBundle][5.1] Commands as services and mark commands as final

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

**Changes**

```
```

**Extra**

I've also added `doctrine/doctrine-migration-bundle` as a dependency because there is a command depending on this bundle. 

**Next up**

In version 6 we should replace the `extend ContainerAwareCommand` by just `extend Command`, remove the `$this->getContainer()` usages and remove the BC layer from the constructor of the commands. I've added `NEXT_MAJOR` docblocks to easily search things we should change/cleanup in version 6

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/Command/ExportTranslationsCommand.php
+++ b/src/Kunstmaan/TranslatorBundle/Command/ExportTranslationsCommand.php
@@ -3,14 +3,42 @@
 namespace Kunstmaan\TranslatorBundle\Command;
 
 use Kunstmaan\TranslatorBundle\Model\Export\ExportCommand;
+use Kunstmaan\TranslatorBundle\Service\Command\Exporter\ExportCommandHandler;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Serializer\Exception\InvalidArgumentException;
 
+/**
+ * @final since 5.1
+ * NEXT_MAJOR extend from `Command` and remove `$this->getContainer` usages
+ */
 class ExportTranslationsCommand extends ContainerAwareCommand
 {
+    /**
+     * @var ExportCommandHandler
+     */
+    private $exportCommandHandler;
+
+    /**
+     * @param ExportCommandHandler|null $exportCommandHandler
+     */
+    public function __construct(/* ExportCommandHandler */ $exportCommandHandler = null)
+    {
+        parent::__construct();
+
+        if (!$exportCommandHandler instanceof ExportCommandHandler) {
+            @trigger_error(sprintf('Passing a command name as the first argument of "%s" is deprecated since version symfony 3.4 and will be removed in symfony 4.0. If the command was registered by convention, make it a service instead. ', __METHOD__), E_USER_DEPRECATED);
+
+            $this->setName(null === $exportCommandHandler ? 'kuma:translator:export' : $exportCommandHandler);
+
+            return;
+        }
+
+        $this->exportCommandHandler = $exportCommandHandler;
+    }
+
     protected function configure()
     {
         $this
@@ -28,8 +56,12 @@ class ExportTranslationsCommand extends ContainerAwareCommand
         $format =       $input->getOption('format');
         $locales =      $input->getOption('locales');
 
-        if (is_null($format)) {
+        if (null === $format) {
             throw new InvalidArgumentException('A format should be defined, e.g --format yml');
+        }
+
+        if (null === $this->exportCommandHandler) {
+            $this->exportCommandHandler = $this->getContainer()->get('kunstmaan_translator.service.exporter.command_handler');
         }
 
         $exportCommand = new ExportCommand();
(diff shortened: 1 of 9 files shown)


src/Kunstmaan/TranslatorBundle/Command/ExportTranslationsCommand.php
```diff
@@ -3,14 +3,39 @@
...
+    private $exportCommandHandler;
+
+    public function __construct($translationRepository = null)
+    {
+        parent::__construct();
+
+        if (!$translationRepository instanceof ExportCommandHandler) {
+            @trigger_error(sprintf('Passing a command name as the first argument of "%s" is deprecated since version symfony 3.4 and will be removed in symfony 4.0. If the command was registered by convention, make it a service instead. ', __METHOD__), E_USER_DEPRECATED);
+
+            $this->setName(null === $translationRepository ? 'kuma:translator:export' : $translationRepository);
```
sandergo90: Can be changed to $this->setName($translationRepository ?? 'kuma:translator:export');

src/Kunstmaan/TranslatorBundle/Command/ExportTranslationsCommand.php
```diff
@@ -3,14 +3,39 @@
...
+ * NEXT_MAJOR extend from `Command` and remove `$this->getContainer` usages
+ */
 class ExportTranslationsCommand extends ContainerAwareCommand
 {
+    /**
+     * @var ExportCommandHandler
+     */
+    private $exportCommandHandler;
+
+    public function __construct($translationRepository = null)
```
sandergo90: Docblock

src/Kunstmaan/TranslatorBundle/Command/ImportTranslationsCommand.php
```diff
@@ -3,13 +3,38 @@
...
+ * NEXT_MAJOR extend from `Command` and remove `$this->getContainer` usages
+ */
 class ImportTranslationsCommand extends ContainerAwareCommand
 {
+    /**
+     * @var ImportCommandHandler
+     */
+    private $importCommandHandler;
+
+    public function __construct($importCommandHandler = null)
```
sandergo90: Docblock

src/Kunstmaan/TranslatorBundle/Command/ExportTranslationsCommand.php
sandergo90: Use same variable name translationRepository <--> exportCommandHandler

src/Kunstmaan/TranslatorBundle/Command/TranslationCacheCommand.php
```diff
@@ -2,14 +2,46 @@
...
+     * @var ResourceCacher
+     */
+    private $resourceCacher;
+
+    /**
+     * @var CacheValidator
+     */
+    private $cacheValidator;
+
+    public function __construct($translationRepository = null, $cacheValidator = null)
```
sandergo90: Idem

src/Kunstmaan/TranslatorBundle/Command/TranslationFlagCommand.php
```diff
@@ -1,16 +1,40 @@
...
+ * NEXT_MAJOR extend from `Command` and remove `$this->getContainer` usages
  */
 class TranslationFlagCommand extends ContainerAwareCommand
 {
+    /**
+     * @var TranslationRepository
+     */
+    private $translationRepository;
+
+    public function __construct($translationRepository = null)
```
sandergo90: Idem

src/Kunstmaan/TranslatorBundle/Command/TranslationFlagCommand.php
```diff
@@ -1,16 +1,40 @@
...
+    {
+        parent::__construct();
+
+        if (!$translationRepository instanceof TranslationRepository) {
+            @trigger_error(sprintf('Passing a command name as the first argument of "%s" is deprecated since version symfony 3.4 and will be removed in symfony 4.0. If the command was registered by convention, make it a service instead. ', __METHOD__), E_USER_DEPRECATED);
+
+            $this->setName(null === $translationRepository ? 'kuma:translator:flag' : $translationRepository);
```
sandergo90: Idem

src/Kunstmaan/TranslatorBundle/Command/ExportTranslationsCommand.php
acrobat: Sf 3.4 and the current 5.0 branch (https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/5.0/composer.json#L19) still allow php 5.6. Should we bump this requirement or keep it 5.6 compatible for now?

src/Kunstmaan/TranslatorBundle/Command/ExportTranslationsCommand.php
acrobat: All docblocks are fixed + correct typehints are added in commands so we can "activate" those in 6.0

sandergo90:
Rebase to master

acrobat:
Done

PR #1789, opened 2018-01-18
[UtilitiesBundle][5.1] Commands as services and mark commands as final

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

**Changes**

```
```

**Next up**

In version 6 we should replace the `extend ContainerAwareCommand` by just `extend Command`, remove the `$this->getContainer()` usages and remove the BC layer from the constructor of the commands. I've added `NEXT_MAJOR` docblocks to easily search things we should change/cleanup in version 6

### DIFF
--- a/src/Kunstmaan/UtilitiesBundle/Command/CipherCommand.php
+++ b/src/Kunstmaan/UtilitiesBundle/Command/CipherCommand.php
@@ -2,6 +2,7 @@
 
 namespace Kunstmaan\UtilitiesBundle\Command;
 
+use Kunstmaan\UtilitiesBundle\Helper\Cipher\CipherInterface;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/UtilitiesBundle/Command/CipherCommand.php
```diff
@@ -20,14 +25,31 @@ class CipherCommand extends ContainerAwareCommand
         3 => 'Decrypt file'
     ];
 
+    public function __construct($cipher = null)
```
sandergo90: Same remarks as previous PR, add docblocks everywhere.

sandergo90: And change null comparisons with null coalesce operator

acrobat: Docblocks and typehints are fixed. The null coalesce change I didn't do yet because of comment in the other PR. https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1788#discussion_r162570690

acrobat:
I will rebase this and the other related pr's against master and change the target branch!

sandergo90:
Superb ! Thanks @acrobat

sandergo90:
Rebase to master

acrobat:
Done

PR #1814, opened 2018-02-08
[WIP][AllBundles] remove deprecations for symfony 3.4

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

Most of this PR will add child definitions for existing services. The service id's are now the fully qualified class names instead of id's. The original id's are now child definitions so they can still be used in version 5 but they are deprecated. In version 6 they will be removed.


Numkil:
@sandergo90 tests are failing

sandergo90:
Thats why it is a WIP 😁

sandergo90:
PR ready

acrobat:
I think we should check if we can throw some deprecations here, to make the upgrade to v6 (without the old style service ids) easier.

2 cases (I can think of now) where it could be usefull to throw deprecations:

- When the old style service id is used in user code ($this->get('old.id') or injecting with the old name). We should warn the user these id's are being replaced by the FQCN instead. Currently I don't have an idea how to do this, but I'm thinking about it 😄 
- Deprecate the usage of the `xxxx.class` parameters. If the user needs to override the service they should override the service definition instead of only the class. We could use the new `addParameteredAliases ` method and throw deprecations there. https://github.com/sandergo90/KunstmaanBundlesCMS/blob/c8e18c1c71f8b1bf3b9a8f2df7820d41d59f12e5/src/Kunstmaan/AdminListBundle/DependencyInjection/KunstmaanAdminListExtension.php#L84..L95
This can be done in a separate pr ofcourse.

sandergo90:
@acrobat was thinking about the same thing indeed! But currently have no idea how we could implement it. I did some research for the aliases deprecation, but an alias can not throw any deprecations 😞

sandergo90:
But like you said, for the second thing, we can indeed throw deprecations there.

sandergo90:
As discusses with @acrobat we will add child definitions instead of aliases. This child definitions can be deprecated.

sandergo90:
I've adapted all bundles and added a compiler pass to deprecate the old services.

acrobat:
I've tested this on a fresh 5.0 project but when upgrading to this branch I get an error. 
```The service "Kunstmaan\AdminBundle\EventListener\PasswordResettingListener" has a dependency on a non-existent service "FOS\UserBundle\Model\UserManagerInterface".```

We use the `@FOS\UserBundle\Model\UserManagerInterface` in AdminBundle/.../services.yml. The problem is the kuma standard edition locks the fosuserbundle to beta1 and this alias was introduced in 2.0.2. 

I will open a PR to the standardedition to remove this hard lock. But we also should add a conflict rule to the composer file of the bundles so 5.1 won't be installed together with fos/user-bundle < 2.0.2.

```
    "conflict": {
        "friendsofsymfony/user-bundle": "<2.0.2"
    },
```

Alternatively we could remove the "dependecy" on the alias and use the old style service id for that service. But still add a conflict rule to "enforce" fos userbundle >=2.0.0 (as this will be needed to fix the bundle inheritance deprecation, but I will open a PR for that!)

sandergo90:
@acrobat true ! I've updated the fos user-bundle in my own project but it needs to be updated in the standard edition to indeed. The part of fos-user must be removed from standard edition because it's already put in the KunstmaanBundlesCMS. I think it's good to already use the `@FOS\UserBundle\Model\UserManagerInterface` instead of the service id because we will need to change it eventually.

sandergo90:
@acrobat I will add the conflict rule in this PR.

PR #1828, opened 2018-02-15
[WIP][NodeSearchBundle]: add tag for search service

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |

**#1814 needs to be merged first !!!**

When deprecating the container access in version 6, we need to make sure we rewrite everything.

As discusses with @acrobat, i've created a service tag for the NodeSearcher class. Instead of getting the service from the container, you need to tag it with "kunstmaan_node_search.node_searcher". Upgrade guide updated.


src/Kunstmaan/NodeSearchBundle/KunstmaanNodeSearchBundle.php
```diff
@@ -3,6 +3,7 @@
 namespace Kunstmaan\NodeSearchBundle;
 
 use Kunstmaan\NodeSearchBundle\DependencyInjection\Compiler\DeprecationsCompilerPass;
+use Kunstmaan\NodeSearchBundle\DependencyInjection\Compiler\NodeSearcherCompilerPass;
```
acrobat: I think this can be removed?

src/Kunstmaan/NodeSearchBundle/Services/SearchService.php
```diff
@@ -120,13 +125,28 @@ public function search()
...
+                if ($entity->getSearcher() === \get_class($possibleSearcher)) {
+                    $searcher = $possibleSearcher;
+                }
+            }
+        }
+
+        if (!isset($searcher)) {
+            @trigger_error(
+                'Getting the node searcher from the container is deprecated in KunstmaanNodeSearchBundle 5.1 and will be removed in KunstmaanNodeSearchBundle 6.0.
+                Tag your service with the kunstmaan_node_search.node_searcher tag to add a searcher.',
```
acrobat: Add double quotes around the tag name for clarity

src/Kunstmaan/NodeSearchBundle/Services/SearchService.php
```diff
@@ -120,13 +125,28 @@ public function search()
...
+            foreach ($this->searchers as $possibleSearcher) {
+                if ($entity->getSearcher() === \get_class($possibleSearcher)) {
+                    $searcher = $possibleSearcher;
+                }
+            }
+        }
+
+        if (!isset($searcher)) {
+            @trigger_error(
+                'Getting the node searcher from the container is deprecated in KunstmaanNodeSearchBundle 5.1 and will be removed in KunstmaanNodeSearchBundle 6.0.
```
acrobat: Maybe use `Getting the node searcher "*the actual service id*" ...` so the users directly know which service to tag

src/Kunstmaan/NodeSearchBundle/Services/SearchService.php
```diff
@@ -120,13 +125,28 @@ public function search()
...
+        $pageNumber = $this->getRequestedPage($request);
         $entity = $request->attributes->get('_entity');
 
-        $pageNumber = $this->getRequestedPage($request);
-        $searcher   = $this->container->get($entity->getSearcher());
+        if (!empty($this->searchers)) {
+            foreach ($this->searchers as $possibleSearcher) {
+                if ($entity->getSearcher() === \get_class($possibleSearcher)) {
+                    $searcher = $possibleSearcher;
+                }
```
acrobat: Maybe add a `break;` when we found a searcher as there is no need to keep looping then

sandergo90:
@acrobat done!

src/Kunstmaan/NodeSearchBundle/Services/SearchService.php
```diff
@@ -120,13 +125,29 @@ public function search()
...
 
         // Retrieve the current page number from the URL, if not present of lower than 1, set it to 1
+        $pageNumber = $this->getRequestedPage($request);
         $entity = $request->attributes->get('_entity');
 
-        $pageNumber = $this->getRequestedPage($request);
-        $searcher   = $this->container->get($entity->getSearcher());
+        if (!empty($this->searchers)) {
+            foreach ($this->searchers as $possibleSearcher) {
+                if ($entity->getSearcher() === \get_class($possibleSearcher)) {
```
acrobat: We may need to rethink this check logic. This won't work for users still using "old style" service id's in their custom code. I think it's better to still make a compiler pass and inject an array, with keys = serviceIds and value = the reference object, into this class. This will make it also easier to check if we have a tagged service for that searcher or not (if the serviceId as key exist, we are good to go).

PR #1767, opened 2017-12-17
[AllBundle] Fix unused method, property, variable or parameter

cv65kr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets | no

Fix unused method, property, variable or parameter from Insight Sensiolabs.

PR #1838, opened 2018-02-18
[AllBundles]: Backport pr #1767 but keeping BC

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no (cleanup)
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | related to #1767

### DIFF
--- a/src/Kunstmaan/ConfigBundle/Controller/ConfigController.php
+++ b/src/Kunstmaan/ConfigBundle/Controller/ConfigController.php
@@ -49,11 +49,6 @@ class ConfigController
     private $configuration;
 
     /**
-     * @var ContainerInterface $container
-     */
-    private $container;
-
-    /**
      * @var FormFactoryInterface $formFactory
      */
     private $formFactory;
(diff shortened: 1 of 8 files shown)


src/Kunstmaan/ConfigBundle/Controller/ConfigController.php
```diff
@@ -73,15 +68,23 @@ public function __construct(
...
         $this->authorizationChecker = $authorizationChecker;
         $this->em = $em;
         $this->configuration = $configuration;
-        $this->container = $container;
+
+        if (func_num_args() > 6) {
+            @trigger_error(sprintf('Passing the "container" as the sixth argument in "%s" is deprecated in KunstmaanConfigBundle 5.1 and will be removed in KunstmaanConfigBundle 6.0. Remove the "container" argument from your service definition.', __METHOD__), E_USER_DEPRECATED);
+
+            $this->formFactory = func_get_arg(6);
+
```
sandergo90: The services.yml file needs to be updated to.

acrobat: Good catch! Fixed!

sandergo90: And I also think we need to keep the container variable set. Because when people are still using the container, they will still get an error because the private $container does not exist anymore.

acrobat: The container propery was private and there is no getter defined in the controller, so extending the controller will have the container property available in user code. So I think we can safely remove the property

sandergo90: True !

PR #1839, opened 2018-02-19
[WIP][AllBundles]: Deprecate NodePagesConfiguration

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets |

The NodePagesConfiguration class was to large. I've created an extra helper "PageHelper" to split things up. 

Another big thing is that the "service" action is being removed. I will add a PR to the master branch to deprecate this function. The use of this service action was actually "ugly" because it needed to be added to the entity itself. You need to use the SlugActionInterface for this. The NodePagesConfiguration class also used the service action to add extra content to the search view. This is now changed to an event that you can subscribe to.

Changes:

- PageHelper class
- getDefaultSearchViewTemplate function
- Remove of container
- Inject correct services
- Remove of "service" action
- Add of event on indexing search view

**#1814 needs to be merged first.**

### DIFF
--- a/UPGRADE-5.1.md
+++ b/UPGRADE-5.1.md
@@ -1,0 +1,8 @@
+# UPGRADE FROM 5.0 to 5.1
+
+## Deprecate use of old service id's
+
+PR https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1814 introduces the use of FQCN as service id's instead of the old id's.
+Therefore we added child definitions for the old service id's to still work.
+When you are still using the old service id's in your project, you will get deprecation messages.
+In KunstmaanBundlesCMS 6.0 the old service id's will be removed, so be sure to update your custom bundles.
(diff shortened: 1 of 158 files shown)


mlebkowski:
The `PageHelper` is breaking single responsibility principle. You had a class thad was to big and did too much (`NodePagesConfiguration`), and in result you now have two. In my opinion, this is a cosmetic change and has little effect on the overall architecture. It isn’t worth the BC break for me.

I would instead consider extracting private/protected methods into separate services (like `PageDomContentExtractor` for the `removeHtml` method). The `renderCustomSearchView`, `renderDefaultSearchView`, `addCustomData` methods are there to modify the `$doc` structure — they can be extracted into a collection of „providers” implementing some kind of `SeachContentProvider::addContent()` interface. They could be wired together using a separate container tag, an event dispatcher, or a mix of those methods.

The `getDefaultSearchViewTemplate` method would work better as a configuration option injected into one of the above `SearchContentProvider`s, so the developers don’t have to override classes to change simple configuration. Forcing your project users to override your classes has unpleasant side effects — suddenly, your protected methods become your public api, etc.

PR #1844, opened 2018-02-19
[WIP][NodeBundle]: deprecate container access

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/NodeBundle/Helper/Services/PageCreatorService.php
+++ b/src/Kunstmaan/NodeBundle/Helper/Services/PageCreatorService.php
@@ -3,6 +3,7 @@
 namespace Kunstmaan\NodeBundle\Helper\Services;
 
 use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\ORMException;
 use Kunstmaan\AdminBundle\Repository\UserRepository;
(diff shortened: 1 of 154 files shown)


src/Kunstmaan/NodeBundle/Helper/Services/PageCreatorService.php
```diff
@@ -58,54 +93,58 @@ public function setUserEntityClass($userEntityClass)
...
+     *                                           $translation->setTitle('NL titel');
+     *                                           }
+     *                                           ],
+     *                                           [   "language" => "fr",
+     *                                           "callback" => function($page, $translation) {
+     *                                           $translation->setTitle('FR titel');
+     *                                           }
+     *                                           ]
+     *                                           ]
+     *                                           Perhaps it's cleaner when you create one array and append another array for each language.
```
acrobat: This should be reverted. But I think if you move this example to the phpdoc description part, it won't get reformatted

acrobat:
Imo we should also remove/don't add docblocks like the `Class Xxxclass`, `XxxClass constructor.` or `@package Kunstmaan/xx/xx/xxx` as those don't really add much value

PR #1856, opened 2018-02-28
[LeadGenerationBundle] Remove jquery dependency leadgeneration bundle

janb87:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1853


src/Kunstmaan/LeadGenerationBundle/Resources/public/js/rule/AfterXScrollPercentRule.js
```diff
@@ -34,9 +30,9 @@
             }
             _locked = true;
 
-            var wintop = _$window.scrollTop(),
-                docheight = _$document.height(),
-                winheight = _$window.height(),
+            var wintop = window.scrollTop,
+                docheight = document.height,
```
janb87: This is actually not correct, window.height and document.height don't exist.

Issue #1857, opened 2018-03-01
[AdminBundle] Images and internal links in WYSIWYG fields are broken in 5.0.4


mtnorthrop:
Images and internal links inside the WYSIWYG field are broken.

When the field is saved, the square brackets in `img` tags' `src` and `a` tags' `href` attributes are URL encoded and are saved as `%5B` and `%5D` instead of `[` and `]`.

This means that all links and images added to WYSIWYG fields do not work.

Issue #1826, opened 2018-02-15
MediaTokenTransformer altering posted form content


delboy1978uk:
Using a wysiwyg in a form, I posted this:

```html
<p>Toegang tot %detailinfo% voor per aanvraag</p>
```

Because of line 70 of the media token transformer, this gets urldecoded to :
```html
<p>Toegang tot �tailinfo% voor per aanvraag</p>
```
This is the reverseTransform() method in question:
```php
 public function reverseTransform($content)
    {
        if (!trim($content)) {
            return '';
        }

        $crawler = new Crawler();
        $crawler->addHtmlContent($content);

        // Get all img tags and parse the token.
        $crawler->filter('img')->each(
            function (Crawler $node) {
                $image = $node->getNode(0);
                $src = $image->getAttribute('src');
                $parsed = parse_url($src, PHP_URL_QUERY);
                parse_str($parsed, $query);

                if (isset($query['token'])) {
                    $image->setAttribute('src', $query['token']);
                }
                $image->setAttribute('data-src', $src);
            }
        );

        return urldecode($crawler->filter('body')->html());
    }
```

The fix as suggested by @sandergo90 is to do the urldecoding on the crawled image's `$src`, and not the final html.

PR #1858, opened 2018-03-01
[AdminBundle] Fix escaping of img src attributes in WYSIWYG fields

mtnorthrop:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1857

The `MediaTokenTransformer::reverseTransform()` method uses the `Crawler::html()` method to produce the HTML it returns, but this method automatically URL-encodes the `src` attribute of image tags.

This PR does a search and replace on the final HTML to make sure that the square brackets used for the tokens are preserved.

_NB: the Symfony docs mention that the DomCrawler Component is not designed to re-dump HTML, and so it might be worth considering looking for a different tool to use in this class. (See https://symfony.com/doc/current/components/dom_crawler.html#component-dom-crawler-dumping)._

### DIFF
--- a/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
+++ b/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
@@ -63,10 +63,21 @@ class MediaTokenTransformer implements DataTransformerInterface
                 if (isset($query['token'])) {
                     $image->setAttribute('src', $query['token']);
                 }
-                $image->setAttribute('data-src', urldecode($src));
+                $image->setAttribute('data-src', $src);
             }
         );
 
-        return $crawler->filter('body')->html();
+        $html = $crawler->filter('body')->html();
+
+        // URL-decode square brackets in img and a tags
+        $html = preg_replace_callback(
+            '/<(img|a)\s+[^>]*>/',
+            function($matches) {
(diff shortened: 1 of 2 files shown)


wesleylancel:
The same seems to be happening to href-attributes of a-tags. Could this PR
be extended to fix that too?

mtnorthrop:
No problem, I've updated the PR to decode `<a>` tags as well. I'm doing the search and replace only within specific tags, and not on everything in the form field, in order to avoid reintroducing the bug that was reported in #1826.

wesleylancel:
@mtnorthrop Thanks!

This is quite an annoying bug, any chance of a quick merge and release?

mtnorthrop:
I hope this can be reviewed, merged and released soon, because in AdminBundle 5.0.4 all internal links and images added to the WYSIWYG fields are currently broken, and this PR fixes the issue.

Until this fix is released, anybody using 5.0.4 is going to have to explain to their clients that it's currently not possible to add links and images to their pages. :-(

sandergo90:
@mtnorthrop I've reviewed it and merged it now. We will create a new release tomorrow!

mtnorthrop:
@sandergo90 Thank you!

sandergo90:
@mtnorthrop new release submitted!

PR #1868, opened 2018-03-13
[KunstmaanAdminBundle]: access level for google should add groups

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When logging in with Google, the user will receive the ROLES given in access_roles in your configuration. This should not add roles, but should add the kuma groups instead.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserCreator.php
+++ b/src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserCreator.php
@@ -3,11 +3,12 @@
 namespace Kunstmaan\AdminBundle\Helper\Security\OAuth;
 
 use Doctrine\ORM\EntityManagerInterface;
+use FOS\UserBundle\Model\GroupInterface;
 use Kunstmaan\AdminBundle\Entity\User;
 
 class OAuthUserCreator implements OAuthUserCreatorInterface
 {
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/AdminBundle/Helper/Security/OAuth/OAuthUserCreator.php
```diff
@@ -56,7 +58,11 @@ public function getOrCreateUser($email, $googleId)
             }
 
             foreach ($this->getAccessLevels($email) as $accessLevel) {
-                $user->addRole($accessLevel);
+                /** @var GroupInterface $group */
+                $group = $this->em->getRepository('KunstmaanAdminBundle:Group')->findOneBy(['name' => $accessLevel]);
+                if (null !== $group) {
```
Numkil: Maybe we need some fallback so the process doesn't create a user without a role or group and effectively makes the username unusable. Some projects might have custom groups, whereas you can always be sure the roles will be the same.

sandergo90: Hmm, you configure the groups you want to add in the config.yml so it's your own responsibility to check if the group has roles in it IMHO.

Issue #1781, opened 2018-01-08
[DashboardBundle] GA API V3 is offline and bundle not compatible with V4


Devolicious:
So, I've noticed that Google had killed the Analytics api V3 in their library, old projects can still use them but new projects are not allowed to activate that api anymore. So if you setup you GA Dashboard you will get an error that the api is not enabled. AFAIK the main problem is that the name changed from `analytics.googleapis.com` to `analyticsreporting.googleapis.com`. But that is probably the first part of a bigger issue :)

deZinc:
Google analytics V3 is back online.
But we might want to refactor to the 4 anyway?

PR #1872, opened 2018-03-16
[DashboardBundle] change google api library

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets | #1781 

In preparation of issue #1781, I removed the old google api that we forked from alchemy/google-plus-api-client and changed if with the official google api client. Therefore some BC breaks are introduced, because some classes are having a different name. 

In a next PR, when google-api-client is ready for Analytics V4, we can change our bundles again. Probably this will be done before the 6.0 release so we don't have to BC break twice.

**I also did a lot of cleanup in the DashboardBundle!**


Numkil:
@sandergo90 can you rebase this PR? I think most of the content here isn't supposed to be included in the PR

acrobat:
Can't we create BC friendly version of this pr or a separate layer that can be introduced in 5.x so we can deprecate the old layer?

tarjei:
For 5.x is it possible to move the dependency into suggests or will that also break all code inside the bundle?

I just want to write some stuff to google sheets and it seems I'll have to define a whole separate project due to this dependency...

tarjei:
I'm sorry for hijacking this thread, but I thought I'd leave this comment somewhere someone would find it. 

_if_ you need to combine modern google apis with version 5.x _and_ you can forgo the nice analytics dashboard, you can do the following:

1. Disable loading the old api in composer.json by adding to the `autoload` key:
```composer.json
{
"autoload": 
  {
     "exclude-from-classmap": ["vendor/kunstmaan/google-api-custom/*"]
  }
}
```
2. Create a dummy config service for the frontend: 
```services.yml
services:
  kunstmaan_dashboard.helper.google.analytics.config:
      class: SiteBundle\Dashboard\DummyConfigHelper

```

```php 
/**
 * Overide dashboard config to not connect to google using the old api that has been removed.
 */
class DummyConfigHelper
{

    public function tokenIsSet(): bool
    {
        return false;
    }

    public function getAuthUrl() : string
    {
        return "https://www.novap.no";
    }

}
```

PR #1874, opened 2018-03-20
[GeneratorBundle] Change logic of page title

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/layout/Resources/views/Layout/layout.html.twig
+++ b/src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/layout/Resources/views/Layout/layout.html.twig
@@ -13,7 +13,17 @@
         {% set node = nodetranslation.node %}
     {% endif %}
     {% set nodemenu = get_node_menu(app.request.locale, node) %}
-    {{ websitetitle | trans }}{% block title %}{% if page is defined and nodemenu.getActiveForDepth(2) is not null %} - {{ get_title_for(page) }}{% endif %}{% endblock %}
+    {% block title %}
+        {% if page is defined and nodemenu.getActiveForDepth(2) is not null %}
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/layout/Resources/views/Layout/layout.html.twig
```diff
@@ -13,7 +13,17 @@
         {% set node = nodetranslation.node %}
     {% endif %}
     {% set nodemenu = get_node_menu(app.request.locale, node) %}
-    {{ websitetitle | trans }}{% block title %}{% if page is defined and nodemenu.getActiveForDepth(2) is not null %} - {{ get_title_for(page) }}{% endif %}{% endblock %}
+    {% block title %}
+        {% if page is defined and nodemenu.getActiveForDepth(2) is not null %}
+            {{ get_title_for(page) ~ ' - ' ~ websitetitle|trans }}
+        {% else %}
+            {{ websitetitle | trans }}
```
diskwriter: Shouldn't this be more of an OR situation? 
As in: As soon as the SEO title is set it should replace the website title from translations completely. Otherwise the user is limited in their SEO titles because it will always include the site title.

sandergo90: Yeah I also think this change is not needed. The website name always needs to be first in the title. I don't see any reasons why we should change this. Seo metatitle is needed for SEO and not for setting the title of a page. Is there any reason for this @treeleaf ?

sandergo90:
I don't think it's a bugfix. So @treeleaf can you rebase to master?

Issue #1799, opened 2018-01-31
[SearchBundle] Make ElasticSearch 6 compatible


Devolicious:

Numkil:
Closing because merged.

PR #1875, opened 2018-03-20
[NodeSearchBundle] Elastic search version 6 support

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1799

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/DependencyInjection/Configuration.php
+++ b/src/Kunstmaan/NodeSearchBundle/DependencyInjection/Configuration.php
@@ -44,6 +44,7 @@ class Configuration implements ConfigurationInterface
 
         if (ElasticSearchUtil::useVersion6()) {
             $properties->children()->booleanNode('fielddata');
+            $properties->children()->booleanNode('doc_values');
             $properties->children()
                 ->scalarNode('index')
                 ->beforeNormalization()
--- a/src/Kunstmaan/NodeSearchBundle/DependencyInjection/Configuration.php
+++ b/src/Kunstmaan/NodeSearchBundle/DependencyInjection/Configuration.php
@@ -5,6 +5,7 @@ namespace Kunstmaan\NodeSearchBundle\DependencyInjection;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Kunstmaan\NodeSearchBundle\Helper\ElasticSearchUtil;
 
 /**
  * This is the class that validates and merges configuration from your app/config files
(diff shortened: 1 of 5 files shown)


src/Kunstmaan/NodeSearchBundle/DependencyInjection/Configuration.php
```diff
@@ -28,16 +29,35 @@ public function getConfigTreeBuilder()
...
-        $properties->children()->booleanNode('include_in_all');
+        ];
+        if (!ElasticSearchUtil::useVersion6()) {
+            $types[] = 'string';
+        }
+
+        $properties->children()->scalarNode('type')->beforeNormalization()->ifNotInArray($types)->thenInvalid('type must be one of: ' . implode(', ', $types));
+
+        if (ElasticSearchUtil::useVersion6()) {
+            $properties->children()->booleanNode('fielddata');
```
sandergo90: I was reading the https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html article and like they say, it's not a good idea to use this fielddata element because it will cause a lot of memory issues. We should use https://www.elastic.co/guide/en/elasticsearch/reference/current/doc-values.html.

src/Kunstmaan/NodeSearchBundle/Helper/ElasticSearchUtil.php
```diff
@@ -0,0 +1,17 @@
...
+ * Class ElasticSearchUtil
+ */
+final class ElasticSearchUtil
+{
+    /**
+     * @return bool
+     */
+    public static function useVersion6()
+    {
+        return (PHP_VERSION[0] == 7 && !class_exists('\Elastica\Tool\CrossIndex'));
```
sandergo90: PHP_MAJOR_VERSION

src/Kunstmaan/NodeSearchBundle/Resources/doc/ElasticSearch.md
```diff
@@ -0,0 +1,50 @@
+# Elastic search support
+
+We have added support for Elastic search up to version 6.x.
+When you have a project that is using PHP 7 then you can use version 6.x when the project is using ruflin/elastica >= 6.0.
+
+## Dynamic configuration
+We have added dynamic support to detect which version your project is using.
+Dependancies for using the configuration of version 6 is that your project uses PHP 7,
```
sandergo90: Typo, can you check the document again for typo's ? 😄

src/Kunstmaan/NodeSearchBundle/DependencyInjection/Configuration.php
treeleaf: yes, using doc_values seems a good alternative, however, I don't think we should prohibit it. We are not using fielddata, but allowing that it is used as a mapping value. I'll add the doc_values as well.

src/Kunstmaan/NodeSearchBundle/Resources/doc/ElasticSearch.md
treeleaf: I have updated the doc. and removed the typo :)

src/Kunstmaan/NodeSearchBundle/DependencyInjection/Configuration.php
sandergo90: Alright, super

PR #1877, opened 2018-03-27
[NodeSearchBundle] [SearchBundle] ElasticSearch Backport so we can support bundles 3.6 websites on servers with ES6

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets | 

It is necessary to break our own rules regarding BC-breaks and Deprecations so we can comply with new security rules.


Numkil:
This includes everything up to #1875

treeleaf:
The tests are failing on 5.5. Can we drop support for PHP 5.5 ? @Numkil @Devolicious

Devolicious:
Yes we can

acrobat:
Semver allows bumping php versions in minor releases so 👍 on bumping to php 5.6

PR #1889, opened 2018-04-06
[MultiDomainBundle]: host override should be set before getting default locale

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When switching host in the backend on a multidomain site, the default locale is being called for the new host. But because the host override is set after the getDefaultLocale call, the locale of the host switched from will be returned.


CHANGELOG-5.X.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 5.0.5 / 2018-03-05
```
Devolicious: This does not belong in this PR

Issue #1899, opened 2018-04-11
[Node] Node online filter translations


dannyvw:
Kunstmaan 5.0.5
Twig 2.4.8

With the latest twig version, the "true", "false" boolean is converted to 1. See screenshot

![image](https://user-images.githubusercontent.com/1218389/38611412-e443b5d6-3d83-11e8-86a3-74c4c393a230.png)

Numkil:
Very strange behaviour indeed. It looks like it has something to do with the translation yml having the word true and false in it.

Numkil:
If you need an immediate fix you can in the translations tab of the settings overwrite the following keys.

filter.true
and filter.false
if they come from the database the error does not happen either.

PR #1909, opened 2018-04-13
[AdminListBundle] avoid translations freakout when the key is true or false without escaping

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1899

Translations would freak out and display 1 or '' in english locale without adding these quotes.

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Resources/translations/messages.en.yml
+++ b/src/Kunstmaan/AdminListBundle/Resources/translations/messages.en.yml
@@ -14,8 +14,8 @@ kuma_admin_list:
 
 # Filters
 filter:
-    true: true
-    false: false
+  "true": "true"
+  "false": "false"
     before: before
     after: after
     in: contains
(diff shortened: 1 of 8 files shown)


src/Kunstmaan/AdminListBundle/Resources/translations/messages.en.yml
```diff
@@ -14,8 +14,8 @@ kuma_admin_list:
 
 # Filters
 filter:
-    true: true
-    false: false
+    "true": true
```
acrobat: shouldn't the value be quoted instead of the key?

acrobat: I think both the key and the value need to be quoted, so it seems when I'm testing..

Numkil: For me it worked if i only commented the key.

Numkil: but i will update everything to be sure

Issue #1760, opened 2017-12-13
AbstractAuthor isn't abstract :-P


delboy1978uk:
I suggest we deprecate this class as the name is misleading (ArticleBundle).

We can move the methods out into a new Author class, and keep the AbstractAuthor class, but make it extend our new Author class. Then we can add a @deprecated annotation on the AbstractAuthor class, so no breaks etc.

PR #1919, opened 2018-04-16
[ArticleBundle] Add deprication error when using entities directly th…

treeleaf:
…at should be abstract

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #1760


sandergo90:
Master branch, it's a deprecation 😉

Issue #1900, opened 2018-04-11
[Media] No pagination in media list view


dannyvw:
In the thumb view you have pagination but not in the list view

Numkil:
Hey @dannyvw 
This has been fixed recently. I think it was done in this PR #1627. This might not yet be deployed on the demo website.

The pagination threshold for the list view is 25 objects per page.
I will close the issue but if you have any other problems please do not hesitate to reopen the issue.

dannyvw:
@Numkil  We updated to version 5.0.5 and the issue still exists. In which version is this fixed?

Numkil:
I use 5.0.x-dev because the fix has not been included in the next 5 release. 
however you other issue about the

dannyvw:
@Numkil Thanks for the information

Issue #1893, opened 2018-04-10
[MediaBundle] Duplicate pagination in media folder


dannyvw:
Duplicate pagination in media folder, see screenshot

![screenshot from 2018-04-10 17-08-49](https://user-images.githubusercontent.com/1218389/38565563-f08eeeca-3ce1-11e8-9298-622c5e24a22c.png)

Numkil:
This is indeed a regression. I will make a PR for this shortly.

Numkil:
This is also caused by the same PR that added pagination in list view
#1627

PR #1927, opened 2018-04-18
[AdminListBundle][MediaBundle] remove duplicate pagination and broken switch pagination limit

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1900 #1893

Pagination was duplicated in this widget.html because already implemented in the twig that includes this file. Also the pagination limit was broken in the media folder. every time you go to list view it would switch to 250 but then each time you did a new action like adding/removing/moving a medio object the pagination would switch back to 24 pagination limit in list view. this was very confusing and because will be improved with changeablelimit in master branch this can be removed for now.


sandergo90:
@dannyvw Can you verify if this fixes your problem? Thanks!

PR #1939, opened 2018-04-23
[UtilitiesBundle] Make parameter kunstmaan_utilities.cipher.secret op…

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Setting the parameter "kunstmaan_utilities.cipher.secret" is no longer required.

### DIFF
--- a/src/Kunstmaan/UtilitiesBundle/DependencyInjection/Configuration.php
+++ b/src/Kunstmaan/UtilitiesBundle/DependencyInjection/Configuration.php
@@ -2,7 +2,6 @@
 
 namespace Kunstmaan\UtilitiesBundle\DependencyInjection;
 
-use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 
@@ -19,22 +18,7 @@ class Configuration implements ConfigurationInterface
     public function getConfigTreeBuilder()
     {
         $treeBuilder = new TreeBuilder();
-        $root = $treeBuilder->root('kunstmaan_utilities');
-
-        /** @var ArrayNodeDefinition $pages */
-        $root
-            ->children()
-                ->arrayNode('cipher')
-                    ->addDefaultsIfNotSet()
-                    ->children()
-                        ->scalarNode('secret')
-                            ->isRequired()
-                            ->cannotBeEmpty()
-                            ->defaultValue('%secret%')
-                        ->end()
-                    ->end()
-                ->end()
-            ->end();
+        $treeBuilder->root('kunstmaan_utilities');
 
         return $treeBuilder;
     }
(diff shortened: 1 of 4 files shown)


src/Kunstmaan/UtilitiesBundle/DependencyInjection/Configuration.php
```diff
@@ -18,7 +19,22 @@ class Configuration implements ConfigurationInterface
...
+        /** @var ArrayNodeDefinition $pages */
+        $root
+            ->children()
+                ->arrayNode('cipher')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('secret')
+                            ->isRequired()
+                            ->cannotBeEmpty()
+                            ->defaultValue('%secret%')
```
sandergo90: The default value here can be removed IMHO.

treeleaf: Why remove? The clue of this change is that if you do not set a parameter value for "kunstmaan_utilities.cipher.secret", that it will use the "secret", which is already defined..

sandergo90: Yes that's true, but that value is being set in your if else statement in the KunstmaanUtilitiesExtension.php, right?

treeleaf: Yes, true.

sandergo90: So we don't need to provide it here then?

src/Kunstmaan/UtilitiesBundle/DependencyInjection/KunstmaanUtilitiesExtension.php
```diff
@@ -20,7 +20,13 @@ class KunstmaanUtilitiesExtension extends Extension
     public function load(array $configs, ContainerBuilder $container)
     {
         $configuration = new Configuration();
-        $this->processConfiguration($configuration, $configs);
+        $config = $this->processConfiguration($configuration, $configs);
+
+        if (isset($config['cipher'], $config['cipher']['secret'])) {
+            $container->setParameter('kunstmaan_utilities.cipher.secret', $config['cipher']['secret']);
+        } else if (!$container->hasParameter('kunstmaan_utilities.cipher.secret')) {
```
Devolicious: Remove this check since default value has already been set.

acrobat:
Can we make the last changes here so this is ready to merge? The "exception" is anoying as you see it each time you run a command 😛

src/Kunstmaan/UtilitiesBundle/DependencyInjection/KunstmaanUtilitiesExtension.php
treeleaf: I have removed the default value in Configuration.php

PR #1940, opened 2018-04-24
[NodeSearchBundle]: elasticsearch util should check the running version of elasticsearch

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When running elasticsearch 2.4 on a PHP 7 environment, the current check will still return that it has to use version 6. This extra check will create a connection to elasticsearch and will get the version.

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Helper/ElasticSearchUtil.php
+++ b/src/Kunstmaan/NodeSearchBundle/Helper/ElasticSearchUtil.php
@@ -2,16 +2,52 @@
 
 namespace Kunstmaan\NodeSearchBundle\Helper;
 
+use Elasticsearch\ClientBuilder;
+use Elasticsearch\Common\Exceptions\NoNodesAvailableException;
+
 /**
  * Class ElasticSearchUtil
  */
 final class ElasticSearchUtil
 {
+    /** @var array */
+    private static $esClientInfo;
+
     /**
      * @return bool
      */
     public static function useVersion6()
     {
-        return (PHP_MAJOR_VERSION == 7 && !class_exists('\Elastica\Tool\CrossIndex'));
+        if (PHP_MAJOR_VERSION < 7 || !class_exists('\Elastica\Tool\CrossIndex')) {
+            return false;
+        }
+
+        $info = self::getESVersionInfo();
+
+        if (null !== $info) {
+            $versionParts = explode('.', $info['version']['number']);
+            $majorVersion = $versionParts[0];
+
+            return ($majorVersion > 2);
+        }
+
+        return false;
+    }
+
+    /**
+     * @return array
+     */
+    private static function getESVersionInfo()
+    {
+        try {
+            if (null === self::$esClientInfo) {
+                $client = ClientBuilder::create()->build();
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/NodeSearchBundle/Helper/ElasticSearchUtil.php
```diff
@@ -12,6 +14,16 @@
...
     {
-        return (PHP_MAJOR_VERSION == 7 && !class_exists('\Elastica\Tool\CrossIndex'));
+        $client = ClientBuilder::create()->build();
+        $info = $client->info();
+
+        if (null !== $info) {
+            $versionParts = explode('.', $info['version']['number']);
+            $majorVersion = $versionParts[0];
+
+            return ($majorVersion > 2 && PHP_MAJOR_VERSION === 7 && !class_exists('\Elastica\Tool\CrossIndex'));
```
treeleaf: I propose to use $majorVersion >= 6

src/Kunstmaan/NodeSearchBundle/Helper/ElasticSearchUtil.php
```diff
@@ -12,6 +14,16 @@
      */
     public static function useVersion6()
     {
-        return (PHP_MAJOR_VERSION == 7 && !class_exists('\Elastica\Tool\CrossIndex'));
+        $client = ClientBuilder::create()->build();
+        $info = $client->info();
```
treeleaf: We can cache this in a static variable. Otherwise it's gets called multiple times (e.g. with cache:clear).

`private static $esClientInfo;`

  ```
  $info = self::getESVersionInfo();

/**
     * @return array
     */
    private static function getESVersionInfo()
    {
        if (null === self::$esClientInfo) {
            $client = ClientBuilder::create()->build();
            $info = $client->info();
            self::$esClientInfo = $info;
        } else {
            $info = self::$esClientInfo;
        }

        return $info;
    }
```

src/Kunstmaan/NodeSearchBundle/Helper/ElasticSearchUtil.php
sandergo90: @treeleaf Are the bundles changes working on elasticsearch 5 to ?

sandergo90: It's also available in ES5, so why checking only for >= 6 ?

treeleaf: I tested it also with ES5. 
Maybe as a second thought, we should first check on `!class_exists('\Elastica\Tool\CrossIndex')` and PHP version before fetching the ES version info. If this class exists, then we know version 5 of the library is installed. Also version 6 of the library only works in PHP 7.

sandergo90: True, I changed the check priority.

PR #1958, opened 2018-05-04
[ConfigBundle] Set the correct configuration entity as active

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When using more than one configuration entity, the original code would set all of them active, which causes weird problems


JZuidema:
@sandergo90 I'm not sure how. It's showing all these commits because i changed the base branch from master to 5.0. Should i create a new PR, or rebase our master branch to 5.0?

sandergo90:
Normally what we do, is creating a feature branch for each different PR, starting from the correct branch. So best is to create a new branch starting from the 5.0 branch 😉

PR #1973, opened 2018-05-11
[GeneratorBundle] Fix unquoted parameter in routing generator

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When generating a new site this creates a deprecation out-of-the-box.


Numkil:
@acrobat Since bugfixes should go to 5.0 and the file already exists in 5.0 branch, with the same error, i think this PR should be rebased to the 5.0 branch instead of master.

PR #1974, opened 2018-05-13
[MediaBundle] Deprecate service container usage in commands

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Follow-up pr that deprecates container access in the commands of the media bundle

### DIFF
--- a/src/Kunstmaan/MediaBundle/Command/CleanDeletedMediaCommand.php
+++ b/src/Kunstmaan/MediaBundle/Command/CleanDeletedMediaCommand.php
@@ -2,14 +2,51 @@
 
 namespace Kunstmaan\MediaBundle\Command;
 
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
+use Kunstmaan\MediaBundle\Helper\MediaManager;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 
+/**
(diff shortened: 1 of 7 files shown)


src/Kunstmaan/MediaBundle/Command/CleanDeletedMediaCommand.php
```diff
@@ -2,14 +2,49 @@
...
 
+use Kunstmaan\MediaBundle\Helper\MediaManager;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 
+/**
+ * @final since 5.1
```
sandergo90: Don't wee need to make them final then? 🤔

src/Kunstmaan/MediaBundle/Command/CleanDeletedMediaCommand.php
```diff
@@ -30,6 +65,11 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if (null === $this->em) {
```
sandergo90: I would do this in the constructor. Every other thing we deprecated is done in the constructur.

acrobat: We can't access the container yet in the constructor as this will be set later. Using the container in the constructor throws this error
```
[LogicException]
  The container cannot be retrieved as the application instance is not yet set.
```

sandergo90: Correct 😉

PR #1980, opened 2018-05-21
[NodeBundle] Remove incorrect deprecation

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

This deprecation is added in #1840 but there is actually no alternative to "fix" this deprecation. So I suggest to remove it for now and add it back if we have an alternative for users to change their code to.

### DIFF
--- a/src/Kunstmaan/NodeBundle/Entity/AbstractPage.php
+++ b/src/Kunstmaan/NodeBundle/Entity/AbstractPage.php
@@ -127,12 +127,9 @@ abstract class AbstractPage extends AbstractEntity implements PageInterface
 
     /**
      * {@inheritdoc}
-     *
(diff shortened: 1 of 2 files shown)


sandergo90:
@acrobat there is an alternative. You should implement the SlugActionInterface.

acrobat:
So you should implement the `SlugActionInterface` and use the `getControllerAction` method instead of the `service` method?

sandergo90:
@acrobat correct!

acrobat:
@sandergo90 I've created a new PR to "fix" the deprection message. See #2004

PR #1986, opened 2018-05-26
[NodeBundle] Select correct root node for the urlchooser in a multidomain setup

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Currently the whole node/pages (all nodes for all domains) tree is shown when picking an url in a multidomain setup instead of only showing the nodes/pages related to the selected domain. The incorrect tree also causes incorrect urls to be selected. The generated url slug was also incorrect as it used the host instead of the config id.

### DIFF
--- a/src/Kunstmaan/NodeBundle/Controller/WidgetsController.php
+++ b/src/Kunstmaan/NodeBundle/Controller/WidgetsController.php
@@ -104,7 +104,14 @@ class WidgetsController extends Controller
         $depth = $this->getParameter('kunstmaan_node.url_chooser.lazy_increment');
 
         if (!$id || $id == '#') {
-            $rootItems = $em->getRepository('KunstmaanNodeBundle:Node')->getAllTopNodes();
+            $domainConfig = $this->get('kunstmaan_admin.domain_configuration');
+
+            if ($domainConfig->isMultiDomainHost()) {
+                $switchedHost = $domainConfig->getHostSwitched();
+                $rootItems = [$domainConfig->getRootNode($switchedHost['host'])];
+            } else {
+                $rootItems = $em->getRepository('KunstmaanNodeBundle:Node')->getAllTopNodes();
+            }
         } else {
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/NodeBundle/Controller/WidgetsController.php
```diff
@@ -176,16 +183,16 @@ protected function nodesToArray($locale, $rootNodes, $depth = 2)
...
+        $switched = $domainconfig->getHost() == $switchedHost['host'];
 
         $results = [];
 
         /** @var Node $rootNode */
         foreach ($rootNodes as $rootNode) {
             if ($nodeTranslation = $rootNode->getNodeTranslation($locale, true)) {
                 if ($isMultiDomain && !$switched) {
-                    $slug = sprintf("[%s:%s]", $switchedHost, "NT" . $nodeTranslation->getId());
+                    $slug = sprintf("[%s:%s]", $switchedHost['id'], "NT" . $nodeTranslation->getId());
```
sandergo90: Wouldn't this be BC breaking?

sandergo90:
@acrobat can you check the failing tests?

src/Kunstmaan/NodeBundle/Controller/WidgetsController.php
acrobat: I don't think so, the current outputted format is "broken". It produces this output for example `[www.example.com:NT55]`. But this won't actually work for [the regex](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/33519719dc36c6adc28caf5a56b294f4cae0750a/src/Kunstmaan/NodeBundle/Helper/URLHelper.php#L76) behind the `replace_url` twig function. It expects a format of [site_name:NT55].

acrobat:
@sandergo90 Test fail is not related. I've created a new PR to fix the test #2002

PR #1987, opened 2018-05-28
[GeneratorBundle] Replace deprecated twig raw tag in scss file

### LINKED ISSUE #1926: [GeneratorBundle] kuma:generate:default-site throws exception with twig 2
When you generate a default site with kuma:generate:default-site and you have twig 2 installed (tested with 2.4.8 and php 7.1), then it will throw an exception.

Twig_Error_Syntax: Unclosed comment. (uncaught exception) at .../vendor/kunstmaan/bundles-cms/src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/layout/Resources/ui/scss/config/general/_colors.scss line 71 while running console command `kuma:generate:default-site`

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | fixes #1926 

Replace the deprecated `raw` tag by `verbatim` to support twig 2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
         "symfony-cmf/routing-bundle": "~2.0",
         "fpn/doctrine-extensions-taggable": "~0.9",
         "sensio/generator-bundle": "~3.0",
+        "twig/twig": "^1.12|^2.0",
         "twig/extensions": "~1.0",
         "egulias/email-validator": "~1.2",
         "box/spout": "^2.5",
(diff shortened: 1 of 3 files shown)


sandergo90:
@acrobat maybe we should keep the raw tag and add 1.12 only to composer.json. In the new major release we can switch to twig 2 then ?

acrobat:
@sandergo90 we already allowed twig 2 by the [symfony 3.4 dependency](https://github.com/symfony/symfony/blob/3.4/composer.json#L23), I just made it an explicit dependency. But I also think it's good to replace the deprecated features so our code is more "future proof" but still be compatible with both versions

Issue #1904, opened 2018-04-12
Export pagetemplate


dannyvw:
Can you explain what the option "Export pagetemplate" does. It gives an modal with some html, but not sure what we can do with this.

Is it possible to disable this for the whole installation?

Numkil:
The only way to remove that button for now is to overwrite this file.
src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/_modals.html.twig
And remove the include line for the export pagetemplate twig.

dannyvw:
Is it possible to create an option to enable this button? Instead of only based on interface and superadmin. 

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/59097b1501cfd670da0e936b35347344a097d147/src/Kunstmaan/NodeBundle/Helper/Menu/ActionsMenuBuilder.php#L184

Numkil:
i am sure it would not be too difficult to do. I will add it to the roadmap.

PR #1994, opened 2018-05-30
[KunstmaanNodeBundle] add configuration option for export page template

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1904

### DIFF
--- a/src/Kunstmaan/NodeBundle/DependencyInjection/Configuration.php
+++ b/src/Kunstmaan/NodeBundle/DependencyInjection/Configuration.php
@@ -51,6 +51,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('publish_later_stepping')->defaultValue('15')->end()
                 ->scalarNode('unpublish_later_stepping')->defaultValue('15')->end()
                 ->booleanNode('show_add_homepage')->defaultTrue()->end()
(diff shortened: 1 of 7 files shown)


src/Kunstmaan/NodeBundle/DependencyInjection/Configuration.php
```diff
@@ -51,6 +51,7 @@ public function getConfigTreeBuilder()
                 ->scalarNode('publish_later_stepping')->defaultValue('15')->end()
                 ->scalarNode('unpublish_later_stepping')->defaultValue('15')->end()
                 ->booleanNode('show_add_homepage')->defaultTrue()->end()
+                ->booleanNode('enable_export_page_template')->defaultTrue()->end()
```
acrobat: Maybe set it default to false? This way, default it won't show up on production (or other env's). But when a frontender want to get the base template he can switch that option to `true` just on his version of the site.

What do you think?

sandergo90: @acrobat true! changes are done now 😉

Issue #1995, opened 2018-05-30
[AdminBundle] impossible to create administrators with the create user command


deZinc:
As the title states, it's currently impossible to create administrators using the kuma:user:create command.

PR #1996, opened 2018-05-30
[AdminBundle] fixed create user command #1995

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1995

### DIFF
--- a/src/Kunstmaan/AdminBundle/Command/CreateUserCommand.php
+++ b/src/Kunstmaan/AdminBundle/Command/CreateUserCommand.php
@@ -12,12 +12,16 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ChoiceQuestion;
 use Symfony\Component\Console\Question\Question;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/AdminBundle/Command/CreateUserCommand.php
```diff
@@ -100,20 +113,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         foreach ($groupNames as $groupName) {
-
-            if ((int)$groupName !== 0) {
-                $group = $em->getRepository('KunstmaanAdminBundle:Group')->findOneBy(array('name' => $groups[$groupName]->getName()));
-                $groupOutput .= $groups[$groupName]->getName() . ', ';
-            } else {
-                $group = $em->getRepository('KunstmaanAdminBundle:Group')->findOneBy(array('name' => $groupName));
-                $groupOutput .= $groupName . ', ';
-            }
+            $group = $em->getRepository('KunstmaanAdminBundle:Group')->findOneBy(array('name' => $groups[$groupName]->getName()));
```
sandergo90: Why is the group loaded again here? We already loaded them above.

deZinc: That's a very solid point, I'll fix that one too.

Issue #1975, opened 2018-05-14
Redirect loop on /[locale]/admin/change-password


jazzzpor:
Steps to reproduce:

1. Following the guide on http://kunstmaanbundlescms.readthedocs.io/en/latest/installation/ 
2. Log in on /nl/admin and enter the admin username and password that was generated during `bin/console kuma:generate:default-site`
3. Redirect loop occurs at /nl/admin/change-password

Edit:
Here are my relevant parameters:

```
parameters:
    defaultlocale: nl
    locale: nl
    multilanguage: true
    requiredlocales: nl
```

mwoynarski:
I've found/fixed the issue and submitted a pull request:
https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1985

In the meantime, you can resolve the issue locally by overriding two routes:

```
KunstmaanAdminBundle_user_change_password:
    path: /%kunstmaan_admin.admin_prefix%/settings/users/{id}/edit
    defaults: { _controller: fos_user.change_password.controller:changePasswordAction }
    methods: [GET, POST]
```

```
fos_user_change_password:
    path: /%kunstmaan_admin.admin_prefix%/profile/change-password
    defaults: { _controller: fos_user.change_password.controller:changePasswordAction }
    methods: [GET, POST]
```

PR #2000, opened 2018-05-31
[AdminBundle] Use FOS service ID instead of class

mwoynarski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1975 

FOS\UserBundle\Controller\ChangePasswordController requires constructor arguments and is registered as a service by ID rather than class - the routing config needs to call that controller by service ID, rather than class name.


acrobat:
I think this is already fixed in #2010 as we fixed the compatibility with fosuserbundle 2.1 and the custom overrides of the bundle routes.

PR #2004, opened 2018-06-03
[NodeBundle] Reword deprecation message and provide alternative to the end-user

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #1980 

I've improved the deprecation message so the end-user actually knows how to fix the deprecation.

### DIFF
--- a/src/Kunstmaan/NodeBundle/Entity/AbstractPage.php
+++ b/src/Kunstmaan/NodeBundle/Entity/AbstractPage.php
@@ -128,11 +128,10 @@ abstract class AbstractPage extends AbstractEntity implements PageInterface
     /**
      * {@inheritdoc}
      *
-     * @deprecated Using the service action is deprecated in KunstmaanNodeBundle 5.1 and will be removed in KunstmaanNodeBundle 6.0. Use the SlugActionInterface instead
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/NodeBundle/Entity/AbstractPage.php
```diff
@@ -128,11 +128,10 @@ public function getDefaultAdminType()
     /**
      * {@inheritdoc}
      *
-     * @deprecated Using the service action is deprecated in KunstmaanNodeBundle 5.1 and will be removed in KunstmaanNodeBundle 6.0. Use the SlugActionInterface instead
+     * @deprecated Using the service method is deprecated in KunstmaanNodeBundle 5.1 and will be removed in KunstmaanNodeBundle 6.0. Implement SlugActionInterface and use the getControllerAction method to provide custom logic instead.
      */
     public function service(ContainerInterface $container, Request $request, RenderContext $context)
     {
-        @trigger_error('Using the service action is deprecated in KunstmaanNodeBundle 5.1 and will be removed in KunstmaanNodeBundle 6.0. Use the SlugActionInterface instead', E_USER_DEPRECATED);
```
acrobat: Remove the deprecation here because it would trigger a deprecation warning even if the user has not defined the method in their child class.

PR #2011, opened 2018-06-08
[AdminBundle] Add missing use statements to fix broken acl setup

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

After the merge of PR #1999 the create of nodes was broken as no acl's were inserted for the object, due to an class not matching. I've added the use statements so it works again.


acrobat:
Fixed with merge of #2001

PR #2022, opened 2018-06-15
[DashboardBundle] Deprecated service container usage in dashboard com…

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |


UPGRADE-5.1.md
```diff
@@ -33,6 +33,18 @@ ConfigBundle
 
  * Passing the `container` as the sixth argument in `Kunstmaan\ConfigBundle\Controller\ConfigController` is deprecated in and will be removed in 6.0.
 
+DashboardBundle
+---------------
+
+ * Injecting the container in the `DashboardCommand` is deprecated and will be removed in 6.0. Inject the required parameters instead.
```
acrobat: The upgrade log should be more like this, as the current message is not correct (the container is not injected in the commands, first argument is actually an optional command name)

```
`XXXXCommand::__construct()` now takes an instance of `XXXX` as the first argument. Not passing it is deprecated and will throw a `TypeError` in 6.0.
```

Same for the other upgrade notes

PR #2026, opened 2018-06-18
[TranslatorBundle] add status field on translations for later use in rest api

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Allows translations to be "deprecated" and "disabled". 
This feature will be invisible for classic websites as the feature does not add much in their use case.

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
+++ b/src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\Query\QueryBuilder;
 use Kunstmaan\AdminListBundle\AdminList\Configurator\AbstractDoctrineDBALAdminListConfigurator;
 use Kunstmaan\AdminListBundle\AdminList\FilterType\DBAL\EnumerationFilterType;
 use Kunstmaan\AdminListBundle\AdminList\FilterType\DBAL\StringFilterType;
+use Kunstmaan\TranslatorBundle\Entity\Translation;
 
 /**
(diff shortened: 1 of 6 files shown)


Numkil:
@Devolicious Updated PR with your comments and added 2 repository tests.

src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
```diff
@@ -132,8 +135,10 @@ public function getQueryBuilder()
         if (is_null($this->queryBuilder)) {
             $this->queryBuilder = new QueryBuilder($this->connection);
             $this->queryBuilder
-                ->select('DISTINCT b.translation_id AS id, b.keyword, b.domain')
-                ->from('kuma_translation', 'b');
+                ->select('DISTINCT b.translation_id AS id, b.keyword, b.domain, b.status')
+                ->from('kuma_translation', 'b')
+                ->andWhere('b.status is null OR b.status != :statusstring')
```
Devolicious: Can you also remove the search on `NULL` here?

src/Kunstmaan/TranslatorBundle/Entity/Translation.php
```diff
@@ -309,4 +323,52 @@ public function getTranslationId()
...
+    public function getStatus()
+    {
+        return $this->status;
+    }
+
+    /**
+     * @param string $status
+     * @return Translation
+     */
+    public function setStatus(string $status): Translation
```
Devolicious: Remove return type. PHP 5.6 does not support it

PR #2027, opened 2018-06-18
[NodeBundle] Split off logic of NodeAdminController in helper class

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

In this PR some logic that is in the NodeAdminController is being put in a helper service class. Currently no changes in the controller yet: the controller will not yet use this code. It will be done in a future PR.


src/Kunstmaan/NodeBundle/Resources/config/services.yml
```diff
@@ -183,6 +183,17 @@ services:
             - "@kunstmaan_node.helper.url"
         public: true
 
+    kunstmaan_node.helper.node:
+        class: Kunstmaan\NodeBundle\Helper\NodeHelper
+        arguments:
+            - "@doctrine.orm.entity_manager"
+            - "@kunstmaan_node.admin_node.publisher"
+            - "@kunstmaan_node.admin_node.node_version_lock_helper"
```
Devolicious: @treeleaf This argument does not match the constructor of the `NodeHelper`

src/Kunstmaan/NodeBundle/Resources/config/services.yml
```diff
@@ -183,6 +183,17 @@ services:
...
+            - "@security.token_storage"
+            - "@kunstmaan_admin.clone.helper"
+            - "@event_dispatcher"
+            - { timeout: "%kunstmaan_node.version_timeout%"}
```
Devolicious: This parameter is also not available in the constructor method.

PR #2028, opened 2018-06-19
[MediaBundle] Fixed BackgroundFilterLoader issue

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

The BackgroundFilterLoader class as it was throws an error because the Color class no longer exists. Besides, overriding the BackgroundFilterLoader from liip is no longer necessary as the original issue is fixed in the imagine & liip versions the current bundle version uses.


acrobat:
@deZinc This is added to the 5.1.0 milestone, shouldn't it be for the 5.0.8?

Devolicious:
@acrobat @deZinc, It's kind of conflicting with the rules... Deprecations should only be done on master but it's also a bugfix which should happend on the latest minor branch. Basically it should be split in 2 seperate PR's. What do you guys think? Split up or #see_no_evil

acrobat:
I would say to split it!

PR #2039, opened 2018-06-25
[AllBundles] rollback sensio framework extra bundle version

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

We have to downgrade the sensio framework-extra-bundle because in v4.0 there's a BC break that causes issues with our templates. We'll have to rethink this further down the line for the sf4 compatibility.

> [BC BREAK] changed template name generation from camelCase to under_score for both files and directories

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "symfony/monolog-bundle": "~2.8|~3.0",
         "symfony/security-acl": "~2.8|~3.0",
         "sensio/distribution-bundle": "^5.0",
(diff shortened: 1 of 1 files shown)


acrobat:
Closing this PR as a fix for the BC break is provided in #2045 whiling keeping the latest version of the framework-extra-bundle

PR #2044, opened 2018-06-29
[NodeBundle] Add tabs

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no

Allows form to be split up into tabs. One FormType per tab.


src/Kunstmaan/NodeBundle/Entity/PageTabInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Kunstmaan\NodeBundle\Entity;
+
+use Kunstmaan\NodeBundle\ValueObject\PageTab;
+
+interface PageTabInterface {
```
mlebkowski: Code style for this repo requires newlines before opening class/interface/trait braces

src/Kunstmaan/NodeBundle/EventListener/NodeListener.php
```diff
@@ -45,10 +47,25 @@ public function __construct(AuthorizationCheckerInterface $authorizationChecker,
...
         if ($event->getPage() instanceof HasNodeInterface && !$event->getPage()->isStructureNode()) {
+			$page = $event->getPage();
+			$tabPane = $event->getTabPane();
+
             if ($this->authorizationChecker->isGranted('ROLE_PERMISSIONMANAGER')) {
-                $tabPane = $event->getTabPane();
                 $tabPane->addTab(new Tab('kuma_node.tab.permissions.title', new PermissionsFormWidget($event->getPage(), $event->getNode(), $this->permissionAdmin, $this->permissionMap)));
             }
+
+			if($page instanceof PageTabInterface) {
```
mlebkowski: This block is completely unrelated to the other block. In my opinion this should be split into two separate listeners to keep single responsibility. 

Then, if you have a separate listener, I see no reason for it to be in the core, you can as easily create a bundle for it and use in whichever project you’d like (even with older CMS versions)

src/Kunstmaan/NodeBundle/Tests/Entity/NodeTest.php
```diff
@@ -161,6 +170,36 @@ public function testToString()
...
+		$permissionMap = $this->getMock(PermissionMapInterface::class);
+
+		$entity = new TestEntity();
+		$this->object->setRef($entity);
+
+		$tabPane = new TabPane('id', $request, $formFactory);
+
+		$adaptFormEvent = new AdaptFormEvent($request, $tabPane, $entity);
+
+		$nodeListener = new NodeListener($authorizationChecker, $permissionAdmin, $permissionMap);
```
mlebkowski: as mentioned in my other comment, none of those dependencies are actually needed to test your behaviour, they are just noise.

src/Kunstmaan/NodeBundle/Tests/Entity/NodeTest.php
```diff
@@ -161,6 +170,36 @@ public function testToString()
         $this->assertEquals('node 1, refEntityName: Kunstmaan\NodeBundle\Tests\Entity\TestEntity', $this->object->__toString());
     }
 
+
+	public function testPageShouldHaveTab() {
+		$request = new Request();
+		$request->request = new ParameterBag();
+
+		$formFactory = $this->getMock(FormFactory::class, [], [], '', false);
```
mlebkowski: FYI, instead of a bunch parameters nobody knows the purpose for, you can write:

```php
$this->getMockBuilder(FormFactory::class)->disableOriginalConstructor()->getMock()
```

src/Kunstmaan/NodeBundle/Tests/Entity/TestEntity.php
```diff
@@ -87,4 +89,17 @@ public function isStructureNode()
...
+	/**
+	 * @return PageTab[]
+	 */
+    public function getTabs()
+	{
+		return [
+			(new PageTab())
+				->setTabTitle('tab1_title')
+				->setInternalName('tab1_name')
+				->setPosition(1)
```
mlebkowski: There is no `formType` configured, shouldn’t it be required? see my other comment 👇

src/Kunstmaan/NodeBundle/ValueObject/PageTab.php
```diff
@@ -0,0 +1,125 @@
+<?php
+
+namespace Kunstmaan\NodeBundle\ValueObject;
+
+class PageTab {
```
mlebkowski: value objects should not have setters, but require all of the fields in a constructor. This way you can be sure the object is always in a valid state, and there is no way for a programmer to make a mistake like this.

src/Kunstmaan/NodeBundle/EventListener/NodeListener.php
JZuidema: Thanks for your feedback. I'm a bit confused by one of your lines. You mention you see no reason for it to be in the core. Do you mean you do not want it in the NodeBundle (as in, create another bundle for it) or do you not want it in Kunstmaan at all?

mlebkowski: I don’t see why it should be in Kunstmaan at all, as you can easily get the same result by using an external bundle, installed separately.

JZuidema: One of the reasons why we offer Kunstmaan to our customers is because it's clear and user-friendly. Being able to split a form up into multiple tabs, rather than keeping all the input fields in the Content tab, would contribute greatly to giving customers a clear overview. We could include this separately, but we believe that it should be added into the core of Kunstmaan as other users might like to offer this feature as well.

mlebkowski: It’s just my opinion, but I’d rather see the core with a set of basic functionalities. You can publish your code on an MIT license with a two-line install instructions, so everyone can benefit. I have contributed a bunch of stuff here, but it’s mostly bug fixes or changes to the core, not new additions. If I have an idea for a feature that’s completely isolated I don’t hesitate to publish it as a separate package (and my organisation already has a dozen of those).

YMMV. And of course I am not the maintainer here, so they can have a different take on this.

Devolicious: @JZuidema @mlebkowski, all new features are a great addition. @mlebkowski I get what you mean but in cases like this we prefer to have in in the core, since it's more of an option to the `FormPage` itself. In cases like our `CookieBundle` we opted to make it a seperate bundle because it's completely isolated. If every option got it's own bundle, the power of the cms get's scattered around and users would all start to implement their own version of something that already exists. It's a thin line when you're wondering to make a seperate bundle or not but if you do, please add it to the `README`so other users can also find you're contributions :)

wesleylancel:
Can we get some docs on what this does exactly? :)

Devolicious:
@mlebkowski thx for the thorough review!

JZuidema:
@Devolicious I've applied the suggested changes by @mlebkowski. Also added a description in the UPGRADE-5.1.md, although I'm not exactly sure this was the description you're looking for.

Anyway, in short, some of entities had quite some fields in it, which made the admin add/edit view rather chaotic. With these changes, you can add tabs to the already existing tab menu at the top, and split your form up into smaller forms.

Devolicious:
@JZuidema can you also provide a bit of documentation in [docs/bundle/node-bundle](https://github.com/Kunstmaan/KunstmaanBundlesCMS/tree/master/docs/bundles/node-bundle)? Maybe create a new md file, `split-page-into-tabs.md` or something like that and then you can refer to it in your text you've put in the upgrade guide.

This is a very useful feature so I think users will benefit from it a lot. Or better yet, you can also add it as a cookbook article with some code examples on how to use your feature. That would be awesome if you could do that.

Thx!

mlebkowski:
For what it’s worth, in our recent project we implemented the exact same feature. The only noticeable difference is that we simplified the interface — the page returns a map: `$tabName => $formType`, but other than that, it’s basically the same.

JZuidema:
@Devolicious I have added a description of what it does and how to use it in src/Kunstmaan/NodeBundle/entity-tabs.md. Could you review it?

sandergo90:
@Devolicious can you review again?

Issue #580, opened 2015-07-23
`createModal.html.twig` template is missing


mlebkowski:
![Template missing](https://www.evernote.com/l/AAR11iM1ACdGs5OyhLuNL4HD_HB09WIifP0B/image.png)

`\Kunstmaan\MediaBundle\Controller\MediaController::createModalAction` throws an exception because the `createModal.html.twig` template is missing

yoshz:
Had this exception a few times, but couldn't reproduce it quickly.
Seems like to occur when you try to add a new file from an existing modal.

jeroenjans:
I was able to reproduce the issue by uploading a file that was bigger than the `upload_max_filesize` directive in php.ini (which in my case was 2M)

aistis-:
I also run into the same issue.

jockri:
I think this was solved in #580. If not please reopen the issue.

PR #2045, opened 2018-06-30
[AllBundles] Fix BC break after upgrading to sensio/framework-extra-bundle 5.0

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | closes #2039 (can be closed as the downgrade is not required anymore with this fix)

Upgrading to sensio/framework-extra-bundle 5.0 caused a bc break for camelcase template paths which were not explicitly configured (Symfony would guess the template path for the empty `@Template()` annotations. This is fixed by explicitly settings the template path for these cases. 

I've also remove some unused `@Template` annotations as these actions returned a redirect so the annotation was unused.

### DIFF
--- a/src/Kunstmaan/MediaBundle/Controller/MediaController.php
+++ b/src/Kunstmaan/MediaBundle/Controller/MediaController.php
@@ -122,7 +122,7 @@ class MediaController extends Controller
      * @param int $folderId
      *
      * @Route("bulkupload/{folderId}", requirements={"folderId" = "\d+"}, name="KunstmaanMediaBundle_media_bulk_upload")
-     * @Template()
+     * @Template("@KunstmaanMedia/Media/bulkUpload.html.twig")
      *
      * @return array|RedirectResponse
      */
@@ -141,7 +141,6 @@ class MediaController extends Controller
      * @param int     $folderId
      *
      * @Route("bulkuploadsubmit/{folderId}", requirements={"folderId" = "\d+"}, name="KunstmaanMediaBundle_media_bulk_upload_submit")
-     * @Template()
      *
      * @return JsonResponse
      */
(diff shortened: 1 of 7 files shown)


src/Kunstmaan/MediaBundle/Controller/MediaController.php
```diff
@@ -438,7 +436,6 @@ private function createAndRedirect(Request $request, $folderId, $type, $redirect
      *
      * @Route("create/modal/{folderId}/{type}", requirements={"folderId" = "\d+", "type" = ".+"}, name="KunstmaanMediaBundle_media_modal_create")
      * @Method({"GET", "POST"})
-     * @Template()
      *
```
treeleaf: method 'createAndRedirect' can still return an array. In that case it could break. So either 'createAndRedirect' should be adapted or a default template should be provided.

acrobat: Good catch! I'm not sure this method is actually used anymore. I can only find 1 usage in the code, but I don't know yet how to "trigger" that usage. Currently there is no corresponding twig template for this action, so I will look into this.

acrobat: This action was not used as a "GET" anywhere only as a POST for the add file/image/video popup. So I've removed GET as allowed method. Also this code couldn't have worked before as the is no template `createModal.html.twig` or `create_modal.html.twig`

acrobat:
This PR is ready for review/merge

src/Kunstmaan/MediaBundle/Controller/MediaController.php
mlebkowski: Thanks for fixing a [three year old bug](https://github.com/Kunstmaan/KunstmaanBundlesCMS/issues/580)

PR #1981, opened 2018-05-21
[AdminBundle] Extend Role from base symfony class

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Directly implementing the `RoleInterface` is deprecated instead we should extend from the `Role` class. This is no bc break as the `BaseRole` implements the `RoleInterface`

PR #2046, opened 2018-07-01
[AdminBundle] Add compatibility with the new symfony Role class

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

I've fixed the `RoleInterface` deprecation in #1981, this PR fixes an incompatibilty issue in the `PermissionAdmin` where the `Role` object is checked.


acrobat:
@sandergo90 If a users has a custom `Role` class and doesn't extend from the `Role` class but still implements the interface the check on the `Role` class will not pass. So keeping this for BC for now I think and we can remove this when we drop sf3.4 support.

Example: https://3v4l.org/6iidr

sandergo90:
Alright, good point! 👌

PR #2047, opened 2018-07-01
[PagePartBundle] Allow bundless config for pagepart and page yaml files paths

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

With the current code it's required to have a bundle namespace in the yaml path config (eg. `WebsiteBundle:homepage`). With this pr we allow "bundless" paths, this also to have compatibilty with sf4 where the concept of bundles in the `src/` dir is not recommended anymore. 

So the locator will look in 2 possible paths and will return the first hit it finds.
For a bundless setup in sf 3.4 it will look into `src/Resources/config/(pageparts|pagetemplates)`. And for the sf 4 direcoty structure it will look into `config/kunstmaancms/(pageparts|pagetemplates)`.


mlebkowski:
Or you could just deprecate all of it and move to [semantic configuration](https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1109) instead. This solution needlessly parses the yaml file (multiple files even) on each request

acrobat:
@mlebkowski hmm I actually didn't know about that semantic configuration was a valid option. Let me look into it, as it might indeed be better to deprecate the old system and move fully to the new way.

PR #2048, opened 2018-07-03
[AdminBundle] Remove id of nested form item containers

SpadXIII:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Nested item containers conflict with page parts, because they use the same id: #-pp-container
As far as I could tell, the id of this element is never used, but the id of page part container elements is.

I have a page part with a (one to many) collection field. When the page has multiple page parts, deleting the second one actually triggers the removal of the second item of the collection field, while also marks the page part for deletion. When saving, this results in a error: Call to a member function setContext() on null at vendor/kunstmaan/bundles-cms/src/Kunstmaan/PagePartBundle/Repository/PagePartRefRepository.php:195.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Form/fields.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Form/fields.html.twig
@@ -183,7 +183,7 @@
 
                 {# Items #}
                 {% for obj in form %}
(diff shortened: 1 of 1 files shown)


SpadXIII:
Closed because I used master instead of the 5.0 branch. Will start over again!

PR #2055, opened 2018-07-17
[KunstmaanTranslatorBundle] export all translations, use correct locales

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

- When exporting translations to a csv file, only the messages domain is exported. For now we removed the messages limit. If people want to have the possibility to specify the domains to export, a PR can be created for this.

- When exporting translations, the requiredlocales parameter is being used but this should be the managed locales. In the dependency injection, i've also added the requiredlocales as default for the managed_locales parameter.

- Formatting


sandergo90:
Tests are fine now!

PR #2058, opened 2018-07-25
[MediaBundle] Add web_root configuration to support both web and public as the root dir for uploads

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

In sf4 the web root was moved from `web` to `public` but the web directory was hardcoded for the full media upload path in our code. This change adds a configuration option to the mediabundle to override the web root path. Default it will use the symfony kernel version to set a sensible default value, `web` for <sf4 and `public` for >sf4.


sandergo90:
@acrobat Thanks for adding new tests! It will only make our CMS better!

PR #2068, opened 2018-08-02
[AllBundles] Support symfony/swiftmailerBundle 3

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

To support sf4 we also need to allow `symfony/swiftmailer-bundle` 3.0. But swiftmailer 3 only supports swiftmailer 6. To make sure our code works with both swiftmailer 5 and 6 I've modified the code so we are using the forward compatibility layer.


acrobat:
First commit adds a new test (which will fail) and second commit does the code changes and will make the test pass 🎉

PR #2074, opened 2018-08-09
[NodeBundle] Use nodeHelper in NodeAdminController

treeleaf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

This is a followup on #2027 where the NodeHelper service is used in the NodeAdminController.

### DIFF
--- a/src/Kunstmaan/NodeBundle/Controller/NodeAdminController.php
+++ b/src/Kunstmaan/NodeBundle/Controller/NodeAdminController.php
@@ -346,9 +346,25 @@ class NodeAdminController extends Controller
 
         $this->denyAccessUnlessGranted(PermissionMap::PERMISSION_DELETE, $node);
 
-        $response = $this->nodeHelper->deletePage($node, $this->locale);
+        $nodeTranslation = $node->getNodeTranslation($this->locale, true);
+        $nodeVersion = $nodeTranslation->getPublicNodeVersion();
+        $page = $nodeVersion->getRef($this->em);
 
-        if (null === $response) {
+        $this->get('event_dispatcher')->dispatch(
+            Events::PRE_DELETE,
+            new NodeEvent($node, $nodeTranslation, $nodeVersion, $page)
+        );
+
+        $node->setDeleted(true);
+        $this->em->persist($node);
+
+        $children = $node->getChildren();
+        $this->deleteNodeChildren($this->em, $this->user, $this->locale, $children);
+        $this->em->flush();
+
+        $event = new NodeEvent($node, $nodeTranslation, $nodeVersion, $page);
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/NodeBundle/Helper/NodeHelper.php
```diff
@@ -264,12 +265,12 @@ public function deletePage(Node $node, $locale)
         $this->deleteNodeChildren($node, $locale);
         $this->em->flush();
 
-        $this->eventDispatcher->dispatch(
+        $event = $this->eventDispatcher->dispatch(
             Events::POST_DELETE,
             new NodeEvent($node, $nodeTranslation, $nodeVersion, $page)
         );
 
-        return $nodeTranslation;
```
Devolicious: BC?

src/Kunstmaan/NodeBundle/Controller/NodeAdminController.php
```diff
@@ -685,31 +550,14 @@ public function addHomepageAction(Request $request)
...
+        $nodeTranslation = $this->nodeHelper->createPage($type, $title, $this->locale);
+        $nodeNewPage = $nodeTranslation->getNode();
 
         // Set default permissions
-        $this->container->get('kunstmaan_node.acl_permission_creator_service')
-            ->createPermission($nodeNewPage);
-
-        $nodeVersion = $nodeTranslation->getPublicNodeVersion();
-
-        $this->get('event_dispatcher')->dispatch(
```
Devolicious: BC? Order when the event is dispatched has been changed.

src/Kunstmaan/NodeBundle/Helper/NodeHelper.php
treeleaf: Currently I don't see how to do it without BC. We can modify later that we use a listener instead of handling the response in the controller..

src/Kunstmaan/NodeBundle/Controller/NodeAdminController.php
treeleaf: I have adapted the NodeHelper (doing more BC) so that the event is dispatched at the same momenten as before the change.

treeleaf:
I have further modified the NodeHelper (BC breaking). I don't see how we can do this in a clean way without breaking the NodeHelper. As suggested by @acrobat, I have added an "experimental" annotation, like they do with Symfony. With this annotation we mark that this class can still be changed in this release.

acrobat:
PR is outdated and will need extra work when actually implementing this.

PR #2091, opened 2018-08-30
[NodeSearchBundle] Fix search index of childpages of a structured node

### LINKED ISSUE #2083: [NodeSearchBundle] Saving a page under a structered node doesn't trigger ES index
When saving a page that is nested under a StructuredNode, the save action won't trigger an ES index of the page because the index logic checks if all parent pages are online but a structured node is offline by default. 

The fix is to adapt the check in `NodeIndexUpdateEventListener:: hasOfflineParents` to check if the page is online or if it's a structured node. See https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/e21649aaac51269a1957fe385e1bac92050cbccf/src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php#L115..L125

Reference: https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1550/files#diff-f17d05895e6833712a319565f1bda8acR696

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | fixes #2083 
| Tests pass? | Yes 🎉 

Indexing child pages of a structured node (or any of the parents is a structured node) was broken because structured nodes are always offline. Added a check to allow structured node parents to be offline. I've also added tests to check the expected behavior.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "behat/mink-goutte-driver": "dev-master",
         "behat/mink-sahi-driver": "dev-master",
         "symfony/phpunit-bridge": "~3.0",
-        "phpunit/phpunit": "~4.4",
+        "phpunit/phpunit": "^5.7",
         "fzaninotto/faker": "~1.6",
         "nelmio/alice": "^2.1.4"
     },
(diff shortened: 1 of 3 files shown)


sandergo90:
@acrobat builds failing

acrobat:
Phpunit 4 :( Bumped the version to 5 just to get the tests passing again. On the master branch we're already using phpunit 5

mlebkowski:
Why are Structure Pages marked as offline? I have a listener that flips them online after save and I’ve been using it for over three years now, see #702

Issue #2033, opened 2018-06-21
[AllBundles] Update composer dependencies of the individual packages


acrobat:
To support symfony 4, and as general good practice, we shouldn't require symfony/symfony. Instead require the individual symfony packages.

So the best aproach is to check all the composer files of each individual bundle and make sure these are up to date. After that we can set the correct dependencies in the main `composer.json`.

TODO:
- [x] AdminBundle (#2249)
- [x] AdminListBundle (#2251)
- [x] ArticleBundle (#2254)
- [x] BehatBundle (#2251)
- [x] CacheBundle (#2259)
- [x] ConfigBundle (#2251)
- [x] DashboardBundle (#2258)
- [x] FixturesBundle (#2272)
- [x] FormBundle (#2255)
- [x] GeneratorBundle (#2257)
- [x] LeadGenerationBundle (#2270)
- [x] MediaBundle (#2251)
- [x] MediaPagePartBundle (#2256)
- [x] MenuBundle (#2260)
- [x] MultiDomainBundle (#2261)
- [x] NodeBundle (#2251)
- [x] NodeSearchBundle (#2265)
- [x] PagePartBundle (#2251)
- [x] RedirectBundle (#2269)
- [x] SearchBundle (#2262)
- [x] SeoBundle (#2093)
- [x] SitemapBundle (#2092)
- [x] TaggingBundle (#2082)
- [x] TranslatorBundle (#2084)
- [x] UserManagementBundle (#2037)
- [x] UtilitiesBundle (#2248)
- [x] VotingBundle (#2034)

acrobat:
All separate bundles are done.

PR #2092, opened 2018-08-31
[SiteMapBundle] update composer dependencies

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2033


acrobat:
These dependencies should be added to the `required-dev` section as they are required to run the tests.

```
"codeception/codeception": "^2.4",
        "matthiasnoback/symfony-config-test": "^4.0",
        "matthiasnoback/symfony-dependency-injection-test": "^3.0"
```

Tests can start now but won't actually run now because the bundles are no separate packages yet. And for example the node-bundle is required in the tests.

sandergo90:
@acrobat Added the dev dependencies.

PR #2098, opened 2018-09-06
[MediaBundle] Liip imagine bundle version 2 support

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

~~_**Only merge after #2090. Ping me after merge to rebase this PR**_~~

~~Diff best viewed against the prepare-5.2 branch [prepare-5.2-php-changes...liip-imagine-bundle-2-support](https://github.com/acrobat/KunstmaanBundlesCMS/compare/prepare-5.2-php-changes...acrobat:liip-imagine-bundle-2-support).~~

#2090 is merged, rebase done.

This PR upgrades the liip/imagine-bundle to version 2, as this is the first version that supports sf4. All custom overrides made in the kunstmaan/media-bundle are ported to the new version.

### DIFF
--- a/src/Kunstmaan/MediaBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/MediaBundle/Resources/config/services.yml
@@ -125,3 +125,37 @@ services:
         calls:
             - [setMimeTypeGuesser, ['@kunstmaan_media.mimetype_guesser.factory']]
             - [setExtensionGuesser, ['@kunstmaan_media.extension_guesser.factory']]
+
+    kunstmaan_media.imagine.filter.loader.background:
+        class: '%liip_imagine.filter.loader.background.class%'
+        arguments:
+            - '@liip_imagine'
+        tags:
+            - { name: 'liip_imagine.filter.loader', loader: 'background' }
+
+    Kunstmaan\MediaBundle\Helper\Imagine\ImagineController:
+        arguments:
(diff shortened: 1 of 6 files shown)


src/Kunstmaan/MediaBundle/Resources/config/services.yml
```diff
@@ -125,3 +125,40 @@ services:
...
+        class: Kunstmaan\MediaBundle\Helper\Imagine\ImagineController
+        arguments:
+            - '@liip_imagine.service.filter'
+            - '@liip_imagine.data.manager'
+            - '@liip_imagine.cache.signer'
+        tags:
+            - { name: 'controller.service_arguments' }
+        public: true
+
+    kunstmaan_media.imagine.resolver.web_path:
```
sandergo90: Do we need still the service id's? Can't we just use the classname as service id?

acrobat: Good point, I will use fqcn's for the new services!

acrobat: @sandergo90 Fixed! Only the `kunstmaan_media.imagine.filter.loader.background` service could not renamed as the class is based on a parameter. This could be updated in v6

PR #2101, opened 2018-09-06
[AdminBundle] Fixed issue with locale in kuma:user:create command

veloxy:
When the locale is passed to the kuma:user:create command, it is being overwritten by the default local due to an incorrect if statement.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


veloxy:
Closing and re-submitting for 5.0

PR #2104, opened 2018-09-06
[AdminBundle] Fixed issue with locale in kuma:user:create command

veloxy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When the locale is passed to the kuma:user:create command, it is being overwritten by the default local due to an incorrect if statement.


veloxy:
Not sure why it still says "It seems that our checklist is missing or incomplete"

Issue #2096, opened 2018-09-05
[FormBundle] add possibility to generate form pageparts to your own bundle


sandergo90:
When working on a project with form pages, most of the time the form pageparts need to be changed with custom stuff. It would be best practice to be able to generate the formpageparts to your own bundle.

PR #2106, opened 2018-09-07
[GeneratorBundle, FormBundle] add generator for form pageparts

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2096

When generating a form page, you will get the option to generate all the form pageparts to your own custom bundle. I've also added a "internal name" field to all form pageparts. By filling in the internal name of a form pagepart, you will get the option to do something with it when using the events coming from the form submission.

### DIFF
--- a/src/Kunstmaan/FormBundle/Entity/FormSubmissionField.php
+++ b/src/Kunstmaan/FormBundle/Entity/FormSubmissionField.php
@@ -3,7 +3,6 @@
 namespace Kunstmaan\FormBundle\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
-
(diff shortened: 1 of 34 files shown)


acrobat:
@sandergo90 The default answer for the question "Do you want to generate default form pageparts in your bundle?" is yes, but what is the behavior if the pageparts already exist in your code? Does the `kuma:generate:form-pageparts` stop or does it overwrite your files?

sandergo90:
@acrobat It will override your files like all our generator do ;)

Issue #2063, opened 2018-08-01
[Tests] codeception tests are not in proper directory structure for codeception.


Numkil:
[Tests] codeception tests are not in proper directory structure for codeception.

Tests need to be moved in a sub folder unit under folder Tests. and not directly in it.

PR #2107, opened 2018-09-10
[FormBundle] [Tests] Fix namespace + modernize old fashioned mocks and use codeception

Numkil:
This fixes the issue where codeception could not run individual test classes for this bundle. Also gets rid of some old fashioned mocks that were causing namespace issues. 

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2063


acrobat:
@Numkil with the new `unit` dir, should we update our namespaces too? `Kunstmaan\FormBundle\Tests\Unit` or change the composer autoloading so it maps `Kunstmaan\FormBundle\Tests` to the unit dir? Mapping it to the unit dir will give us trouble when adding other types of tests

PR #2115, opened 2018-09-10
[NodeBundle] Node choice type fix

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

`NodeChoiceType` should always check the deleted flag, online status and language, not only when a `page_class` is passed as an option.


deZinc:
I requested to revert this in PR #2133 , further explanation can be found there.

PR #2120, opened 2018-09-12
[NodeSearchBunde] [SearchBundle]  Add parameters for ES authentication and use them in version check

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/DependencyInjection/KunstmaanNodeSearchExtension.php
+++ b/src/Kunstmaan/NodeSearchBundle/DependencyInjection/KunstmaanNodeSearchExtension.php
@@ -56,7 +56,19 @@ class KunstmaanNodeSearchExtension extends Extension implements PrependExtension
     {
         $hosts = [];
         if ($container->hasParameter('kunstmaan_search.hostname') && $container->hasParameter('kunstmaan_search.port')) {
-            $hosts[] = $container->getParameter('kunstmaan_search.hostname').':'.$container->getParameter('kunstmaan_search.port');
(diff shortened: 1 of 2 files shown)


mlebkowski:
Please stop this madness. How much more fixes for this feature are you willing to make? 

1. Your symfony semantical configuration should *not* depend on any runtime dependencies / services. This is just wrong. There is no point of defining the configuration format in runtime, basing on the current ES version — you have your config in `app/config` for either `ES 1.x` or `ES 6.x`. If you switch ES version in runtime, your config will break if it doesn’t match.
2. The [build and run steps should be separated](https://12factor.net/build-release-run). I don’t want my ES instance running during build step!
3. You should not rely on container parameters. Those should be used *only* to fill out the semantic configuration, so:

    ```yaml
     parameters:
        kunstmaan.elasticsearch: http://foo:bar@elastic:9200/prefix
     kunstmaan_search_bundle:
          hosts:
             - "%kunstmaan.elasticsearch%"
    ```

    Then you can use your Bundle Extension to convert this configuration into your own private parameter, so you can configure your services for example.

4. You should not repeat the logic of building the elasticsearch DSN. This is why you have so many bugs you have to fix.
5. Please don’t introduce this kind of things in patch versions… :(

All of the above combines and enforces the first point: you should not do this kind of things in bundle extension. If you want to support numerous configuration formats, add two separate branches for it and use one or another, users choice! Or transform the old format into new format automatically! 

BTW, why ES version larger than 2 is automatically assumed 6?

PR #2124, opened 2018-09-13
[MediaBundle] Folder types as constants

SpadXIII:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

I noticed a difference between the FolderType rel-field and the ones added in the fixtures. Extracted all used values into a class with constants.

ps. I'm not sure why, but it's showing a change in fields.html.twig (src/Kunstmaan/AdminBundle/Resources/views/Form/fields.html.twig) again, while that's a change already merged into 5.0 ?


SpadXIII:
Someday I'll get it right the first time... will start over again :)

PR #2133, opened 2018-09-18
[NodeBundle] Revert "Fix bug in NodeChoiceType (#2115)"

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This reverts commit 31d0cbb67dddb448656a01670017b1b625be615a.

This reverts pull request #2115 because it's current implementation is BC breaking. In the NodeChoiceType, it's possible to overwrite the querybuilder with a custom querybuilder, for example:
```
        $builder
            ->add('node', NodeChoiceType::class, [
                'required' => true,
                'multiple' => false,
                'attr' => [
                    'class' => 'js-advanced-select',
                ],
                'label' => 'Some label',
                'query_builder' => function (NodeRepository $repo) use ($options) {
                    return $repo->createQueryBuilder('n')
                        ->innerJoin('n.nodeTranslations', 'nt')
                        ->innerJoin('nt.publicNodeVersion', 'nv')
                        ->andWhere(.....
                        ;
                },
            ]) 
```
With the the version of the NodeChoiceType as in PR #2115 , the page is broken because:

> [Semantical Error] line 0, col 259 near 'nt INNER JOIN': Error: 'nt' is already defined.

@wesleylancel Could you have another look at this issue, trying to find a non BC breaking solution?

### DIFF
--- a/src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php
+++ b/src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php
@@ -42,20 +42,18 @@ class NodeChoiceType extends AbstractType
                 $queryBuilder = call_user_func($queryBuilder, $options['em']->getRepository($options['class']));
             }
 
-            $queryBuilder
-                ->select('n, nt')
-                ->innerJoin('n.nodeTranslations', 'nt')
-                ->innerJoin('nt.publicNodeVersion', 'nv')
-                ->andWhere('nt.online = :online')
-                ->andWhere('nt.lang = :lang')
-                ->andWhere('n.deleted != 1')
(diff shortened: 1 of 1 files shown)


acrobat:
Possible fix could be to move the extra checks (like is node online, not deleted etc) to [the default `query_builder`](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/31d0cbb/src/Kunstmaan/NodeBundle/Form/NodeChoiceType.php#L34..L36)

And change the transformer to only add the `refEntityName` where if the `page_class` option is set. This way the default choice type will always work (and will show only online, not deleted nodes) and if someone specified a custom qb, like the example, it won't break.

PR #2143, opened 2018-09-24
[AdminListBundle] Feature/sortable adminlist

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

When using an adminlist for pages, sorting does not work. The $entityId is the id of the NodeTranslation, not the entity. Which means that the page is never found, because $repo = $em->getRepository($configurator->getRepositoryName()); returns the repository of the page.
This fix checks if the adminlist is for a page, if it is then it should always use the NodeTranslationRepository, if not then it uses the repository of the entity.


JZuidema:
@Numkil I'm confused, is it okay now?

PR #2158, opened 2018-10-10
[MediaBundle] Added image sizes to media chooser

jordanmoon:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? |no
| Fixed tickets | n/a

### DIFF
--- a/src/Kunstmaan/MediaBundle/Resources/views/Chooser/chooserShowFolder.html.twig
+++ b/src/Kunstmaan/MediaBundle/Resources/views/Chooser/chooserShowFolder.html.twig
@@ -217,7 +217,10 @@
                                             <i class="fa fa-file-o media-thumbnail__icon"></i>
                                         {% endif %}
                                         <figcaption class="media-thumbnail__caption">
-                                            {{ media.name }}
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/MediaBundle/Resources/views/Chooser/chooserShowFolder.html.twig
```diff
@@ -217,7 +217,10 @@
                                             <i class="fa fa-file-o media-thumbnail__icon"></i>
                                         {% endif %}
                                         <figcaption class="media-thumbnail__caption">
-                                            {{ media.name }}
+                                            {{ media.name }}<br>
+                                            {% if media.metadata.original_width is not empty and media.metadata.original_height %}
```
sandergo90:
```suggestion
                                            {% if media.metadata.original_width is not empty and media.metadata.original_height is not empty %}
```
Can you add the not empty check at the end? thanks!

jordanmoon:
@sandergo90 Thanks for pointing that out. I've committed that in. Cheers!

PR #2162, opened 2018-10-17
[NodeBundle] Improved error shown when a page does not have a view path

veloxy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Better error message when view (path) is empty


Numkil:
@veloxy Thanks! this is indeed easier to problem solve.

Issue #2153, opened 2018-10-08
It seems to be possible to queue a page for "publish later" with a date in the past.


Numkil:
It seems to be possible to queue a page for "publish later" with a date in the past.

<img width="948" alt="screen shot 2018-10-08 at 11 24 39" src="https://user-images.githubusercontent.com/5604686/46601263-dd60c780-caec-11e8-9754-fd682c390a99.png">

PR #2163, opened 2018-10-17
[NodeBundle] Fixed issue where date in passed could be selected when publishing later

veloxy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2153

### DIFF
--- a/src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/Modals/_publish.html.twig
+++ b/src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/Modals/_publish.html.twig
@@ -28,7 +28,7 @@
                     <div id="publish-later">
                         <!-- Date -->
                         <div class="form-group">
-                            <div class="js-datepicker input-group" data-default-date="true">
+                            <div class="js-datepicker input-group" data-min-date="today" data-default-date="true">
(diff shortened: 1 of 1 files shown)


Numkil:
@veloxy  This is a good start but i think it is also necessary to add some sort of validation in the backend. Otherwise people could still tamper with the request and publish in the past while avoiding the ui. 
I think you could take a look at this file. ' src/Kunstmaan/NodeBundle/Helper/NodeAdmin/NodeAdminPublisher.php' and maybe improve the 'chooseHowToPublish' method. 
Thanks for helping out with the issues on the kunstmaanbundles!

veloxy:
@Numkil that is very true and would be the ideal case. But this is an admin area and isn't exposed to any kind of user, and highly unlikely to users that have the skills or bad intentions to do such things. I'll take a look but can not guarantee I have the time to implement server-side validation and properly handle it. Maybe a separate issue to handle is serverside would be a better option? Scheduled unpublish has the same issue and it also limited to client-side validation.

Numkil:
@veloxy 
OK i have decided to make a seperate issue for the backend validation. That way i can merge this now as a preliminary fix. 

Thanks again for taking the time to make a  pull request.

PR #2182, opened 2018-11-30
[AdminBundle] Catch InvalidArgumentException in MediaTokenTransformer

SpadXIII:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When (reverse-)transforming content in the MediaTokenTransformer, the crawler can throw an InvalidArgumentException-exception.
For example with just a javascript-tag as content.

Surrounded the html()-calls with a try-catch that returns the content if the exception is caught.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
+++ b/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
@@ -10,84 +10,84 @@ use Symfony\Component\Form\DataTransformerInterface;
  */
 class MediaTokenTransformer implements DataTransformerInterface
 {
-	/**
-	 * @param mixed $content
-	 *
-	 * @return string
-	 */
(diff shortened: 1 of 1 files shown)


sandergo90:
@SpadXIII Thanks for the fix, maybe we should throw the error message with the logger instead. This way a tool like Sentry can still catch the errors, and a developer can check ?

SpadXIII:
@sandergo90 throwing it into the logger sounds like a good idea, yes

I don't think this error would happen frequently though; the developed solution that made this error appear is not a good one :)


ps. would there be another way to get this error? If there is not, I'm ok with closing this pull request

sandergo90:
Maybe we should catch all errors here and log them instead of only the InvalidArgumentException.

PR #2185, opened 2018-12-03
[AdminBundle] Remove unused RoleInterface on the group entity

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

I wouldn't consider removing the interface as a "real" BC break the code for users keeps working and the `RoleInterface` shouldn't be used the indentify the `Group` entity in the first place. If we do consider this a BC break, I can rewrite the path to something [like they do in symfony/panthere](https://github.com/symfony/panther/blob/master/src/PantherTestCase.php#L19-L29).


sandergo90:
When upgrading to 5.2, we force users to change code, so maybe we should see it as a BC break?

acrobat:
Yeah, you're right. I will provide a BC layer!

acrobat:
@sandergo90 I've add a BC layer for this change!

PR #2187, opened 2018-12-04
[NodeBundle] Fixed schedule publish route typo

jordanmoon:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Deprecations? |no
| Fixed tickets | n/a

Currently publishing pages later breaks due to a typo in routes. This is a temporary fix for that. It looks like the routing needs checking in `Kunstmaan\NodeBundle\Controller\NodeAdminController`.

### DIFF
--- a/src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/edit.html.twig
+++ b/src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/edit.html.twig
@@ -18,7 +18,7 @@
                     |raw
                 }}
             {% endif %}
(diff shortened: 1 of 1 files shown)


acrobat:
@jordanmoon do you override this controller/action? Because the route name used in the twig file should be the correct one. See https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/fe8c716493b73459b8efe88b97c35c29ae449867/src/Kunstmaan/NodeBundle/Controller/NodeAdminController.php#L354

On master it has the same route name

jordanmoon:
@acrobat Apologises, having a further look, we had a bundle included that overrode that route which had the typo in! Sorry about that. Will remove the PR.

acrobat:
@jordanmoon No problem! Glad we sorted it out!

PR #2199, opened 2018-12-13
[AdminBundle] Update _form-group.scss

rmibelgium:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

In our cms, rawHtml type didn't respect the form-group--no-max-width because the textarea under has class form-control which overwrite with max-width: 40rem;

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/ui/scss/default-theme/components/forms/_form-group.scss
+++ b/src/Kunstmaan/AdminBundle/Resources/ui/scss/default-theme/components/forms/_form-group.scss
@@ -12,7 +12,7 @@
     max-width: 40rem;
 }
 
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/AdminBundle/Resources/ui/scss/default-theme/components/forms/_form-group.scss
```diff
@@ -16,6 +16,10 @@
     max-width: none;
 }
 
+.form-group--no-max-width textarea {
```
FVKVN: Could you group both selectors instead of duplicating the style declaration.

PR #2061, opened 2018-07-31
[AllBundles] Require individual symfony packages

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This PR needs to be merged after #2060 (rebased needed). Diff is best viewed as [a0d861a...f04f0fa](https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/2061/files/f04f0fa5526f248d5b6b955553cb7707271bb1f4). **This PR is also best to be merged with a merge commit (not a squash merge)**, to preserve the specific commits for better history.

Don't require the `symfony/symfony` package as this is not recommended anymore. Instead only require the specific packages used by the cms.

PR #2209, opened 2019-01-03
[AllBundles] added sf4 composer

deZinc:
WIP to use SF4 in composer


jockri:
@deZinc can you remove all config changes and only keep the changes in composer.json?

deZinc:
@jockri : done

composer.json
```diff
@@ -64,14 +64,17 @@
         "knplabs/knp-gaufrette-bundle": "~0.1",
         "symfony-cmf/routing-bundle": "~2.0",
         "fpn/doctrine-extensions-taggable": "~0.9",
+        "twig/twig": "^1.12|^2.0",
```
acrobat: This line should be removed, correct `twig/twig` dependency is set on line 68/69

composer.json
```diff
@@ -64,14 +64,17 @@
...
         "twig/twig": "^1.36|^2.6",
         "twig/extensions": "~1.0",
         "egulias/email-validator": "^1.2.8|^2.0",
         "box/spout": "^2.5",
         "ruflin/elastica": "^5.1|^6.0",
         "behat/transliterator": "~1.2",
-        "defuse/php-encryption": "v2.1.0"
+        "defuse/php-encryption": "v2.1.0",
+        "symfony/maker-bundle": "^1.9",
+        "symfony/monolog-bundle": "^3.3"
```
acrobat: Do we need the monolog bundle? As it was removed in #2061

composer.json
```diff
@@ -64,14 +64,17 @@
...
         "sensio/generator-bundle": "~3.0",
         "twig/twig": "^1.36|^2.6",
         "twig/extensions": "~1.0",
         "egulias/email-validator": "^1.2.8|^2.0",
         "box/spout": "^2.5",
         "ruflin/elastica": "^5.1|^6.0",
         "behat/transliterator": "~1.2",
-        "defuse/php-encryption": "v2.1.0"
+        "defuse/php-encryption": "v2.1.0",
+        "symfony/maker-bundle": "^1.9",
```
acrobat: Do we need the maker bundle now? I don't think we use a maker in the current code

composer.json
```diff
@@ -129,4 +132,4 @@
             "dev-master": "5.2-dev"
         }
     }
-}
+}
```
acrobat: Missing newline

PR #2216, opened 2019-01-11
[RedirectBundle] Refactor RedirectRouter without changing entity data and using path m…

jverdeyen:
There is a security issue in the current RedirectRouter which changes the Redirect object' origin and target datamember.
https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/RedirectBundle/Router/RedirectRouter.php#L171

This Redirect object is still managed by the EntityManager, so any flush happening after this will change the Redirect values in the db/storage.

This only happens when using an origin with a '*' wildcard.

Also a refactor to use the router to match a route instead of 'creating' a matching route before handling it to the router.

This also makes it possible to add wildcards as '/path*' instead of '/path/*'.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


jockri:
Hi @jverdeyen, can you change the destination branch to 5.0? All bugfixes are merged to this branch and then upmerged to the other branches. Thx!

jverdeyen:
@jockri Will do!

Maybe this could be a BC in some situations. As a `*` will be a wildcard in a route.
Redirects as `/nl/test*` will work now.

jockri:
I don't think that routes with * are valid url's, so the wildcard is more a new/added feature. It would be even better to split the PR into a bugfix (5.0) and the wildcard feature and refactoring (master).

jverdeyen:
I'll create a new PR for a bugfix into 5.0, so this can be merged into master (when reviewed) @jockri

acrobat:
@jverdeyen is this the feature PR (so it should target master) and is the bugfix PR not yet created?

jverdeyen:
Bugfix PR isn't created yet. The most important part is this:
https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/2216/files#diff-946dc7730fa3b4cf86e3e42778f5fd6eL178

It changes a Redirect entity object from the `em`. When an exception occurs, an `em->flush` is triggered and the Redirect entity object which is changed is also saved.

acrobat:
@jverdeyen  Thanks, I will also try to take a look at it. I will remove the milestone for now, to avoid confusion!

Issue #2211, opened 2019-01-09
[TranslatorBundle] Optimize importer


Numkil:
Translations importer method "importSingleTranslation is horrible slow".
Requires some serious rethinking of the whole translations flow to speed this up.

PR #2264, opened 2019-01-23
[TranslatorBundle] Speed up bulk import of large translation files

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |  #2211 

Old setup worked in this way.
for each translation
     fetch translation from database
     create/update translation
     flush
endfor

new setup will fetch entire translations database in memory and do the logic internally. Memory usage is limited because the translation object is super small. avg 15mb per 5500 records.

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
@@ -55,14 +55,14 @@ services:
 
     kunstmaan_translator.service.importer.importer:
         class: 'Kunstmaan\TranslatorBundle\Service\Command\Importer\Importer'
-        calls:
-            - [setTranslationGroupManager, ['@kunstmaan_translator.service.group_manager']] # default, maken via config var
+        arguments:
+            - '@kunstmaan_translator.service.group_manager'
         public: true
 
     kunstmaan_translator.service.group_manager:
         class: 'Kunstmaan\TranslatorBundle\Service\TranslationGroupManager'
-        calls:
-            - [setTranslationRepository, ['@kunstmaan_translator.repository.translation']]
+        arguments:
+            - '@kunstmaan_translator.repository.translation'
 
     # TODO: needs to be refactored, should not need to be public
     kunstmaan_translator.service.translator.loader:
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
```diff
@@ -52,14 +52,14 @@ services:
 
     kunstmaan_translator.service.importer.importer:
         class: 'Kunstmaan\TranslatorBundle\Service\Command\Importer\Importer'
-        calls:
-            - [setTranslationGroupManager, ['@kunstmaan_translator.service.group_manager']] # default, maken via config var
+        arguments:
+            $translationGroupManager: '@kunstmaan_translator.service.group_manager'
```
acrobat: We don't use autowiring in the bundles, so I would suggest to remove the parameter name here as this might be forgotten when renaming the parameter and the service creation will fail

src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
```diff
@@ -52,14 +52,14 @@ services:
...
         public: true
 
     kunstmaan_translator.service.group_manager:
         class: 'Kunstmaan\TranslatorBundle\Service\TranslationGroupManager'
-        calls:
-            - [setTranslationRepository, ['@kunstmaan_translator.repository.translation']]
+        arguments:
+            $translationRepository: '@kunstmaan_translator.repository.translation'
```
acrobat: We don't use autowiring in the bundles, so I would suggest to remove the parameter name here as this might be forgotten when renaming the parameter and the service creation will fail

src/Kunstmaan/TranslatorBundle/Service/Command/Importer/Importer.php
```diff
@@ -23,9 +22,21 @@ class Importer
      */
     private $translationGroupManager;
 
+    /**
+     * Importer constructor.
+     *
+     * @param TranslationGroupManager $translationGroupManager
```
acrobat: Doc block can be removed as it doesn't add any value

src/Kunstmaan/TranslatorBundle/Service/TranslationGroupManager.php
```diff
@@ -4,59 +4,38 @@
...
-        return $translationGroup;
-    }
+    /** @var int */
+    private $maxId = 1;
 
     /**
-     * Create new TranslationGroup instance (with the given locales of any are set)
-     *
-     * @param array $locales
+     * TranslationGroupManager constructor.
```
acrobat: Doc block can be removed as it doesn't add any value

acrobat:
@Numkil Can you also check the failing tests in travis and the codestyle issues in styleci? Thanks!

Numkil:
@acrobat fixed all the code.

Issue #2282, opened 2019-01-29
[Admin Bundle] Sub Entity arrows don't work


jordanmoon:
Look similar to this issue #2157.

As of v5 Page Parts can be reordered using the arrows, but Sub Entities cannot. It looks like an issue with the data attribute `draggable` being at an incorrect level.

Devolicious:
Hi @jordanmoon & @iknowfoobar.

On version 5.0.13 we do not have this bug as described. Which version are you guys using?

jordanmoon:
Hi @Devolicious 
This is in 5.1.1

Devolicious:
@jordanmoon Thx for the quick response. We will investigate this. TBC

acrobat:
The change in PR #2049 breaks this behaviour. I will create a pr to revert the change and then we can discuss further changes there to fix the issue pr #2049 fixed before.

Issue #2157, opened 2018-10-09
[AdminBundle] Ordering of page part sub entities has stopped working in v5


iknowfoobar:
We have upgraded several of our client sites over the last few months from v4 to v5. A couple of them use custom page parts that have sub entities, and we followed this news article to build them: https://bundles.kunstmaan.be/news/using-sub-entities-in-pageparts

However since upgrading to v5 the Up/Down arrows for the sub entities in the Page Part are no longer working. There is no error either it just appears that the click event no longer exists. Dragging and dropping the sub entities still works.

acrobat:
@iknowfoobar is it possible that it's related to the changes in PR #2049?

iknowfoobar:
@acrobat Thanks, it appears that it is.

iknowfoobar:
Hi @Devolicious, has this been resolved in an unreleased version? The conversation on the related pull request wasn't resolved as far as I can see. Thanks

Devolicious:
@iknowfoobar conversation will be continued in #2282

PR #2283, opened 2019-01-29
[AdminBundle] Revert PR #2049 changes

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2282, #2157

The change in PR #2049 broke the order functionality of sub entities of a pagepart. This PR reverts the change.

@SpadXIII what was the actual issue you fixed in your PR? So we can try to find a solution for both issues!

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/views/Form/fields.html.twig
+++ b/src/Kunstmaan/AdminBundle/Resources/views/Form/fields.html.twig
@@ -183,7 +183,7 @@
 
                 {# Items #}
                 {% for obj in form %}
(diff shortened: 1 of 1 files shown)


SpadXIII:
@acrobat the issue was having a collection type inside a page part; both the page parts containers and the collection type containers use the same id. So removing the second collection type item marks the second page part for removal.

The 'fix' removed the id of the inner items so it won't trigger a removal of a page part item anymore.

Devolicious:
@SpadXIII can you give this fix a go and see if it solves your problem? We believe that this will solve the issue so we're gonna merge but I would like to get your feedback aswell before I close #2282

Thx!

PR #2313, opened 2019-01-31
[SeoBundle] Added option to provide caching for og image dimension calls

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This pr add the logic to cache seo og:image calls by default. The seo twig extension will try to load the `og:image` to provide the `og:image:width` and `og:image:height` tags. This call will be performed on every request, so by default we will cache the result of this call to avoid doing unnecessary calls. A quick win of +-100ms 🎉


acrobat:
@deZinc good catch, fixed!

PR #2341, opened 2019-02-05
[TranslatorBundle] Deprecated the MigrationsDiffCommand & DiffCommand

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
@@ -109,6 +109,7 @@ services:
 
     kunstmaan_translator.service.command.diff:
         class: 'Kunstmaan\TranslatorBundle\Service\Command\DiffCommand'
+        deprecated: 'The "%service_id%" service is deprecated since KunstmaanTranslatorBundle 5.2 and will be removed in 6.0.'
 
     kunstmaan_translator.datacollector:
         class: Kunstmaan\TranslatorBundle\Toolbar\DataCollectorTranslator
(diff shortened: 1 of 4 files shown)


src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
```diff
@@ -107,6 +107,7 @@ services:
             - [setEntityManager, ['@doctrine.orm.default_entity_manager']]
         public: true
 
+    # todo: deprecated in 5.2, remove in 6.0
     kunstmaan_translator.service.command.diff:
         class: 'Kunstmaan\TranslatorBundle\Service\Command\DiffCommand'
```
acrobat: You can deprecate the service also

```yaml
deprecated: ~
```

This will produce a deprecation message like this: `The "%service_id%" service is deprecated. You should stop using it,
as it will soon be removed.`

deZinc: Right you are! Thanks for the tip.

src/Kunstmaan/TranslatorBundle/Resources/config/services.yml
```diff
@@ -109,6 +109,7 @@ services:
 
     kunstmaan_translator.service.command.diff:
         class: 'Kunstmaan\TranslatorBundle\Service\Command\DiffCommand'
+        deprecated: 'The "%service_id%" service is deprecated since TranslatorBundle 5.2 and will be removed in TranslatorBundle 6.0.'
```
acrobat: Mention "KunstmaanTranslatorBundle" instead of "TranslatorBundle" because these deprecations will show up for the whole application, this way the user knows which bundle causes the deprecation

deZinc: True, will change the name.

Issue #1325, opened 2016-10-10
Change Break-point for navigation in Admin Area


alexborton:
Is there a way to change the break-point for the admin area navigation? It's the first time i have come across the issue, but with a very large admin navigation with many extra module areas, the nav breaks onto two lines long before it flips into the "mobile" style of navigation.

As a per-project fix, it would be great to change the width value at which it changes. As a global, permanent solution, the navigation could be measured for fit and the value for the breakpoint become dynamic. 

Thanks

PR #2375, opened 2019-02-19
[AdminBundle] dynamic breakpoint for top navbar in admin area

dbeerten:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1325

Screensize where it fits on one line:
screen shot 2019-02-19 at 11 39 27

Smaller screensize where it didn't fit on one line => show mobile version
screen shot 2019-02-19 at 11 39 14


acrobat:
@dbeerten Is it also possible to let the menu render as normal, but when it would break on multiple lines then hide the "extra" items in an hamburg menu. Instead of hiding the whole menu?

dbeerten:
> @dbeerten Is it also possible to let the menu render as normal, but when it would break on multiple lines then hide the "extra" items in an hamburg menu. Instead of hiding the whole menu?

I created a mockup to test it:
image

Imho, it looks a bit odd that "menu item 2" and "menu item 3" are inside the hamburger now, while the rest is not... I don't know if users will be aware of this.

Numkil:
I agree with @dbeerten that it is better to show all or nothing. The mockup would confuse me.

Issue #1617, opened 2017-10-10
Move scripts to the head and use async / defer


Windvis:
Async / defer attributes on script elements are widely supported now (do we even support < IE11?). We can replace the oldschool 'load scripts at the bottom of the body' style of script loading by adding the scripts to the head and using the appropriate attribute.

More info: 
https://www.html5rocks.com/en/tutorials/speed/script-loading
http://www.growingwiththeweb.com/2014/02/async-vs-defer-attributes.html

sandergo90:
@dbeerten do we still need have work on this?

PR #2378, opened 2019-02-20
[AdminBundle] Load scripts in head with defer

dbeerten:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #1617 

**BC or not?**

We moved our scripts from the end of the body to the head with a `defer` attribute.

If you added custom javascript (for example in `_js_footer_extra.html.twig`) which depends on jQuery, your custom code might be broken because jQuery isn't available yet. If that's the case, you need to make sure to load your scripts with a `defer` attribute:
```
<script defer src="custom-jquery-script.js"></script>
```
or wait for the `DOMContentLoaded`-event for inline scripts:

```
<script>
    document.addEventListener('DOMContentLoaded', () => {
        console.log('$ and jQuery are available here.')
    });
</script>
```
or don't use jQuery :-)

FYI: https://alligator.io/html/defer-async/#async-vs-defer

### DIFF
--- a/UPGRADE-5.2.md
+++ b/UPGRADE-5.2.md
@@ -21,6 +21,7 @@ AdminBundle
  * Setting the `requiredlocales` directly is deprecated and support for this parameter will be removed in 6.0. Provide the `kunstmaan_admin.required_locales` config instead.
  * Setting the `defaultlocale` directly is deprecated and support for this parameter will be removed in 6.0. Provide the `kunstmaan_admin.default_locale` config instead.
  * Setting the `websitetitle` directly is deprecated and support for this parameter will be removed in 6.0. Provide the `kunstmaan_admin.website_title` config instead.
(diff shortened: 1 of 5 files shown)


ProfessorKuma (changes requested):
Hi @dbeerten, your PR needs some changes
* Your answer if this PR is BC breaking seems to be incorrect.

acrobat:
I would move this to at least the next minor version (or even next major?) and definitely provide a warning/explanation in the upgrade doc.

Numkil:
This is indeed a bc break. The code itself is OK so if you rebase to master then we can merge this in :)

ProfessorKuma (changes requested):
Hi @dbeerten, your PR needs some changes
* It seems that you should have submitted to the next major branch.
* This PR seems to need a milestone of a major release.

Devolicious (changes requested):
Temporarily blocking this to be merged. We need to see if we are going to add this in V5 or V6.

ProfessorKuma (approved):
Hi @dbeerten, your PR passed all our requirements.

Thank you for contributing!

PR #2381, opened 2019-02-21
[AdminBundle] Time choice widget inline

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When using the 'choice' widget for a TimeType in an admin form, every choice takes up a whole row.
I've added examples to clarify what I'm talking about.

**Old:**

old

**New:**

new


acrobat:
@JZuidema Thanks for this contribution! @dbeerten improved your fix in #2389 so the whole datetime widget is correctly styled. So I'm closing this PR in favor of #2389.

PR #2235, opened 2019-01-15
[GeneratorBundle] refactored page generator

deZinc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

PR #2382, opened 2019-02-23
[GeneratorBundle] Fixed Page(part) generating

jeroenjans:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This PR fixes some problems when generating a Page(part).

The error that was reported was:

```
In ExtraFunctions.php line 26:
                                                                                                    
  Impossible to invoke a method ("getName") on a string variable ("TestWebsiteBundle").  
```
And
```
In ExtraFunctions.php line 8:
                                                                                                                                                                                                           
  An exception has been thrown during the rendering of a template ("Catchable Fatal Error: Object of class Test\WebsiteBundle\TestWebsiteBundle could not be converted to string").  
                                                                                                                                                                                                           

In Environment.php(378) : eval()'d code line 33:
                                                                                                                                      
  Catchable Fatal Error: Object of class Test\WebsiteBundle\TestWebsiteBundle could not be converted to string  
```

Another possible solution would be to modify the `Entity/Pages/ExtraFunctions.php` file so that it's using `{{ bundle }}` instead of `{{ bundle.getName() }}`.
If this is the desired fix then let me know and I'll change the PR accordingly.

### DIFF
--- a/src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/page/Entity/Pages/ExtraFunctions.php
+++ b/src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/page/Entity/Pages/ExtraFunctions.php
@@ -23,7 +23,7 @@
     {
         return [
 {% for section in sections %}
-            '{% if not isV4 %}{{ bundle.getName() }}:{%endif%}{{ section }}',
+            '{% if not isV4 %}{{ bundle }}:{%endif%}{{ section }}',
 {% endfor %}
         ];
     }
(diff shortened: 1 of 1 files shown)


acrobat:
Hi @jeroenjans,

Thanks for this bugfix! This bug was caused by changes in PR [#2235](https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/2235/files#diff-836881d3f9d1d12dcea92c2d4684fcad). There was a incorrect change for sf4 made. The correct fix would to change the check in the `ExtraFunction.php` from 

```
{% if not isV4 %}{{ bundle.getName() }}:{%endif%}
```
To

```
{% if not isV4 %}{{ bundle }}:{%endif%}
```

Can you update your pr to revert to current fix and implement the suggested fix in `ExtraFunctions.php`? Thanks!

jeroenjans:
Hey @acrobat,

I've updated the PR according to the feedback.
Let me know if I need to change anything else!

acrobat:
Thanks @jeroenjans!

Issue #2390, opened 2019-02-27
[FormBundle] Make sending an email optional with optional eventlistener


devigner:
In `src/Kunstmaan/FormBundle/Helper/FormHandler.php` the `handleForm` method directly uses the `kunstmaan_form.form_mailer` service. The only way to disable sending an email is not setting 'form', 'to' and 'subject' in the custom Form Entity. 

Using the `kunstmaan_form.add_submission` event makes disabling sending much easier and more graceful.

PR #2391, opened 2019-02-27
[FormBundle] Moved FormMailer dependency to separate listener

devigner:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2390

Moved sending the email to a listener. The listener can be removed when no email needs to be send after form submission. For example when the form submission is used to create a MailChimp user.

### DIFF
--- a/src/Kunstmaan/FormBundle/Event/SubmissionEvent.php
+++ b/src/Kunstmaan/FormBundle/Event/SubmissionEvent.php
@@ -21,7 +21,7 @@ class SubmissionEvent extends Event
 
     /**
      * @param FormSubmission $submission
-     * @param AbstractPage   $page
+     * @param FormPageInterface $page
      */
     public function __construct(FormSubmission $submission, FormPageInterface $page)
     {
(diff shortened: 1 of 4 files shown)


devigner:
The 'SymfonyInsight: kunstmaanlabs / KunstmaanBundlesCMS' failure is caused by code outside my PR.

src/Kunstmaan/FormBundle/Event/SubmissionEvent.php
```diff
@@ -3,6 +3,7 @@
 namespace Kunstmaan\FormBundle\Event;
 
 use Kunstmaan\FormBundle\Entity\FormSubmission;
+use Kunstmaan\FormBundle\Helper\FormPageInterface;
 use Kunstmaan\NodeBundle\Entity\AbstractPage;
```
Devolicious: Can you switch out all of the AbstractPage typehints because this is an unused use statement now.

PR #2395, opened 2019-03-08
[NodeSearchBundle] fix #2380 childs of structurenodes never get indexed and tests

### LINKED ISSUE #2380: [NodeSearchBundle] hasOfflineParents checks NodeTranslation for StructureNode
https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php#L132

```php
    /**
     * @param $nodeTranslation
     *
     * @return bool
     */
    private function hasOfflineParents(NodeTranslation $nodeTranslation)
    {
        $lang = $nodeTranslation->getLang();
        foreach ($nodeTranslation->getNode()->getParents() as $node) {
            $nodeNT = $node->getNodeTranslation($lang, true);
            if ($nodeNT && !$nodeNT->isOnline() && !$nodeNT instanceof StructureNode) {
                return true;
            }
        }

        return false;
    }
```

In the snippet above `$nodeNT instanceof StructureNode` will always be false.
So every Node under a `StructureNode` will not be re-indexed when saving. The function `hasOfflineParents` will return true in this case.

The `instanceof StructureNode` compare should be done on a Page entity.

This will only happen when saving a node in the Admin interface. The populate command uses another method for indexing.

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2380 

bugfix without bc break for above ticket.

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php
+++ b/src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php
@@ -2,7 +2,9 @@
 
 namespace Kunstmaan\NodeSearchBundle\EventListener;
 
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Event\LifecycleEventArgs;
+use Kunstmaan\NodeBundle\Entity\Node;
 use Kunstmaan\NodeBundle\Entity\NodeTranslation;
 use Kunstmaan\NodeBundle\Entity\StructureNode;
 use Kunstmaan\NodeBundle\Event\NodeEvent;
@@ -18,16 +20,18 @@ class NodeIndexUpdateEventListener implements NodeIndexUpdateEventListenerInterf
     /** @var ContainerInterface */
     private $container;
 
+    /** @var EntityManagerInterface */
+    private $em;
+
     /** @var NodePagesConfiguration */
     private $nodePagesConfiguration;
 
     /** @var array */
     private $entityChangeSet;
 
-    /**
-     * @param ContainerInterface $container
-     */
-    public function __construct(/* NodePagesConfiguration */ $nodePagesConfiguration)
+    public function __construct(/* NodePagesConfiguration */
+        $nodePagesConfiguration, /* EntityManagerInterface */
+        $em = null)
     {
         if ($nodePagesConfiguration instanceof ContainerInterface) {
             @trigger_error(sprintf('Passing the container as the first argument of "%s" is deprecated in KunstmaanNodeSearchBundle 5.2 and will be removed in KunstmaanNodeSearchBundle 6.0. Inject the "%s" service instead.', __CLASS__, 'kunstmaan_node_search.search_configuration.node'), E_USER_DEPRECATED);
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php
```diff
@@ -18,17 +20,23 @@ class NodeIndexUpdateEventListener implements NodeIndexUpdateEventListenerInterf
...
 
-    /**
-     * @param ContainerInterface $container
-     */
-    public function __construct(/* NodePagesConfiguration */ $nodePagesConfiguration)
+    public function __construct(/** NodePagesConfiguration */
+        $nodePagesConfiguration, /** EntityManagerInterface */
+        $em = null)
     {
+        if (null !== $em) {
```
acrobat: The null check is not needed. We can just assign the value as we check the property in the method below.

src/Kunstmaan/NodeSearchBundle/Resources/config/update_listener.yml
```diff
@@ -4,7 +4,7 @@ parameters:
 services:
     kunstmaan_node_search.node_index_update.listener:
         class: '%kunstmaan_node_search.node_index_update.listener.class%'
-        arguments: ['@kunstmaan_node_search.search_configuration.node']
+        arguments: ['@kunstmaan_node_search.search_configuration.node', '@doctrine.orm.entity_manager.abstract']
```
acrobat: Inject `@doctrine.orm.entity_manager` or `doctrine.orm.default_entity_manager` instead, as these link to the default orm connection.

src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php
```diff
@@ -18,17 +20,23 @@ class NodeIndexUpdateEventListener implements NodeIndexUpdateEventListenerInterf
...
 
     /** @var array */
     private $entityChangeSet;
 
-    /**
-     * @param ContainerInterface $container
-     */
-    public function __construct(/* NodePagesConfiguration */ $nodePagesConfiguration)
+    public function __construct(/** NodePagesConfiguration */
+        $nodePagesConfiguration, /** EntityManagerInterface */
```
acrobat: `EntityManagerInterface` typehint can be set as normal

src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php
```diff
@@ -18,17 +20,23 @@ class NodeIndexUpdateEventListener implements NodeIndexUpdateEventListenerInterf
...
+        $em = null)
     {
+        if (null !== $em) {
+            $this->em = $em;
+        }
+
```
acrobat: Also add a deprecation notice if the entity manager is not passed, so users know this argument will be required in 6.0

src/Kunstmaan/NodeSearchBundle/EventListener/NodeIndexUpdateEventListener.php
```diff
@@ -127,11 +135,28 @@ public function delete(NodeEvent $event)
...
-            if ($nodeNT && !$nodeNT->isOnline() && !$nodeNT instanceof StructureNode) {
-                return true;
+        $node = $nodeTranslation->getNode();
+        if (null !== $this->em) {
+            $em = $this->em;
+        } elseif (null !== $this->container) {
+            $em = $this->container->get('doctrine.orm.entity_manager.abstract');
+        } else {
+            /** Revert to bugged behaviour to avoid BC breaking */
+            return false;
```
acrobat: This will return a different value as it would be the case in the previous (buggy) behavior? Or am I wrong?

acrobat:
@Numkil can you also fix the failed styleci check

Numkil:
@Devolicious i adjusted the code

PR #2400, opened 2019-03-11
WIP: Refactor structure node name

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

Merge after #2395 when ready.


Devolicious:
This PR has BC breaks because the `HasNodeInterface` is being changed. This must be done in a way it is without BC breaks.

Devolicious:
It also has deprecation so that means that it should at least be submitted to master

PR #2408, opened 2019-03-22
[AdminBundle] Log console errors as critical instead of error

Devolicious:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Because our standard sentry config has error logging on `critical`, console errors never showed up.


acrobat:
@Devolicious maybe we should just add a suggest for the sentry/sentry-symfony (and add it to our skeleton) because that bundle has already all listeners and setup done to log all errors to sentry.

https://github.com/getsentry/sentry-symfony/blob/master/src/EventListener/ErrorListener.php

Don't know if we want this but just dropping this here!

PR #2421, opened 2019-04-08
[AllBundles] Replace deprecated getRootDir calls

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


Devolicious:
@acrobat, SymfonyInsight has a good point, maybe you can also deprecate the versionchecker container injection?

acrobat:
@Devolicious yes indeed I also had the idea to do this, but I would suggest doing it in a separate pr (as it goes beyond the scope of this pr) to deprecate the container and do proper injection of all depedencies/parameters.

PR #2422, opened 2019-04-08
[FormBundle] In newer versions of symfony the @templating service is replaced by the delegatingengine by default

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/FormBundle/Helper/FormMailer.php
+++ b/src/Kunstmaan/FormBundle/Helper/FormMailer.php
@@ -5,8 +5,8 @@ namespace Kunstmaan\FormBundle\Helper;
 use Kunstmaan\FormBundle\Entity\FormSubmission;
 use Swift_Mailer;
 use Swift_Message;
-use Symfony\Bundle\TwigBundle\TwigEngine;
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/FormBundle/Helper/FormMailer.php
```diff
@@ -16,18 +16,18 @@ class FormMailer implements FormMailerInterface
...
     private $container;
 
     /**
      * @param Swift_Mailer       $mailer     The mailer service
-     * @param TwigEngine         $templating The templating service
+     * @param DelegatingEngine   $templating The templating service
      * @param ContainerInterface $container  The container
      */
-    public function __construct(Swift_Mailer $mailer, TwigEngine $templating, ContainerInterface $container)
+    public function __construct(Swift_Mailer $mailer, DelegatingEngine $templating, ContainerInterface $container)
```
acrobat: I think it's best to typehint against `Symfony\Component\Templating\EngineInterface` so we don't lock to that deligatingEnigine class

PR #2435, opened 2019-04-26
[FormBundle] EmailValidator does not work correctly with emailformsubmissionfield

Numkil:
… correctly

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | c


Devolicious:
This does not work. We need to change that the constraints are passed through via the `constraints` options as that will set the constraints also on the parent object.

PR #2443, opened 2019-05-14
[NodeBundle] Added missing translation keys

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

These keys are used in the nodeAdmin template but were not translated

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/a7fff6a4045d20deb0b6dc832176c7ddb2254b38/src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/edit.html.twig#L28

Now they are available to be translated in crowdin


acrobat:
@Devolicious Fixed!

PR #2463, opened 2019-05-21
[NodeBundle] Throw a NotFoundException when the page entity is not found

SpadXIII:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Throw a NotFoundException when the page entity is not found in SlugController.
We have a DoctrineFilter active that filters out page entities. So when the SlugController tries to find the node translation it works, but when it searches for the page entity, it gets a `null` returned. Here a NotFoundException should be thrown

### DIFF
--- a/src/Kunstmaan/NodeBundle/Controller/SlugController.php
+++ b/src/Kunstmaan/NodeBundle/Controller/SlugController.php
@@ -54,6 +54,11 @@ class SlugController extends Controller
             $em,
             $nodeTranslation
         );
+
+        if (!$entity) {
+            throw $this->createNotFoundException('No page entity found for slug ' . $url);
+        }
+
         $node = $nodeTranslation->getNode();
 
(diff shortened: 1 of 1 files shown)


acrobat:
@SpadXIII how can it be possible that node translation is found but not the related page? Each nodetranslation has a node version which has a page linked to it.

SpadXIII:
I don't have access to the code base anymore (switched jobs a little while ago), but from what I remember, we had a doctrine filter active which filtered out certain pages. Throwing an exception helps make the handling easier.

Without the doctrine filter, everything should work just fine.

acrobat:
I'm going to close this one for now, a nodetranslation should always have a related page to it. So this is more of an edge case, as you don't have access to the original code anymore I suggest to keep it like this for now and revisit if this comes up again.

PR #2490, opened 2019-05-22
[DashboardBundle] Code improvement found by static code analysis

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/DashboardBundle/Command/GoogleAnalyticsDataCollectCommand.php
+++ b/src/Kunstmaan/DashboardBundle/Command/GoogleAnalyticsDataCollectCommand.php
@@ -209,7 +209,7 @@ class GoogleAnalyticsDataCollectCommand extends ContainerAwareCommand
         }
 
         // create default overviews for this config if none exist yet
-        if (!count($config->getOverviews())) {
+        if (!\count($config->getOverviews())) {
             $overviewRepository->addOverviews($config);
         }
 
@@ -237,7 +237,7 @@ class GoogleAnalyticsDataCollectCommand extends ContainerAwareCommand
 
         foreach ($configs as $config) {
             // add overviews if none exist yet
-            if (count($config->getOverviews()) == 0) {
+            if (\count($config->getOverviews()) == 0) {
                 $overviewRepository->addOverviews($config);
             }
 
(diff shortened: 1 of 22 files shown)


src/Kunstmaan/DashboardBundle/Command/GoogleAnalyticsDataCollectCommand.php
```diff
@@ -186,7 +186,7 @@ private function getOverviewsOfSegment($segmentId)
         $segmentRepository->initSegment($segment);
 
         // get the overviews
-        return $segment->getOverviews();
+        return $segment->getoverviews();
```
Devolicious: I would rather see that the method itself was formatted to a correct camelCase way instead of changing the method call.

src/Kunstmaan/DashboardBundle/Command/GoogleAnalyticsOverviewsListCommand.php
```diff
@@ -124,7 +122,7 @@ private function getOverviewsOfSegment($segmentId)
         }
 
         // get the overviews
-        return $segment->getOverviews();
+        return $segment->getoverviews();
```
Devolicious: idem dito

src/Kunstmaan/DashboardBundle/Controller/GoogleAnalyticsController.php
```diff
@@ -61,7 +61,7 @@ public function widgetAction(Request $request)
         // set the overviews param
         $params['token'] = true;
         if ($segmentId) {
-            $overviews = $em->getRepository('KunstmaanDashboardBundle:AnalyticsSegment')->find($segmentId)->getOverviews();
+            $overviews = $em->getRepository('KunstmaanDashboardBundle:AnalyticsSegment')->find($segmentId)->getoverviews();
```
Devolicious: idem dito

src/Kunstmaan/DashboardBundle/Command/GoogleAnalyticsDataCollectCommand.php
acrobat: Good point! I though it would be a bc break to rename the method, but methods are case insensitive, so rename the method and restored these changes!

PR #2492, opened 2019-05-29
[FixturesBundle] Mark services as public

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

These services are accessed directly from the container so should be public for sf4.


acrobat:
@devigner Can you check this patch if this fixes the symfony 4 compatibility issues? Thanks!

devigner:
@acrobat That is exactly what I did to get it working! Thanks.

Issue #2499, opened 2019-06-05
[FixtureBundle] Default User repository is used hardcoded but needs to be customizable


devigner:
Because overriding the User Entity is possible (see: https://bundles.kunstmaan.be/news/extending-the-admin-user-entity). I must be possible to override the default User Repository inside the PageBuilder.

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/FixturesBundle/Builder/PageBuilder.php#L50

PR #2500, opened 2019-06-05
[FixtureBundle] Inject the User Entity class name so the correct Repository can be used

devigner:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2499

### DIFF
--- a/src/Kunstmaan/FixturesBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/FixturesBundle/Resources/config/services.yml
@@ -1,3 +1,6 @@
+parameters:
+    kunstmaan_fixtures.user.class: Kunstmaan\AdminBundle\Entity\User
+
 services:
     kunstmaan_fixtures.builder.builder:
         class: Kunstmaan\FixturesBundle\Builder\BuildingSupervisor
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/FixturesBundle/Resources/config/services.yml
```diff
@@ -13,6 +16,7 @@ services:
             - '@kunstmaan_fixtures.populator.populator'
             - '@kunstmaan_utilities.slugifier'
             - '@kunstmaan_node.pages_configuration'
+            - '%kunstmaan_fixtures.user.class%'
```
Devolicious: I would not create a new parameter but use the `fos_user.model.user.class` parameter instead since that will be the one you need to configure when you want to change the default `User` object.

src/Kunstmaan/FixturesBundle/Resources/config/services.yml
```diff
@@ -1,3 +1,6 @@
+parameters:
```
Devolicious: This is not necessary if you use the `fos_user.model.user.class`

Devolicious:
@devigner can you rebase your PR from branch 5.3 ? I changed the base branch to 5.3 as it is a bugfix but now it contains commits from master :)

PR #2515, opened 2019-07-05
[AdminBundle] Fix phpunit test warnings

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Symfony deprecated and removed a method on the `RoleHierarchyInterface`. After that change our unittest triggers warning because we can't mock non-existing methods. I think technically this is a bc break from their side, but this will only break in tests. If we move our mock to the actual class everything works correctly and we get the correct deprecation warnings.

See: https://github.com/symfony/symfony/pull/30388


acrobat:
Closing this as a better fix is proposed in #2539

PR #2518, opened 2019-07-08
[NodeBundle] Updated SlugType prefix build

jverdeyen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    |  no
| Deprecations? | no
| Fixed tickets |

Updated SlugType prefix build to avoid double slashes when using StructurePage.

Using the same method as in https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/5.0/src/Kunstmaan/NodeBundle/Entity/NodeTranslation.php#L255

This bug appears on the 'Menu' tab of a page, the prefix to render a preview url (from js) should not contain double slashes. A double slash is possible when the parent page is a structure page (without slug).


acrobat:
@jverdeyen Can you rebase the PR against the 5.3 branch, this branch is used for bugfix releases. Thanks!

acrobat:
Thanks @jverdeyen!

jverdeyen:
Thanks! 🥳

PR #2523, opened 2019-07-19
[AdminBundle] collapse collection items

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

We ran into the problem that once you have a bunch of items in a collection type, while the item itself also has a lot of input fields, that you can't really find anything anymore. This merge requests automatically collapses items of a collection type if the item has more than 5 children.

For example:

Collection item has more than 5 children. Its automatically collapsed. There is a button at the right top to expand it.
expand 1

Button is clicked, and collection item is expanded to show all of its input fields
expand 2


src/Kunstmaan/AdminBundle/Resources/views/Form/fields.html.twig
```diff
@@ -195,10 +195,15 @@
                                 <i class="fa fa-chevron-down"></i>
                             </button>
                             {% endif %}
+                            {% if obj.children | length > 5 %}
```
acrobat: Would it make sense to make this number configurable?

JZuidema: I'm actually not sure how. I can add it to the configuration and then add it to parameters, but you can't inject a parameter into twig, unless its been made available globally.

JZuidema:
> 
> 
> @JZuidema I like the idea, maybe we should also add a expand button in the grey collapsed indicating zone. Small box in the middle of this zone with a caret down icon?
> 
> image
> 
> To make it more clear this zone can be expanded.

Okay well, I've looked into this but frontend is not really my area of expertise. Do you think one of you guys could look into it?

acrobat:
@dbeerten @FVKVN can anyone check the frontend comments and see if we can improve those?

ProfessorKuma (changes requested):

acrobat:
This still needs some FE work to improve the look of this feature. I'm removing the milestone so we finish this somewhere in 6.x

PR #2525, opened 2019-07-19
[AllBundles] Provide autowire aliases for frequently injected classes

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Some classes you might need to inject from time to time, so let's make it easier to autowire these classes by providing an alias

### DIFF
--- a/src/Kunstmaan/MediaBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/MediaBundle/Resources/config/services.yml
@@ -13,6 +13,9 @@ parameters:
     kunstmaan_media.validator.has_guessable_extension.class: 'Kunstmaan\MediaBundle\Validator\Constraints\HasGuessableExtensionValidator'
 
 services:
+    # Autowire aliases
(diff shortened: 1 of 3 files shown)


Numkil:
ik zou hier mischien nog een alias maken voor de volgende veelgebruikte services

DomainConfiguration -> adminbundle
MediaManager -> mediabundle
NodeSearcher -> nodesearchbundle
SearchService -> nodesearchbundle

acrobat:
I've added aliases for the `MediaManager`, `NodeSearcher` and `SearchService`. 

For the `DomainConfiguration` I need to dig in the bundles a bit more to add a check to alias the adminbundle class only if the multidomain bundle is not installed and used, otherwise we need to alias the multidomain class for the `DomainConfigurationInterface`. 
This way users can typehint against the interface and still receive the correct class. I will create a separate PR for that alias

PR #2529, opened 2019-07-19
[NodeBundle] Feature/current url menu tab

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

The url shown in the Menu tab when editing a page is now clickable and opens the current page in a new tab. Here is a screenshot of what it looks like:

image


Numkil:
I like this idea but I do not agree with removing the javascript preview of what the complete slug of the page will be when edititing. 
As the developer that feature might be less usefull but for less techincal people it is very nice to know what the url of my page will be before saving.

JZuidema:
@Numkil but it shows what the url is going to be right? The text field where you change the url is prefixed with the rest of the url. So it actually shows what the url is going to be. Of course we could keep the original javascript, but it would seem a bit unnecessary

Numkil:
@JZuidema I know its very minor but the little javascript shows what the parent slug combined with the new slug will look like. For developers its a non issue i realize but some less technical users probably enjoy seeing that before they actually press save.

JZuidema:
@Numkil I guess we can use the old javascript to display "Updated url" below "Current url" if the url is changed? Would that be acceptable?

Numkil:
Yeah I guess that could work. Thanks for considering it.

JZuidema:
@Numkil I've updated the code to do what we discussed:

Here is what it looks like if the slug has not been changed yet:
updated url1

Here's what it'll show once the slug has been changed
updated url2

The updated url will disappear once the slug is equal to the original slug (by pressing reset for example)

Numkil:
Thanks that looks great and works.

PR #2531, opened 2019-07-30
[NodeBundle] Fix incorrect step in documentation

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


acrobat:
This was already fixed in #2420, but I think the upmerge was not done from the 5.2 branch. I will upmerge this change into master!

acrobat:
I missed that this was about the docs in the bundle itself, the general docs (those used for readthedocs) were already fixed. Let's merge this now, but maybe we should check if the need to remove these bundle doc folders in the future.

PR #2560, opened 2019-09-30
[TranslatorBundle] fix importbutton translations

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

On symfony 3.4 this check is broken because the import button would ALWAYS only import the global translations and not even check for the bundles.

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/Service/Command/Importer/ImportCommandHandler.php
+++ b/src/Kunstmaan/TranslatorBundle/Service/Command/Importer/ImportCommandHandler.php
@@ -37,11 +37,22 @@ class ImportCommandHandler extends AbstractCommandHandler
      */
     public function executeImportCommand(ImportCommand $importCommand)
     {
-        if ($importCommand->getGlobals() || $importCommand->getDefaultBundle() === false || $importCommand->getDefaultBundle() === null) {
-            return $this->importGlobalTranslationFiles($importCommand);
+        $amount = 0;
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/TranslatorBundle/Service/Command/Importer/ImportCommandHandler.php
```diff
@@ -37,11 +37,22 @@ class ImportCommandHandler extends AbstractCommandHandler
...
+
+            $amount = $this->importGlobalTranslationFiles($importCommand);
+        } elseif ($defaultBundleNotSet) {
+            $importCommand->setDefaultBundle('all');
         }
 
-        return $this->importBundleTranslationFiles($importCommand);
+        $amount += $this->importBundleTranslationFiles($importCommand);
+
+        return $amount;
```
acrobat: Maybe the code can be restructed a bit to avoid the double `importGlobalTranslationFiles` function call and the elseif. 

```php
        if ($importCommand->getGlobals()) {
            $amount = $this->importGlobalTranslationFiles($importCommand);
            if ($defaultBundleNotSet) {
                return $amount;
            }
        } 
        
        if ($defaultBundleNotSet) {
            $importCommand->setDefaultBundle('all');
        }

        $amount += $this->importBundleTranslationFiles($importCommand);

        return $amount;
```

acrobat:
Follow up pr with the comments applied -> #2568

PR #2564, opened 2019-10-09
[AdminListBundle] Display associations when using viewAction()

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

`$MetaData->fieldNames` returns a list of regular property names, it does not return names of association mappings. This causes all associations not to be shown when viewing an entity in the adminlist.
`$metaData->getReflectionProperties()` returns all properties. Also added `key | trans` to be able to translate the names of the properties when viewing the data.

I'd like to add that this change could potentially cause an exception for anyone who has uses view option in an admin list, when their entity has assocations and those entities havent implemented __toString()

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
+++ b/src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
@@ -322,10 +322,10 @@ abstract class AdminListController extends Controller
             throw $this->createAccessDeniedException('You do not have sufficient rights to access this page.');
         }
 
-        $MetaData = $em->getClassMetadata($configurator->getRepositoryName());
-        $fields = array();
+        $metaData = $em->getClassMetadata($configurator->getRepositoryName());
+        $fields = [];
         $accessor = PropertyAccess::createPropertyAccessor();
-        foreach ($MetaData->fieldNames as $value) {
+        foreach ($metaData->getReflectionProperties() as $value => $reflectionProperty) {
             $fields[$value] = $accessor->getValue($helper, $value);
         }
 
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/AdminListBundle/Controller/AdminListController.php
```diff
@@ -322,10 +322,10 @@ protected function doViewAction(AbstractAdminListConfigurator $configurator, $en
...
         }
 
-        $MetaData = $em->getClassMetadata($configurator->getRepositoryName());
-        $fields = array();
+        $metaData = $em->getClassMetadata($configurator->getRepositoryName());
+        $fields = [];
         $accessor = PropertyAccess::createPropertyAccessor();
-        foreach ($MetaData->fieldNames as $value) {
+        foreach ($metaData->getReflectionProperties() as $value => $reflectionProperty) {
             $fields[$value] = $accessor->getValue($helper, $value);
```
acrobat: You could add a check here to see if the object has a `__toString` method, otherwise skip the property. Something like

```php
if (\is_object($value) && !method_exists($value, '__toString')) {
    continue;
}
```

JZuidema: That would not return the expected results. The value could be returned by `$accessor->getValue($helper, $value);` without having `__toString()` implemented in the entity. Adding your suggestion would cause no values to be displayed.

I've just tested this with an entity that has relations without having `__toString()` implemented. It caused an error. But the error came from the method `getStringValue()` from `AbstractAdminListConfigurator`, because it just tries to access `getName()` on every entity in the collection (eventhough that most likely does not exist).

I get that you cant add this if it breaks rendering objects  which don't have the __toString method. I think at this point it should only break for anyone who has overwritten the `getStringValue()` method, while not calling `parent::getStringValue()` and not taking in account the `$result` could be an instance of `PersistentCollection`. I don't know what the policy is that you guys have on this, but if this is not good enough then you can close this pull request :)

PR #2565, opened 2019-10-10
[MediaBundle] Fixed compatibility with new LiipImagine generateUrl() function

poveu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/MediaBundle/Helper/Imagine/CacheManager.php
+++ b/src/Kunstmaan/MediaBundle/Helper/Imagine/CacheManager.php
@@ -25,12 +25,12 @@ class CacheManager extends \Liip\ImagineBundle\Imagine\Cache\CacheManager
         }
 
         if (empty($runtimeConfig)) {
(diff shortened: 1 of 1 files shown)


acrobat:
@poveu Can you use the new parameter also on line 28 and 33? Otherwise the parameter is unused and therefor useless. Thanks!

poveu:
> @poveu Can you use the new parameter also on line 28 and 33? Otherwise the parameter is unused and therefor useless. Thanks!

Right, didn't notice that :) Done!

I don't really understand what new issue StyleCI is pointing at.

acrobat:
@poveu It's indicating that there should be an empty line between the variable and the return statement

```php
$info = pathinfo($path);

return $info['dirname'] . DIRECTORY_SEPARATOR . $info['filename'] . '.' . $format;
```

PR #2572, opened 2019-10-23
[NodeBundle] Fix undefined offset error caused by no owner being set for Template

molovo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

In the `onKernelView` method of [`Sensio\Bundle\FrameworkExtraBundle\EventListener\TemplateListener`](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/src/EventListener/TemplateListener.php#L87) there is a `list()` call to pull out the controller and action used by the template.

The owner for the template is normally set automatically within the `onKernelController` event. Because in the node bundle's `SlugController` the Template is being created manually, the owner for the template has not been set as `onKernelController` has already been called. This causes an error `Notice: Undefined offset: 0` when the TemplateListener attempts to pull out the owner.

This PR sets the template owner manually to `[SlugController::class, 'slugAction']` which prevents this error.

Similiarly, in `Kunstmaan\NodeBundle\EventListener\RenderContextListener` the Template is being created manually. Here we are able to pull the current controller and action from the request context and again set the template owner manually.

### DIFF
--- a/src/Kunstmaan/NodeBundle/EventListener/RenderContextListener.php
+++ b/src/Kunstmaan/NodeBundle/EventListener/RenderContextListener.php
@@ -95,8 +95,12 @@ class RenderContextListener
             //the SensioFrameworkExtraBundle kernel.view will handle everything else
             $event->setControllerResult((array) $parameters);
 
-            $template = new Template(array());
+            $template = new Template([]);
             $template->setTemplate($entity->getDefaultView());
+
+            $controllerBits = explode(':', $request->attributes->get('_controller'));
+            $action = array_pop($controllerBits);
+            $template->setOwner([join(':', $controllerBits), $action]);
 
             $request->attributes->set('_template', $template);
         }
--- a/src/Kunstmaan/NodeBundle/Controller/SlugController.php
+++ b/src/Kunstmaan/NodeBundle/Controller/SlugController.php
@@ -109,6 +109,7 @@ class SlugController extends Controller
 
         $template = new Template(array());
         $template->setTemplate($view);
+        $template->setOwner([SlugController::class, 'slugAction']);
 
         $request->attributes->set('_template', $template);
 


acrobat:
@molovo how can the current issue/bug be reproduced?

molovo:
@acrobat we are finding it on any website running Kunstmaan 5.3.5 - simply loading any page with `ENVIRONMENT=dev` in `.env.local` triggers the error

poveu:
@acrobat I don't exactly know what are the exact steps to reproduce, but the same issue happens in our projects (with newest Sensio and Kunstmaan bundle), so it indeed does need a fix.

acrobat:
I was able to reproduce this on a new installation. Part of this fix is ok, the change in the renderContextListener won't work because it only works for the old (and deprecate) `XXXBundle:Controller:action` notation. The new syntax `Namespace\For\Controller::method` doesn't work here

src/Kunstmaan/NodeBundle/EventListener/RenderContextListener.php
```diff
@@ -95,9 +95,13 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             //the SensioFrameworkExtraBundle kernel.view will handle everything else
             $event->setControllerResult((array) $parameters);
 
-            $template = new Template(array());
+            $template = new Template([]);
             $template->setTemplate($entity->getDefaultView());
 
+            $controllerBits = explode(':', $request->attributes->get('_controller'));
```
acrobat: This will only work for the old (and deprecated) `XXXBundle:Controller:action` notation. The new syntax `Namespace\For\Controller::method` doesn't work here

molovo: @acrobat should I update this to only support the new syntax, or do we need to maintain backwards compatibility? The old syntax is still in use quite a bit internally at the moment

acrobat: I would indeed add a BC layer to support both the old and new syntax. This way we easily can remove the old deprecated notation once we update the supported symfony versions and are we able to fix this bug for older installations

acrobat:
@molovo thanks for the work on this, I've create a PR to implement the BC check. See #2594

PR #2581, opened 2019-11-07
[MultiDomainBundle] Use Twig namespace syntax instead of bundle refer…

waaghals:
We've overwritten the view for this bundle using bundle inheritance.
But bundle inheritance is deprecated. When removing the bundle inheritance it is no longer possible for us to override the views for this bundle.

When using the newer twig syntax using namespaces it becomes possible to override the view again by setting the namespace to a different view folder. (By setting `twig.paths` configuration.)

This PR changes the twig paths to the namespace syntax, allowing the view to be overwritten again.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

CC @dannyvw


acrobat:
I'm closing this one in favor of #2553. This PR replaces all twig paths to the correct syntax.

PR #2589, opened 2019-11-19
[MenuBundle] added class for menu item with active subnode

poveu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

MenuBundle has an option to set HTML class for element that's active, but when you use multilevel-menu it's good to know (for CSS styling) if any parent-level menu item has active child.


acrobat:
Thanks @poveu!

PR #2603, opened 2019-12-12
[TranslatorBundle] use the common export logic instead of a custom export

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

[TranslatorBundle] use the common export logic instead of a custom export but only from the settings view and not from commandline.

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
+++ b/src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
@@ -6,6 +6,8 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Kunstmaan\AdminListBundle\AdminList\Configurator\AbstractDoctrineDBALAdminListConfigurator;
 use Kunstmaan\AdminListBundle\AdminList\Configurator\ChangeableLimitInterface;
+use Kunstmaan\AdminListBundle\AdminList\Field;
+use Kunstmaan\AdminListBundle\AdminList\FieldAlias;
 use Kunstmaan\AdminListBundle\AdminList\FilterType\DBAL\EnumerationFilterType;
 use Kunstmaan\AdminListBundle\AdminList\FilterType\DBAL\StringFilterType;
 use Kunstmaan\AdminListBundle\Traits\ChangeableLimitTrait;
@@ -27,6 +29,11 @@ class TranslationAdminListConfigurator extends AbstractDoctrineDBALAdminListConf
      * @var string
      */
     protected $locale;
+
+    /**
+     * @var Field[]
+     */
+    private $exportFields = [];
 
     /**
      * @param Connection $connection
@@ -64,6 +71,31 @@ class TranslationAdminListConfigurator extends AbstractDoctrineDBALAdminListConf
         $this->addField('status', 'kuma_translator.adminlist.header.status', true);
     }
 
+    public function getExportFields()
+    {
+        if (empty($this->exportFields)) {
+            $this->addExportField('domain', 'kuma_translator.adminlist.header.domain');
+            $this->addExportField('keyword', 'kuma_translator.adminlist.header.keyword');
+
+            $this->locales = array_unique($this->locales);
+            // Field building hack...
+            foreach ($this->locales as $locale) {
+                $this->addExportField($locale, strtoupper($locale));
+            }
+
+            $this->addExportField('status', 'kuma_translator.adminlist.header.status');
+        }
+
+        return $this->exportFields;
+    }
+
+    public function addExportField($name, $header, $template = null, FieldAlias $alias = null)
+    {
+        $this->exportFields[] = new Field($name, $header);
+
+        return $this;
+    }
+
     /**
      * @return bool
      */
@@ -88,6 +120,14 @@ class TranslationAdminListConfigurator extends AbstractDoctrineDBALAdminListConf
      * @return bool
      */
     public function canEditInline($item)
+    {
+        return true;
+    }
+
+    /**
+     * @return bool
+     */
+    public function canExport()
     {
         return true;
     }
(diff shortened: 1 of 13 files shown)


src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
```diff
@@ -28,6 +30,11 @@ class TranslationAdminListConfigurator extends AbstractDoctrineDBALAdminListConf
      */
     protected $locale;
 
+    /**
+     * @var Field[]
+     */
+    private $exportFields = array();
```
acrobat: Change this to the short array syntax

src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
```diff
@@ -64,6 +71,30 @@ public function buildFields()
...
+                $this->addExportField($locale, strtoupper($locale));
+            }
+        }
+
+        return $this->exportFields;
+    }
+
+    public function addExportField($name, $header)
+    {
+        $this->exportFields[] = new Field($name, $header, false, null, null);
```
acrobat: The `false, null, null` arguments can be removed as these are the default values

src/Kunstmaan/TranslatorBundle/AdminList/TranslationAdminListConfigurator.php
```diff
@@ -92,6 +123,11 @@ public function canEditInline($item)
         return true;
     }
```
acrobat:
```
/**
 * @return bool
 */
```

src/Kunstmaan/TranslatorBundle/Controller/TranslatorCommandController.php
```diff
@@ -74,6 +74,8 @@ public function importForcedAction()
      */
     public function exportAction()
     {
+        // NEXT_MAJOR this route will be removed because we are now using the common export logic
+        @trigger_error('Calling this route is deprecated since KunstmaanNodeBundle 5.1 and will be removed in KunstmaanNodeBundle 6.0. Use the new route in TranslatorController that uses common export logic.', E_USER_DEPRECATED);
```
acrobat:
```suggestion
        @trigger_error('Using the "KunstmaanTranslatorBundle_command_export" route is deprecated since KunstmaanTranslatorBundle 5.4 and will be removed in KunstmaanTranslatorBundle 6.0. Use the new route in TranslatorController that uses common export logic.', E_USER_DEPRECATED);
```


I don't understand the `Use the new route in TranslatorController that uses common export logic.` part, is there a replacement action? Or is it just the default adminlist export you mean? Otherwise I would reword this to `Use the default adminlist bundle export functionality instead`

src/Kunstmaan/TranslatorBundle/Resources/views/Translator/inline_edit.html.twig
```diff
@@ -1,6 +1,8 @@
-{% if adminlist.configurator.canEditInline(row) %}
+{% spaceless %}
```
acrobat: The `spaceless` tag is deprecated, use `{% apply spaceless%}{% endapply%}` instead

src/Kunstmaan/TranslatorBundle/Controller/TranslatorCommandController.php
```diff
@@ -74,6 +74,8 @@ public function importForcedAction()
      */
     public function exportAction()
     {
+        // NEXT_MAJOR this route will be removed because we are now using the common export logic
+        @trigger_error('Calling this route is deprecated since KunstmaanNodeBundle 5.1 and will be removed in KunstmaanNodeBundle 6.0. Use the default adminlist bundle export functionality instead.', E_USER_DEPRECATED);
```
acrobat:
```suggestion
        @trigger_error('Using the "KunstmaanTranslatorBundle_command_export" route is deprecated since KunstmaanTranslatorBundle 5.4 and will be removed in KunstmaanTranslatorBundle 6.0. Use the default adminlist bundle export functionality instead.', E_USER_DEPRECATED);
```

src/Kunstmaan/TranslatorBundle/Controller/TranslatorCommandController.php
```diff
@@ -74,6 +74,8 @@ public function importForcedAction()
      */
```
acrobat: Also add a `@deprecated` docblock with the same text as the deprecation for IDE completion

src/Kunstmaan/TranslatorBundle/Controller/TranslatorCommandController.php
```diff
@@ -71,9 +71,12 @@ public function importForcedAction()
 
     /**
      * @Route("/export", name="KunstmaanTranslatorBundle_command_export")
+     * @Deprecated Using the "KunstmaanTranslatorBundle_command_export" route is deprecated since KunstmaanTranslatorBundle 5.4 and will be removed in KunstmaanTranslatorBundle 6.0. Use the default adminlist bundle export functionality instead.
```
acrobat:
```suggestion
     * @deprecated Using the "KunstmaanTranslatorBundle_command_export" route is deprecated since KunstmaanTranslatorBundle 5.4 and will be removed in KunstmaanTranslatorBundle 6.0. Use the default adminlist bundle export functionality instead.
```

`@deprecated` should be all lowercase otherwise this error is triggered

```
[Semantical Error] The annotation "@Deprecated" in method Kunstmaan\TranslatorBundle\Controller\TranslatorCommandController::exportAction() was never imported. Did you maybe forget to add a "use" statement for this annotation?
```

acrobat:
@Numkil Can you check the import from spreadsheet with the exported files. As I get some errors I  think are related to the column headers being a different case (eg nl vs NL).

Numkil:
@acrobat 

I tested with a clean and an old project both using master version of the bundles.
If the managed locales in the config is the same as the locales in the headers. (The importer is case insensitive so doesnt matter if in headers = NL and in managed locales = nl). Everything works perfectly. 

With the clean project i only had to remove 'de' from the managed locales because that was not in my test file. (or add de to headers in excel sheet), after that it imported everything correctly.

acrobat:
@Numkil Can you fix the small code style issue reported by styleci? I will also retest my setup, could be possible the error is on my side.

PR #2609, opened 2020-01-10
WIP Feature/croponthefly

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

will probably delete later and make a clean pr


src/Kunstmaan/MediaBundle/Helper/ManipulateImageService.php
```diff
@@ -0,0 +1,59 @@
...
+    /** @var FilterService */
+    private $filterService;
+
+    public function __construct(Slugifier $slugifier, FilterService $filterService)
+    {
+        $this->slugifier = $slugifier;
+        $this->filterService = $filterService;
+    }
+
+    public function manipulateOnTheFly(CroppableMediaLink $croppableMediaLink, string $view = '', string $filter = 'optim'): string
```
acrobat: Maybe call this class `ImageCropService`? Or keep the current classname but rename this method to `crop`/`cropImage`?

src/Kunstmaan/MediaBundle/Resources/config/services.yml
```diff
@@ -164,6 +172,18 @@ services:
             - '@event_dispatcher'
             - '%liip_imagine.cache.resolver.default%'
 
+    Kunstmaan\MediaBundle\Helper\ManipulateImageService:
+        arguments:
+            - '@kunstmaan_utilities.slugifier'
+            - '@liip_imagine.service.filter'
+        public: true
```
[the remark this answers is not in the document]
Numkil: ja anders moet ik de mediacontroller in de bundles refactoren naar die nieuwe abstractcontroller en wou dit niet in deze pull request doen want er is al genoeg complexiteit in deze pr

src/Kunstmaan/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -42,13 +42,52 @@ public function getConfigTreeBuilder()
...
+                        ->arrayNode('focus_point_classes')
+                            ->defaultValue(['top-left' ,'top-center', 'top-right', 'center-left', 'center', 'center-right', 'bottom-left', 'bottom-center', 'bottom-right'])
+                            ->prototype('array')->end()
+                        ->end()
+                        ->arrayNode('custom_views')
+                            ->useAttributeAsKey('groupName')
+                            ->arrayPrototype()
+                                ->addDefaultsIfNotSet()
+                                ->children()
+                                    ->booleanNode('useFocusPoint')->defaultFalse()->end()
```
acrobat: `use_focus_point`

src/Kunstmaan/MediaBundle/Entity/CroppableMediaLink.php
```diff
@@ -0,0 +1,51 @@
...
+     * @ORM\ManyToOne(targetEntity="Kunstmaan\MediaBundle\Entity\Media", cascade={"persist"})
+     * @ORM\JoinColumn(name="media_id", referencedColumnName="id")
+     * @Assert\NotNull()
+     */
+    private $media;
+
+    /**
+     * @ORM\Column(name="run_time_config", type="text", nullable=true)
+     */
+    private $runTimeConfig;
```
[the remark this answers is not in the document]
acrobat: Also runtimeConfig for the property then. If you refactor it with phpstorm the getter/setters and other usages are also fixed (+check the form config).

src/Kunstmaan/MediaBundle/Form/CroppableMediaLinkAdminType.php
```diff
@@ -0,0 +1,71 @@
...
+        if ($croppingViewGroup !== self::DEFAULT && isset($this->croppingViews[self::CUSTOM_VIEWS][$croppingViewGroup]['views'])) {
+            $selectedCroppingViews = $this->croppingViews[self::CUSTOM_VIEWS][$croppingViewGroup]['views'];
+            $useFocusPoint = $this->croppingViews[self::CUSTOM_VIEWS][$croppingViewGroup]['useFocusPoint'] ?? false;
+        }
+        $builder->add('media', MediaType::class, [
+            'label' => 'mediapagepart.image.choosefile',
+            'mediatype' => 'image',
+            'show_cropper_modal' => true,
+            'use_focus_point' => $useFocusPoint,
+            'focus_point_classes' => json_encode($focusPointClasses),
```
acrobat: Maybe we should just pass array's here (also the `cropping_views ` option) which makes it easier to validate this option in the form type. And in the `MediaType` json_encode the value in the `buildView` method

src/Kunstmaan/MediaBundle/Form/CroppableMediaLinkAdminType.php
```diff
@@ -0,0 +1,71 @@
...
+namespace Kunstmaan\MediaBundle\Form;
+
+use Kunstmaan\MediaBundle\Entity\CroppableMediaLink;
+use Kunstmaan\MediaBundle\Form\Type\MediaType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\HiddenType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class CroppableMediaLinkAdminType extends AbstractType
```
acrobat: I think we should rename this formType to a more suitable name as it can do more than cropping

src/Kunstmaan/MediaBundle/Form/CroppableMediaLinkAdminType.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Returns the name of this type.
+     *
+     * @return string The name of this type
+     */
+    public function getBlockPrefix()
+    {
+        return 'croppable_media_link';
```
acrobat: block prefix can be removed as this is already a unique form type and will not interfere with other form types

src/Kunstmaan/GeneratorBundle/Command/GenerateDefaultPagePartsCommand.php
```diff
@@ -67,6 +67,7 @@ protected function doExecute()
             'AbstractPagePart',
             'AudioPagePart',
             'ButtonPagePart',
+            'CroppableImagePagePart',
```
acrobat: We should rename this pagepart as it does more than cropping by now

src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/pagepart/Entity/PageParts/CroppableImagePagePart/CroppableImagePagePart.php
```diff
@@ -0,0 +1,171 @@
...
+use Kunstmaan\MediaBundle\Entity\CroppableMediaLink;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * {{ pagepart }}
+ *
+ * @ORM\Entity
+ * @ORM\Table(name="{{ prefix }}{{ underscoreName }}s")
+ */
+class {{ pagepart }} extends AbstractPagePart
```
acrobat: We should rename this pagepart as it does more than cropping by now

src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/pagepart/Form/PageParts/CroppableImagePagePart/CroppableImagePagePartAdminType.php
```diff
@@ -0,0 +1,73 @@
...
+     * This method is called for each type in the hierarchy starting form the
+     * top most type. Type extensions can further modify the form.
+     * @param FormBuilderInterface $builder The form builder
+     * @param array                $options The options
+     *
+     * @see FormTypeExtensionInterface::buildForm()
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        parent::buildForm($builder, $options);
```
acrobat: The parent call is useless here as the abstract type doesn't add fields

src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/pagepart/Resources/views/PageParts/CroppableImagePagePart/view.html.twig
```diff
@@ -0,0 +1,46 @@
...
+        {{ buildImage.setFilters(resource.media, resource.altText, imageType, resource.media.media.metadata.original_width) }}
+    {% endset %}
+
+    {% if resource.media is not empty %}
+        {% if app.request %}
+            <figure class="image-pp">
+                {% set imageType = resource.media.media.originalFilename|lower|split('.')|last %}
+
+                {% if resource.link is defined and resource.link != '' %}
+                    <a href="{{ resource.link | replace_url }}" {% if resource.openinnewwindow %}target="_blank" rel="noopener"{% endif %}>
```
acrobat:
```suggestion
                    <a href="{{ resource.link | replace_url }}" {% if resource.openInNewWindow %}target="_blank" rel="noopener"{% endif %}>
```

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -27,21 +37,30 @@ public function load(array $configs, ContainerBuilder $container)
         $configuration = new Configuration();
         $config = $this->processConfiguration($configuration, $configs);
 
-        $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
+        if (!isset($config['cropping_views'])) {
+            $config['cropping_views'] = self::DEFAULT_CROPPING_CONFIG;
+        } else {
+            if (empty($config['cropping_views']['default'])) {
```
acrobat: `elseif`

src/Kunstmaan/MediaBundle/Entity/CroppableMediaLink.php
```diff
@@ -0,0 +1,51 @@
...
+     * @Assert\NotNull()
+     */
+    private $media;
+
+    /**
+     * @ORM\Column(name="runtime_config", type="text", nullable=true)
+     */
+    private $runTimeConfig;
+
+    public function getMedia()
```
acrobat: As this is a new entity/class I would add typehints to all methods/parameters/etc

src/Kunstmaan/MediaBundle/Helper/ManipulateImageService.php
```diff
@@ -0,0 +1,82 @@
...
+                $class = $runTimeConfig[$view]['class'];
+            }
+        }
+
+        return $class;
+    }
+
+    public function cropImage(CroppableMediaLink $croppableMediaLink, string $view = '', string $filter = 'optim'): string
+    {
+        /** @var Media $media */
```
acrobat: Can be removed when the correct typehints are added to the entity

src/Kunstmaan/MediaBundle/Resources/config/services.yml
```diff
@@ -40,6 +40,13 @@ services:
         tags:
             - { name: form.type}
 
+    form.type.croppable_media_link:
+        class: Kunstmaan\MediaBundle\Form\CroppableMediaLinkAdminType
```
acrobat: I would change the id to the fqcn and remove the class parameter

```yaml
Kunstmaan\MediaBundle\Form\CroppableMediaLinkAdminType:
    arguments:
        - '%kunstmaan_media.cropping_views%'
    tags:
        - { name: form.type}
```

src/Kunstmaan/MediaBundle/Twig/MediaTwigExtension.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return [
+            new TwigFunction(
+                'cropped_imagine_filter',
+                [$this, 'getCroppedImage']
+            ),
+            new TwigFunction(
+                'get_focus_point_class',
+                [$this, 'getFocusPointClass']
+            ),
```
acrobat: This can be changed to a single line

src/Kunstmaan/MediaBundle/Helper/ManipulateImageService.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $class = '';
+        if ($croppableMediaLink->getRunTimeConfig() !== null) {
+            $runTimeConfig = json_decode($croppableMediaLink->getRunTimeConfig(), true);
+
+            if (
+                is_array($runTimeConfig)
+                && !empty($view)
+                && isset($runTimeConfig[$view]['class'])
+            ) {
```
acrobat: This can be changed to a single line

src/Kunstmaan/MediaBundle/Helper/ManipulateImageService.php
```diff
@@ -0,0 +1,82 @@
...
+    public function __construct(Slugifier $slugifier, FilterService $filterService)
+    {
+        $this->slugifier = $slugifier;
+        $this->filterService = $filterService;
+    }
+
+    public function getFocusPointClass(CroppableMediaLink $croppableMediaLink, string $view = ''): string
+    {
+        $class = '';
+        if ($croppableMediaLink->getRunTimeConfig() !== null) {
```
acrobat: I would changes to method so it does more early returns. This will result in less nested if's and more readable code. Example:

```
if (null === $croppableMediaLink->getRunTimeConfig()) {
    return '';
}
```

src/Kunstmaan/MediaBundle/Twig/MediaTwigExtension.php
```diff
@@ -0,0 +1,49 @@
...
+        ];
+    }
+
+    public function getCroppedImage(CroppableMediaLink $croppableMediaLink, string $view = '', string $filter = null)
+    {
+        if ($filter) {
```
acrobat: Why is this if needed? `$filter` can always be passed to the `cropImage` method, no?

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -16,6 +16,16 @@
  */
 class KunstmaanMediaExtension extends Extension implements PrependExtensionInterface
 {
+    const DEFAULT_CROPPING_CONFIG = [
```
acrobat: add const visibility

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -16,6 +16,16 @@
...
+        'default' => [
+            ['name' => 'desktop', 'width' => 1, 'height' => 1, 'lockRation' => true],
+            ['name' => 'tablet', 'width' => 1, 'height' => 1, 'lockRatio' => true],
+            ['name' => 'phone', 'width' => 1, 'height' => 1, 'lockRatio' => true],
+        ],
+        'focus_point_classes' => ['top-left' ,'top-center', 'top-right', 'center-left', 'center', 'center-right', 'bottom-left', 'bottom-center', 'bottom-right'],
```
acrobat: Move this to a separate constant here or in the Service class. And use that constant also in the configuration default value

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -16,6 +16,16 @@
  */
 class KunstmaanMediaExtension extends Extension implements PrependExtensionInterface
 {
+    const DEFAULT_CROPPING_CONFIG = [
+        'default' => [
+            ['name' => 'desktop', 'width' => 1, 'height' => 1, 'lockRation' => true],
```
acrobat: Typo
```suggestion
            ['name' => 'desktop', 'width' => 1, 'height' => 1, 'lockRatio' => true],
```

And also rename this to `lock_ratio` when the configuration is changed

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -27,21 +37,30 @@ public function load(array $configs, ContainerBuilder $container)
         $configuration = new Configuration();
         $config = $this->processConfiguration($configuration, $configs);
 
-        $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
+        if (!isset($config['cropping_views'])) {
+            $config['cropping_views'] = self::DEFAULT_CROPPING_CONFIG;
```
acrobat: Why don't we set this value as the default value in the configuration class? Would be more consistent and more the "symfony way"

src/Kunstmaan/MediaBundle/Entity/CroppableMediaLink.php
```diff
@@ -0,0 +1,51 @@
...
+namespace Kunstmaan\MediaBundle\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Kunstmaan\AdminBundle\Entity\AbstractEntity;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @ORM\Entity()
+ * @ORM\Table(name="kuma_croppable_media_link")
+ * @ORM\HasLifecycleCallbacks
```
acrobat: Annotation can be removed as it has no lifecycle callbacks?

src/Kunstmaan/MediaBundle/Entity/CroppableMediaLink.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * @ORM\ManyToOne(targetEntity="Kunstmaan\MediaBundle\Entity\Media", cascade={"persist"})
+     * @ORM\JoinColumn(name="media_id", referencedColumnName="id")
+     * @Assert\NotNull()
+     */
+    private $media;
+
+    /**
+     * @ORM\Column(name="runtime_config", type="text", nullable=true)
```
acrobat: Maybe it's better to use the doctrine json type directly here and let doctrine handle the encode/decode? Wdyt?

This will also allow the database to optimize the actual data storage. For example newer mysql servers and postgres have extended json support in their engines. Also our code would be sure to always get an array of data

src/Kunstmaan/MediaBundle/Helper/ManipulateImageService.php
Numkil: This method never returns ' '. it will always return a useable url to the image even if there is no configuration present. for that reason i wrote it without this early return.

src/Kunstmaan/MediaBundle/Twig/MediaTwigExtension.php
Numkil: I wrote it so that when a developer explicitely provides an empty string to the twig function it won't cause the page to give a 500 error. by checking for empty here and allowing the default value of 'optim' to be used in the service method we can avoid that.

src/Kunstmaan/MediaBundle/Entity/CroppableMediaLink.php
Numkil: Its because the frontend writes the array already json encoded in the hidden textfield in the form. otherwise im going to have to decode it in the form type just to have doctrine encode it again when saving

src/Kunstmaan/MediaBundle/Helper/ManipulateImageService.php
acrobat: But it actually does no? If the runtime config is null you won't enter the if and the `$class = '';` value is returned

Numkil: yes you are right i need to do the empty check. thanks i missed that

Numkil: yes sorry i was mistaken

src/Kunstmaan/MediaBundle/Resources/views/Form/formWidgetsOld.html.twig
```diff
@@ -0,0 +1,288 @@
+{% import _self as mediaChooser %}
```
acrobat, line 1: Do we need to copy html to the new template or can the file be removed?

src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/pagepart/Resources/views/PageParts/CroppableImagePagePart/view.html.twig
```diff
@@ -0,0 +1,46 @@
...
+                        {{ image }}
+                    </a>
+                {% else %}
+                    {{ image }}
+                {% endif %}
+
+                {% if resource.caption %}
```
acrobat, line 38: `resource.caption` might not always be declared in the pagepart class, add a `is defined` check

src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/pagepart/Resources/views/PageParts/CroppableImagePagePart/view.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    </noscript>
+{% endmacro %}
+
+{% if resource.media is not empty %}
+    {% set image %}
+        {% set imageType = resource.media.media.originalFilename|lower|split('.')|last %}
+        {{ buildImage.setFilters(resource.media, resource.altText, imageType, resource.media.media.metadata.original_width) }}
+    {% endset %}
+
+    {% if resource.media is not empty %}
```
acrobat, line 25: This must be `resource.media.media` because media is the wrapped class. But seeing this made me think if we should rename that first media property to something more clear? (in the docs + generator)

src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/pagepart/Resources/views/PageParts/CroppableImagePagePart/view.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    <img srcset="{{ srcSet }}" sizes="100vw"{% if altText %} alt="{{ altText }}"{% endif %} width="{{ originalWidth }}"/>
+    <noscript>
+        <img src="{{ imgUrl_big }}"{% if altText %} alt="{{ altText }}"{% endif %} />
+    </noscript>
+{% endmacro %}
+
+{% if resource.media is not empty %}
+    {% set image %}
+        {% set imageType = resource.media.media.originalFilename|lower|split('.')|last %}
+        {{ buildImage.setFilters(resource.media, resource.altText, imageType, resource.media.media.metadata.original_width) }}
```
acrobat, line 22: `resource.altText` might not always be declared in the pagepart class, only pass it if it's defined

docs/bundles/media-bundle/cropping-media-objects.md
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="app_croppable_image_page_parts")
+ */
+class CroppableImagePartPart extends AbstractPagePart
+{
+    /**
+     * @ORM\ManyToOne(targetEntity="Kunstmaan\MediaBundle\Entity\EditableMediaWrapper", cascade={"persist"})
+     * @ORM\JoinColumn(name="media_id", referencedColumnName="id")
+     * @Assert\NotNull()
```
acrobat, line 32: I think this should be a `@Assert\Valid()` as the NotNull is enforced on the Wrapper entity media property.

This and the other comments below  👇 should be checked/applied to the classes in the generator bundle

PR #2612, opened 2020-01-15
[GeneratorBundle] Update DefaultSiteGenerator to support symfony 4

Minhan12:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |

Added check for the new routes.yaml file in symfony 4. If not symfony 4, it will return regular symfony 3 path


acrobat:
@Minhan12 I'm closing this for now as the default recipes include multilanguage routes, so this check is obscelete for installs with sf4. The routing files can be changes in a project when single language is required, [see the recipes for an example](https://github.com/symfony/recipes-contrib/blob/6f9f29199d3cbdb6c8d274e6c5575d445248c952/kunstmaan/admin-bundle/5.2/config/routes/kunstmaan_admin.yaml).

PR #2626, opened 2020-01-28
[NodeBundle] Old version admin preview

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

In the admin you can view older versions of your pages:
versions

However, the preview button always shows the latest version, since the version parameter is completely ignored.

This code will retrieve the correct NodeVersion, instead of always getting the public node version.

### DIFF
--- a/src/Kunstmaan/NodeBundle/Entity/NodeTranslation.php
+++ b/src/Kunstmaan/NodeBundle/Entity/NodeTranslation.php
@@ -395,7 +395,8 @@ class NodeTranslation extends AbstractEntity
      *
      * @return object|null
      */
-    public function getRefByNodeVersion(EntityManagerInterface $em, NodeVersion $nodeVersion) {
+    public function getRefByNodeVersion(EntityManagerInterface $em, NodeVersion $nodeVersion)
+    {
         return $em->getRepository($nodeVersion->getRefEntityName())->find($nodeVersion->getRefId());
     }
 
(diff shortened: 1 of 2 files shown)


acrobat:
@JZuidema thanks for the PR! Are you sure this is the current behaviour? When I test this with the demo application and add a pagepart to the page. The previous version shows correctly without the added pp and the current version shows with the pp.

JZuidema:
@acrobat I've just checked it out, and indeed it works in your demo environment. But you guys are still running version 5.0.12 there. I've just checked the code in 5.0.12, and the code that took care of selecting the right node version actually exists. I'm not sure at which point it gets removed, but the point being is that it does not work anymore in ^5.3

acrobat:
@JZuidema The demo application I mentioned is a locally created project (with demo content) but this one is running v5.4.1.

Can create a small reproducer application? Or check if this is fixed in v5.4? Thanks!

acrobat:
@JZuidema Also the logic to get the correct version is still present in 5.3. The slug controller will get the correct page/nodetranslation and add it to the request.

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/e92fb1178071b25f6ab0774dc79e7919783f6323/src/Kunstmaan/NodeBundle/Controller/SlugController.php#L127-L147

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/e92fb1178071b25f6ab0774dc79e7919783f6323/src/Kunstmaan/NodeBundle/Controller/SlugController.php#L75-L83

So the rendercontext should be correctly populated. Do you have another listener that modifies the response data?

JZuidema:
@acrobat I finally had the time to check this. I've tested this in another kunstmaan project, and this doesn't seem to be an issue there (the older versions are displayed correctly). So I agree with the assumption it is probably a listener that's messing it up.

PR #2627, opened 2020-01-28
[AdminListBundle] Refactor extra actions

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

Turns out not all of the extra actions arent the same for every time of admin list. We couldnt add an export button to `AbstractArticlePageAdminListConfigurator`, because the template used for that adminlist is different than `AbstractDoctrineORMAdminListConfigurator`.
I've tried to merge the `list.html.twig` files as much as possible for the extra actions.

### DIFF
--- a/src/Kunstmaan/ArticleBundle/AdminList/AbstractArticlePageAdminListConfigurator.php
+++ b/src/Kunstmaan/ArticleBundle/AdminList/AbstractArticlePageAdminListConfigurator.php
@@ -196,9 +196,4 @@ abstract class AbstractArticlePageAdminListConfigurator extends AbstractDoctrine
      * @return \Doctrine\ORM\EntityRepository
      */
     abstract public function getOverviewPageRepository();
-
-    public function getIterator()
-    {
-        return $this->getItems();
-    }
 }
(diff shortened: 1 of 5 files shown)


src/Kunstmaan/ArticleBundle/AdminList/AbstractArticlePageAdminListConfigurator.php
```diff
@@ -196,4 +196,9 @@ public function getEntityClassName()
      * @return \Doctrine\ORM\EntityRepository
      */
     abstract public function getOverviewPageRepository();
+
+    public function getIterator()
```
acrobat: Why is this method needed?

acrobat: @JZuidema ☝️

acrobat:
What does actually happen when a user does a override over the list template (what the article generator does by default)? Do the buttons disapear in their application or does everything remain the same?

I would also like to move this to the 5.5 release as this is more of a improvement.

JZuidema:
@acrobat I've applied this patch in one of our projects, and then changed the default list template back to its original form. Basically everything kept working the way it was. And looking at the code, I don't see why it wouldn't. Basically the add_action block is not created, thats the only difference really. The removed code in this merge request was mostly just a duplicate.

PR #2632, opened 2020-02-11
[MediaBundle] Catch exception in DoctrineMediaListener

Minhan12:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When using flush on a new Media with that contains a pdf as content, will throw an exception that will cause the entity manager to close. If using this in an iteration (import) there is no way to open the entity manager again.

This fix makes sure File will be created with url to the file, instead of the content and catched the exception, before being throwed during the flush.


acrobat:
Hi @Minhan12, thanks for your PR. What is the exact reproducer for this issue? I was able to succesfully upload a pdf without any issues with doctrine.

PR #2633, opened 2020-02-17
[RedirectBundle] Fix target column length for long urls

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Target urls could potentially be long and 255 chars could be too few.


Numkil:
@acrobat can't find any sideeffects other than that its possible to break the layout of the adminlist a little bit since the target value is printed out in full in the list view. I think its good

acrobat:
The column grows with the content so it should be ok!

PR #2640, opened 2020-03-02
[AdminBundle] Avoid accessing user token when it is not needed

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Symfony changed the default behavior of the tokenstorage service in symfony 4.4, to track and increase the session usage index. This will trigger the response to be private. 

By changing a few checks we can avoid triggering this meganism when is it not necessary.


acrobat:
Travis failure not related to this change.

PR #2649, opened 2020-03-16
[AllBundles] Add configuration option to disable permission module/checking

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This will allow a project to disable the permission module. This will disable the permission checking on frontend pages (VIEW) and in the admin (VIEW, EDIT, (UN)PUBLISH, ...) and will hide the permissions tab in the page admin.

### DIFF
--- a/src/Kunstmaan/AdminBundle/DependencyInjection/Compiler/EnablePermissionsPass.php
+++ b/src/Kunstmaan/AdminBundle/DependencyInjection/Compiler/EnablePermissionsPass.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Kunstmaan\AdminBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class EnablePermissionsPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if ($container->hasParameter('kunstmaan_node.permissions.enabled')) {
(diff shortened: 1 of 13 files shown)


src/Kunstmaan/AdminBundle/DependencyInjection/Compiler/EnablePermissionsPass.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class EnablePermissionsPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if ($container->hasParameter('kunstmaan_node.permissions.enabled')) {
+            $permissionsEnabledParameter = $container->getParameter('kunstmaan_node.permissions.enabled');
```
Numkil: I would, if possible, create a new parameter specific for the adminbundle with the value of parameter you fetch here. And then use that parameter in the services.yml instead of setting the values directly in this compiler pass.

Only because i think that will make it easier for people who need it in potential future services to know where to get that value and maybe also better for autocompleting

acrobat: ideally this parameter should be removed in another compiler pass as it is only created to pass the config value to the required services. But as this is in line with how we do it in other places, I suggest to keep it for now. We can think of a parameter "policy" in v6, so we only expose parameters that a user should/can use.

PR #2656, opened 2020-03-25
[AdminBundle] Manually register the twig KunstmaanAdminBundle override directory

### LINKED ISSUE #2655: Unable to override Kunstmaan template
Hi,

In all of our Kunstmaan projects we override the default Kunstmaan login template to add our own branding to the page. We do this via app/resources/KunstmaanAdminBundle/views/Security/login.html.twig and it works fine for all our projects that are < Symfony 3.4. However for newer projects this does not seem to work anymore.

I've read in the Symfony docs that they've added a new directory to override bundle templates (templates/bundle), but that does not seem to make any difference.

Is there anyone else who has faced the same issue and found a way to solve this?
Please let me know.

Kind regards,
Kevin

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | fixes #2655

### DIFF
--- a/src/Kunstmaan/AdminBundle/DependencyInjection/KunstmaanAdminExtension.php
+++ b/src/Kunstmaan/AdminBundle/DependencyInjection/KunstmaanAdminExtension.php
@@ -121,6 +121,10 @@ class KunstmaanAdminExtension extends Extension implements PrependExtensionInter
         $monologConfig['handlers']['main']['level'] = 'debug';
         $container->prependExtensionConfig('monolog', $monologConfig);
 
+        // Manually register the KunstmaanAdminBundle folder as a FosUser override for symfony 4.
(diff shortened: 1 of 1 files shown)


acrobat:
@KevinSleegers can you verify this patch works for both your sf3 and sf4 projects? Thanks!

acrobat:
@KevinSleegers does this fix your issue?

acrobat:
ping @KevinSleegers

JoppeDC:
@acrobat I just checked, and can confirm that this allows for project level overrides.

Twig debug command also confirms this.

Screenshot 2020-07-01 at 10 41 00

acrobat:
Thanks for verifying @JoppeDC!

PR #2657, opened 2020-03-27
[NodeBundle] Introduce page_render revert to replace the SlugActionInterface

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

I've deprecated the `SlugActionInterface` and `getControllerAction` method because this solution could not be used during the search index logic. In the search index code was still the old `service` method used to customize the page render. This solution is one that can be used in both cases and without any extra code. Just create an eventsubscribe that hooks into the `kunstmaan_node.page_render` event and check in the event listener/subscriber for what page the event is dispatched. 

TODO:

- [ ] Fix AbstractSearchPage and AbstractFormPage usages of SlugActionInterface. (generate eventsubscriber with both generator commands and deprecate controller action usage for pages generate before this change)


ProfessorKuma (approved):

Thank you for contributing!

acrobat:
Replaced by #2949

PR #2670, opened 2020-05-05
[TranslatorBundle] Choose correct default translation folder based on Symfony version

wesleylancel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | n/a

Currently translations in bundles are not loaded correctly because the default folder is fixed to `Resources/translations` while it should be `translations` for SF4.

### DIFF
--- a/src/Kunstmaan/TranslatorBundle/Service/TranslationFileExplorer.php
+++ b/src/Kunstmaan/TranslatorBundle/Service/TranslationFileExplorer.php
@@ -4,16 +4,10 @@ namespace Kunstmaan\TranslatorBundle\Service;
 
 use Kunstmaan\TranslatorBundle\Service\Exception\TranslationsNotFoundException;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\HttpKernel\Kernel;
 
 class TranslationFileExplorer
 {
-    /**
-     * Symfony default translation folder (in a bundle)
-     *
-     * @var string
-     */
-    private $defaultTranslationFolder = 'Resources/translations';
-
     /**
      * An array of supported file formats to look for
      *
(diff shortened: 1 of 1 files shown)


ProfessorKuma (changes requested):

acrobat:
@wesleylancel this seems indeed a case we forgot to update, where did you encounter an issue with this piece of code?

Can you also change the target branch to 5.4 (and rebase against that branch)? Thanks!

wesleylancel:
@acrobat rebase should be okay. Bug happens when you try to import from composer required bundles.

acrobat:
@wesleylancel Does the bundle use the new bundle structure? Because I think we should keep the default value for now but expand the method importing bundle translation to check both the old and new structure

wesleylancel:
@acrobat Apparently I was wrong and that structure change is a 5.0 change and not 4.0 so the current version number check is wrong either way. 

Do you want to just search both directories in every case?

acrobat:
@wesleylancel Maybe checking both directories would be better, but I which case does it go wrong? As there are a few kernel version checks in place to pass the correct translation folder to the `Kunstmaan\TranslatorBundle\Service\TranslationFileExplorer::find` method.

<details>
  <summary>Imports using a version check</summary>
  
https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/13039a1d896a01b25e2678e9b5890617b30ebd5f/src/Kunstmaan/TranslatorBundle/Service/Command/Importer/ImportCommandHandler.php#L68-L73

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/13039a1d896a01b25e2678e9b5890617b30ebd5f/src/Kunstmaan/TranslatorBundle/Service/Command/Importer/ImportCommandHandler.php#L279
</details>

Does it fix your case if we add the same check to the `importSingleBundleTranslationFiles` method? See https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/13039a1d896a01b25e2678e9b5890617b30ebd5f/src/Kunstmaan/TranslatorBundle/Service/Command/Importer/ImportCommandHandler.php#L195

ProfessorKuma (approved):

Thank you for contributing!

PR #2672, opened 2020-05-07
[AdminBundle, MediaBundle] replace deprecated integer as errorcode in constraints

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |


Numkil:
@acrobat i agree with using the uuid's in the same way as symfony uses it. I have made the requested adjustments

PR #2680, opened 2020-05-22
[mediabundle] Do not allow any and all filetypes to be uploaded to the cms.

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Security testing advised us to remove the ability to upload any and all sorts of media to the cms that could then be used for nefarious purposed through the media/uploads folder.


src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -2,25 +2,80 @@
...
 use Symfony\Component\HttpFoundation\File\File;
+use Symfony\Component\Mime\MimeTypesInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 /**
  * MediaManager
  */
 class MediaManager
 {
+    private const ALLOWED_EXTENSIONS = [
```
acrobat, line 18: I would make both lists of extension configurable through the media bundle config and inject them in the constructor. We can keep the constants (but rename them to `DEFAULT_ALLOWED_EXTENSIONS`?) here but set these arrays ase the default values in the config. This way a config-reference dump will also show these values.

This will give the enduser control over the allowed extensions

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -131,26 +186,30 @@ public function removeMedia(Media $media)
...
+                if ($result) {
+                    return $result;
+                }
             }
+
+            return null;
         }
 
-        return null;
+        return false;
```
acrobat, line 204: `false` is an unexpected return type compared to the previous code (kind of a bc break), would it make sense to return `null` when the extension is not allowed?

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -160,4 +219,26 @@ public function getFolderAddActions()
...
+     */
+    public function isExtensionAllowed(File $file): bool
+    {
+        $mimeType = $this->mimeTypes->guessMimeType($file->getRealPath());
+        $extension = $this->mimeTypes->getExtensions($mimeType)[0];
+        if (!in_array($extension, self::ALLOWED_EXTENSIONS, true)) {
+            return false;
+        }
+
+        if (in_array($extension, self::IMAGE_EXTENSIONS, true)) {
```
acrobat, line 234: I would return early here to avoid deeper nesting of the if's

```php
if (!in_array($extension, self::IMAGE_EXTENSIONS, true)) {
    return true;
}
```

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -160,4 +219,26 @@ public function getFolderAddActions()
 
         return $result;
     }
+
+    /**
+     * @throws Exception
```
acrobat, line 224: Which exeception is thrown?

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -160,4 +219,26 @@ public function getFolderAddActions()
...
+    public function isExtensionAllowed(File $file): bool
+    {
+        $mimeType = $this->mimeTypes->guessMimeType($file->getRealPath());
+        $extension = $this->mimeTypes->getExtensions($mimeType)[0];
+        if (!in_array($extension, self::ALLOWED_EXTENSIONS, true)) {
+            return false;
+        }
+
+        if (in_array($extension, self::IMAGE_EXTENSIONS, true)) {
+            $fileUploadModel = new ImageUploadModel($file);
```
acrobat, line 235: I'm not sure if we need a whole object to just check if a file is corrupt or not, we could perform the check behind the `detectCorrupted` flag manually here. But no strong opinion on this, so others can decide what to do here.

If we keep the object I would unset it after the check just to avoid memory issues on large bulk uploads

src/Kunstmaan/MediaBundle/Model/ImageUploadModel.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Kunstmaan\MediaBundle\Model;
+
+use Symfony\Component\HttpFoundation\File\File;
+use Symfony\Component\Validator\Constraints as Assert;
+
+class ImageUploadModel
```
acrobat, line 10: If we decide to keep this class I would make it final and add the `@internal` phpdoc to exclude it from the BC promise.

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
Numkil: I chose this way because when symfony improves the Assert\Image annotation we would automatically benefit from the improvements without having to adjust any code. If there are better ways to achieve this I am happy to rewrite it.

Numkil:
same issue with symfonyinsight complaining about movemedia not being called movemediaaction

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -160,4 +180,22 @@ public function getFolderAddActions()
 
         return $result;
     }
+
+    public function isExtensionAllowed(File $file): bool
+    {
+        $mimeType = $this->mimeTypes->guessMimeType($file->getRealPath());
+        $extension = $this->mimeTypes->getExtensions($mimeType)[0] ?? 'not_allowed';
```
acrobat, line 187:
```suggestion
        $extensions = $this->mimeTypes->getExtensions($mimeType);
        $extension = array_shift($extensions);
```

This is maybe a better approach, as the value return by `$this->mimeTypes->getExtensions()` will always be an array. `array_shift` will return `null` if the array is empty but that won't break the `in_array` checks.

src/Kunstmaan/MediaBundle/Controller/MediaController.php
```diff
@@ -243,6 +243,15 @@ public function bulkUploadSubmitAction(Request $request, $folderId)
         $folder = $em->getRepository('KunstmaanMediaBundle:Folder')->getFolder($folderId);
         $file = new File($filePath);
 
+        try {
```
acrobat, line 246: Why do we need the full try catch here? Can't we just do

```php
$result = $this->get('kunstmaan_media.media_manager')->isExtensionAllowed($file);
if (false === $result) {
    return $this->returnJsonError('104', 'This file type is not allowed on the server');
}
```

src/Kunstmaan/MediaBundle/Controller/MediaController.php
```diff
@@ -314,20 +323,30 @@ public function dropAction(Request $request, $folderId)
...
+        $mediaManager = $this->get('kunstmaan_media.media_manager');
+        $media = $mediaManager->createNew($drop);
         if ($media) {
             $media->setFolder($folder);
             $em->getRepository('KunstmaanMediaBundle:Media')->save($media);
 
             return new JsonResponse(['status' => $this->get('translator')->trans('kuma_admin.media.flash.drop_success')]);
         }
 
+        if ($mediaManager->isExtensionAllowed($drop)) {
```
acrobat, line 335: Should this be `false === $mediaManager->isExtensionAllowed($drop)`?

acrobat: After reading the code again I see what is happening here, maybe we should turn it around to make it more clear?

Numkil: No I do this second check to keep the old message that would have been returned if a file upload fails because of unknown reasons. To keep the responses like it was and just add my own message if it was specifically not allowed

src/Kunstmaan/MediaBundle/Controller/MediaController.php
```diff
@@ -376,14 +395,29 @@ private function createAndRedirect(Request $request, $folderId, $type, $redirect
             $form->handleRequest($request);
 
             if ($form->isSubmitted() && $form->isValid()) {
-                $media = $helper->getMedia();
-                $media->setFolder($folder);
-                $em->getRepository('KunstmaanMediaBundle:Media')->save($media);
+                $allowed = $mediaManager->isExtensionAllowed($helper->getFile());
```
acrobat, line 398: Looks lika an unnecessary variable and can be inlined in the if statement

src/Kunstmaan/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -69,33 +50,34 @@ public function getConfigTreeBuilder()
...
-                ->end()
-                ->arrayNode('blacklisted_extensions')
-                    ->defaultValue(array('php', 'htaccess'))
-                    ->prototype('scalar')->end()
-                ->end()
-                ->scalarNode('web_root')
-                    ->defaultValue(SymfonyVersion::getRootWebPath())
-                    ->cannotBeEmpty()
-                ->end()
+            ->scalarNode('soundcloud_api_key')->defaultValue('YOUR_CLIENT_ID')->end()
```
acrobat, line 53: Can you restore the indentation?

src/Kunstmaan/MediaBundle/Resources/config/services.yml
```diff
@@ -127,6 +128,14 @@ services:
         arguments:
             - '@kunstmaan_media.filesystem_adapter'
 
+    kunstmaan_media.validator.is_extension_allowed:
+        class: Kunstmaan\MediaBundle\Validator\Constraints\IsExtensionAllowedValidator
```
acrobat, line 132:
```suggestion
    Kunstmaan\MediaBundle\Validator\Constraints\IsExtensionAllowedValidator:
```

src/Kunstmaan/MediaBundle/Validator/Constraints/IsExtensionAllowed.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Kunstmaan\MediaBundle\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @Annotation
+ */
+class IsExtensionAllowed extends Constraint
+{
+    public const NOT_ALLOWED = 1;
```
acrobat, line 12: I would do it the "symfony way" and use an uuid as unique message id. This id will be unique over all constraints.

See: https://github.com/symfony/symfony/blob/e6e1ca38c05d3c42780eb06bc98161ab4629adbf/src/Symfony/Component/Validator/Constraints/Blank.php#L24

src/Kunstmaan/MediaBundle/Validator/Constraints/IsExtensionAllowedValidator.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Component\Translation\TranslatorInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\ConstraintDefinitionException;
+
+class IsExtensionAllowedValidator extends ConstraintValidator
+{
+    /**
+     * @var MediaManager
+     */
```
acrobat, line 17:
```suggestion
    /** @var MediaManager */
```

To keep the same style as the next docblock

Numkil:
@acrobat indentation restored and symfony way used for validator error id

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -160,4 +184,26 @@ public function getFolderAddActions()
...
+    public function isExtensionAllowed(File $file): bool
+    {
+        if (!$this->limitAllowedExtesions) {
+            return true;
+        }
+
+        $mimeType = $this->mimeTypes->guessMimeType($file->getRealPath());
+        $extensions = $this->mimeTypes->getExtensions($mimeType);
+        $extension = array_shift($extensions);
+        if (!in_array($extension, $this->allowedExtensions, true)) {
```
acrobat, line 197: The code itself looks ok, but it yields unexpected results. Forexample if I only allow .txt files to be uploaded and I upload a .json file, this checks allows it.
This is because `$this->mimeTypes->guessMimeType()` for a json files returns `text/plain`, which results in `"txt", "text","conf","def","list","log","in","asc"` as extensions from `$this->mimeTypes->getExtensions` and so is `.txt` picked. And the upload is allowed. Although there is technically no bug in this code, the result is unexpected.

So we should think about a solution so that allowed extensions list is actually the only extensions allowed

Numkil: Ik heb een soort van fallback gemaakt waarbij we kijken dat indien de locale fileinfo text/plain teruggeeft en het een uploadedfile is we fallbacken naar de clientmimetype. Dat is natuurlijk niet 100% foolproof want is spoofable maar at least kunnen we toch iets meer onderscheid maken tussen al die files die anders als "text/plain" gewoon zomaar toegelaten worden.

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -160,4 +184,26 @@ public function getFolderAddActions()
...
+            return false;
+        }
+
+        if (!in_array($extension, $this->imageExtensions, true)) {
+            return true;
+        }
+
+        $errors = $this->validator->validate($file, new Image(['detectCorrupted' => true]));
+
+        return 0 === $errors->count();
```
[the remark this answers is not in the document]
Numkil: @acrobat zou het voor jouw ok zijn als we de translation gewoon wat meer inclusive maken? Bv 

"Your file has been rejected by the server because either the extension is not allowed or the file is corrupt"

?

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
acrobat: I'm not sure about this update as symfony indicates (and you) this value can't be trusted. https://github.com/symfony/symfony/blob/0baa58d4e4bb006c4ae68f75833b586bd3cb6e6f/src/Symfony/Component/HttpFoundation/File/UploadedFile.php#L128-L129

Would it be a solution to move away from the idea of listing extensions and have the user provide an allowed list of mimetypes? Like you would do with the [File constraint](https://symfony.com/doc/current/reference/constraints/File.html#mimetypes)?

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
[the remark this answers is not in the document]
acrobat: Imo we shouldn't group exception messages as it's not clear to the user what the exact problem is and how they can resolve it.

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
```diff
@@ -131,26 +154,53 @@ public function removeMedia(Media $media)
...
+        if ($mimeType === 'text/plain' && $file instanceof UploadedFile) {
+            $mimeType = $file->getClientMimeType();
+        }
+        $extensions = $this->mimeTypes->getExtensions($mimeType);
+        $extension = array_shift($extensions);
+        if (!in_array($extension, $this->allowedExtensions, true)) {
+            return false;
+        }
+
+        if (!in_array($extension, $this->imageExtensions, true)) {
```
acrobat, line 189: And I guess we can also replace the image extension config by just checking the mimetype of the upload and verifying it is image/* . As the Image constraint we are using below is just an extension of the `File` constraint with just [fixed allowed mimetypes](https://github.com/symfony/symfony/blob/367aa1df6caa2c739a58165ead81e7fe50e6599e/src/Symfony/Component/Validator/Constraints/Image.php#L60)

Numkil: Yes good idea that will simplify things.

src/Kunstmaan/MediaBundle/Helper/MediaManager.php
Numkil: The thing is that its just a server thing. because of the way the upload works the uploaded file will be put in the temp folder as "9d0329d30" or something like that without an extension. Because files like json,xml etc are not executable nor have a header in the file to specify what file type they are this effectively makes them just a plaintext file. The good thing is that because of that they are not a threat to the server as they wouldn't be able to be executed. Therefore i think its more a matter of cleanliness to do this optional extra check for files we have already verified are not a threat to the system to deny them from being uploaded if we have not allowed that extension.

Using mimetypes would give us the exact same "problem" because someone might be confused as to why a file with mimetype "text/json" is allowed on the server if he hasn't put it in his allowed configuration.

But yeah tldr: files that get detected by fileinfo on the server as plaintext are no threat and therefore its just a matter of cleanliness to choose to allow/block them from being uploaded which we can do using the clientinfo.

acrobat:
@Numkil I was doing some fixes in the mediabundle and came across this config. Can't this existing feature be used implement the required behaviour?

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/7b775cf26008d63cd5093f391d986bc26e7cc2bd/src/Kunstmaan/MediaBundle/DependencyInjection/Configuration.php#L44-L47

PR #2681, opened 2020-05-27
[MediaBundle][UserManagementBundle] Improve csrf protection in several calls

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Refactor several calls to be more csrf safe


src/Kunstmaan/UserManagementBundle/Resources/views/AdminList/ItemAction/delete.html.twig
```diff
@@ -0,0 +1,42 @@
...
+
+            <!-- Body -->
+            <div class="modal-body">
+                <p>{{ 'settings.user.delete_body'|trans }}</p>
+            </div>
+
+            <!-- Footer -->
+            <div class="modal-footer">
+                {% set deleteform = get_user_delete_form(item) %}
+                {{ form_start(deleteform, {'action': path('KunstmaanUserManagementBundle_settings_users_delete'), 'method': 'POST'}) }}
```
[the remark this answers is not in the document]
acrobat: It does indeed render warnings

image

src/Kunstmaan/UserManagementBundle/Controller/UsersController.php
```diff
@@ -197,57 +196,55 @@ public function editAction(Request $request, $id)
...
-            $em->remove($user);
-            $em->flush();
-
-            $this->addFlash(
-                FlashTypes::SUCCESS,
-                $this->container->get('translator')->trans('kuma_user.users.delete.flash.success.%username%', [
-                    '%username%' => $user->getUsername(),
-                ])
-            );
+        $user = new User();
```
acrobat, line 228: Can we do this? I think we need to use the userManager to get a new instance or inject the correct parameter because we let the user override the user class

src/Kunstmaan/MediaBundle/Controller/MediaController.php
```diff
@@ -159,6 +159,11 @@ public function bulkUploadSubmitAction(Request $request, $folderId)
             @\mkdir($targetDir);
         }
 
+        $submittedToken = $request->headers->get('x-upload-token');
```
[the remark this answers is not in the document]
Numkil: Skipping the check if the header is not present would lead to it being completely useless. The idea is to avoid people being able to send a link to people logged in as admin where they could unknowingly upload media. If people can send the link without the header the security check would not work.

src/Kunstmaan/UserManagementBundle/Controller/UsersController.php
Numkil: Its only for storing the ID in the form. The correct user type is later gotten from the repository with the ID stored in this entity. I will refactor it anyway to use the correct class from the beginning

src/Kunstmaan/UserManagementBundle/Controller/UsersController.php
```diff
@@ -197,57 +196,55 @@ public function editAction(Request $request, $id)
...
         return $params;
     }
 
     /**
-     * Delete a user
-     *
-     * @param Request $request
-     * @param int     $id
-     *
-     * @Route("/{id}/delete", requirements={"id" = "\d+"}, name="KunstmaanUserManagementBundle_settings_users_delete", methods={"POST"})
```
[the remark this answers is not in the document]
Numkil: Do you have any idea how we could avoid BC breaking but still solving the issue until we remove the unsafe method? Otherwise I might have to do some custom work for our project since the client thinks this is quite important.

Numkil:
I fixed all the automated checks except for the warning over the "action" missing in method name of controller since you asked me to revert it.

src/Kunstmaan/UserManagementBundle/Controller/UsersController.php
```diff
@@ -223,14 +219,50 @@ public function editAction(Request $request, $id)
...
+
+    /**
+     * @Route("/form-delete/{id}", requirements={"id" = "\d+"}, name="KunstmaanUserManagementBundle_settings_users_form_delete", methods={"POST"})
+     */
+    public function deleteFormAction(Request $request, $id)
     {
         $this->denyAccessUnlessGranted('ROLE_SUPER_ADMIN');
 
+        $submittedToken = $request->request->get('token');
+        // 'delete-item' is the same value used in the template to generate the token
```
acrobat, line 261: Incorrect comment can be removed

acrobat:
Can you add some lines to the upgrade doc, so users know what is changed. Forexample the required "x-upload-token" on the bulkupload route, deprecated delete route etc

Numkil:
@acrobat I updated the upgrade file

UPGRADE-5.6.md
```diff
@@ -4,10 +4,24 @@ UPGRADE FROM 5.5 to 5.6
...
 SearchBundle
 ------------
 
-Support for `ruflin/elastica` and elasticsearch versions below 7 is deprecated and will be removed in 6.0. Upgrade to `ruflin/elastica` ^7.0 and elasticsearch 7.
+* Support for `ruflin/elastica` and elasticsearch versions below 7 is deprecated and will be removed in 6.0. Upgrade to `ruflin/elastica` ^7.0 and elasticsearch 7.
+
+MediaBundle
+------------
+
+* The controller method "KunstmaanMediaBundle_media_bulk_upload_submit" is now POST only.
```
acrobat, line 18: This entry has to be in new `UPGRADE-5.7.md` file, 5.6 was already released!

UPGRADE-5.6.md
```diff
@@ -4,10 +4,24 @@ UPGRADE FROM 5.5 to 5.6
...
+This call now also requires a mandatory header called "x-upload-token" The content of this header
```
acrobat, line 19:
```suggestion
This call now also requires a mandatory header called "x-upload-token". The content of this header
```

UPGRADE-5.6.md
```diff
@@ -4,10 +4,24 @@ UPGRADE FROM 5.5 to 5.6
...
+should be a csrf token generated by the Symfony csrf service using the name "bulk-upload-media".
+The reason for this change is that we detected it would be possible to send an admin who is logged in to the system
+a link that upon being clicked would upload a random file of the internet to the cms.
```
acrobat, line 22:
```suggestion
This header is an added security layer to avoid unintended and malicious uploads.
```

Numkil:
Moved the content over to 5.7

PR #2692, opened 2020-06-19
[AdminBundle][AdminListBundle] translation in admin fixes

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

Fix tag in services for adminlocalelistener and hardcode export header translations so it doesn't break import on cms without an english admin

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Service/ExportService.php
+++ b/src/Kunstmaan/AdminListBundle/Service/ExportService.php
@@ -92,7 +92,7 @@ class ExportService
             $row = [];
             /** @var Field $column */
             foreach ($adminList->getExportColumns() as $column) {
-                $row[] = $this->translator->trans($column->getHeader());
+                $row[] = $this->translator->trans($column->getHeader(), [], null, 'en');
             }
             $writer->addRow($row);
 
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/AdminListBundle/Service/ExportService.php
```diff
@@ -92,7 +92,7 @@ protected function streamOutput(ExportableInterface $adminList, $format)
             $row = [];
             /** @var Field $column */
             foreach ($adminList->getExportColumns() as $column) {
-                $row[] = $this->translator->trans($column->getHeader());
+                $row[] = $this->translator->trans($column->getHeader(), [], null, 'en');
```
acrobat, line 95: Why is this change needed?

acrobat:
Can you rebase this onto the 5.4 branch? Thanks!

src/Kunstmaan/AdminListBundle/Service/ExportService.php
Numkil: because the importer logic is hardcoded to english. on projects where your admin has an adminlocale different to english the exported file will have the headers domain/keyword/status translated. When importing the same excel file after lets say adding some new translations the cms will respond saying the excel file is invalid because it does not find the required headers domain/keyword/status

Numkil: By making it so that the excel file exported will contain the headers standardized in english the export / import flow will be garuanteed to stay working

Issue #2689, opened 2020-06-11
[AdminListBundle] Modify EnumerationFilterType to use js-advanced-select


molovo:
The standard multi-select box used for filtering with the EnumerationFilterType is very unwieldy to use when there is a long list of options. It would be much more user-friendly to enable the JS Advanced Select to allow searching through the options.

Attempts at implementing this internally failed as when the relevant classes are added to the enumeration field, the JS which builds the field names skips over any fields using the advanced select.

acrobat:
Fixed in #2707

PR #2707, opened 2020-08-04
[AdminBundle] advanced js select for enumeration filter

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2689

Fixed displaying the multiselect as advanced js select when using EnumerationFilter. There were a couple problems:
- Adding the `advanced-js-select` class broke the javascript, because it creates an input type, which doesn't have the attribute name (which the code assumes it does).
- The advanced js select is broken when the page initializes, because the filters are hidden by default. Therefor it doesn't initialize correctly.
- The `select2-container` class has a width of `100% !important`, which causes the advanced js select in the filter to barely have any width.

afbeelding


acrobat:
Thanks @JZuidema!

PR #2739, opened 2020-09-11
[TranslatorBundle] Don't use temporary filename for spreadsheet translation imports

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

When importing translations from a spreadsheet the full temporary filepath (eg. `/private/var/folders/7x/tddvmt1x6lx5frst752qqzp80000gy/T/phpE8splr`) is used and this might exceed the column length of `Translation::$file`. So don't pass it for spreadsheet imports as there is no filename available that would make sense.


ProfessorKuma (approved):

Thank you for contributing!

PR #2744, opened 2020-09-14
[NodeSearchBundle] Introduce SearchViewRenderer service to deprecate/remove remaining templating usages

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
+++ b/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
@@ -16,6 +16,7 @@ use Kunstmaan\NodeBundle\Helper\RenderContext;
 use Kunstmaan\NodeSearchBundle\Event\IndexNodeEvent;
 use Kunstmaan\NodeSearchBundle\Helper\IndexablePagePartsService;
 use Kunstmaan\NodeSearchBundle\Helper\SearchViewTemplateInterface;
+use Kunstmaan\NodeSearchBundle\Services\SearchViewRenderer;
 use Kunstmaan\PagePartBundle\Helper\HasPagePartsInterface;
 use Kunstmaan\SearchBundle\Configuration\SearchConfigurationInterface;
 use Kunstmaan\SearchBundle\Provider\SearchProviderInterface;
@@ -23,7 +24,6 @@ use Kunstmaan\SearchBundle\Search\AnalysisFactoryInterface;
 use Kunstmaan\UtilitiesBundle\Helper\ClassLookup;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
-use Symfony\Component\DomCrawler\Crawler;
 use Symfony\Component\EventDispatcher\LegacyEventDispatcherProxy;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Acl\Domain\ObjectIdentity;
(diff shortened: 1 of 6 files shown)


ProfessorKuma (approved):

Thank you for contributing!

src/Kunstmaan/NodeSearchBundle/Services/SearchViewRenderer.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->indexablePagePartsService = $indexablePagePartsService;
+        $this->requestStack = $requestStack;
+    }
+
+    public function renderDefaultSearchView(NodeTranslation $nodeTranslation, HasPagePartsInterface $page, string $defaultView = '@KunstmaanNodeSearch/PagePart/view.html.twig')
+    {
+        $html = $this->twig->render($defaultView, [
+            'locale' => $nodeTranslation->getLang(),
+            'page' => $page,
+            'pageparts' => $this->indexablePagePartsService->getIndexablePageParts($page),
```
Numkil, line 37: This is maybe nitpicky because i understand that this code is copied from where it was before and not new. But it would be really helpful if this pageparts variable was also available in the renderCustomSearchView parameters. That way if you only need a very basic adjustment to be done requiring a custom search view but you don't wanna bother with rendering the pagetemplates and such it would be possible.

src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
```diff
@@ -706,33 +728,19 @@ protected function getUTCDateTime(\DateTime $dateTime)
...
-        // Remove HTML markup
-        $result = strip_tags($text);
+        @trigger_error(sprintf('The "%s" method is deprecated since KunstmaanNodeSearchBundle 5.7 and will be removed in KunstmaanNodeSearchBundle 6.0. Use the "removeHtml" method of the "%s" service instead.', __METHOD__, SearchViewRenderer::class), E_USER_DEPRECATED);
 
-        // Decode HTML entities
-        $result = trim(html_entity_decode($result, ENT_QUOTES));
+        $searchViewRenderer = $this->container->get('kunstmaan_node_search.service.search_view_renderer');
 
-        return $result;
+        return $searchViewRenderer->removeHtml($text);
```
Numkil, line 743: is there a reason called the searchViewRenderer service only for the removeHTML call and not for the other 2?

Numkil: I think the reason is if someone had only overwritten the removeHTML call then their changes would break on upgrade because it would now be called somewhere else..

acrobat: Yes because fro the methods in the class renderDefaultSearchView and renderCustomSearchView we need to call those methods in case they are overriden by the user otherwise we can just call the new code (this way we won't trigger deprecations ourself

Numkil (approved):

src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
acrobat: We still call `$this->removeHtml()` in the old methods and if users override those render methods they still will can `$this->removeHtml()` so it would end up in their override or in the method and it will use the new class

src/Kunstmaan/NodeSearchBundle/Services/SearchViewRenderer.php
acrobat: This can be implemented in a separate PR as I would consider it a improvement/feature

PR #2765, opened 2020-09-23
[WIP][NodeBundle] Disallow trailing slash in slug router

waaghals:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | N/A

Changed the regex to not allow trailing slashes. This way, trailing slashes will fallback to the symfony router, which (when correctly configured) will redirect the page without the slash.

This fixes a bug where the main page under a locale path (example.com/en) has duplicate content because both the trailing slash and non trailing slash return the same content.


ProfessorKuma (changes requested):

acrobat:
@waaghals I think we then also need to make sure that when saving a slug we remove any trailing slashes, because currently that might be possible

ProfessorKuma (changes requested):

ProfessorKuma (approved):

Thank you for contributing!

waaghals:
I realized this won't work when the slug is left empty when saving a Node. I'll have to figure out a different solution

PR #2776, opened 2020-10-02
[AllBundles] Configure asset packages and add package to asset call

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Configure asset package in Symfony and use this for all asset calls. With this change it is possible to configure a different base_url for different assets. The generate urls are default the same as before.

Also a few changes from #2769 are added in this PR.


ProfessorKuma (changes requested):

ProfessorKuma (approved):

Thank you for contributing!

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -116,6 +116,16 @@ public function prepend(ContainerBuilder $container)
...
 
+        $frameworkExtensionConfig = [
+            'assets'=> [
+                'packages' => [
+                    'asset' => null,
+                    'media' => null,
+                ],
+            ],
+        ];
+        $container->prependExtensionConfig('framework', $frameworkExtensionConfig);
```
dannyvw, line 127: Not sure if this is the right bundle, but there is no core bundle available. Maybe move to the admin bundle?

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -116,6 +116,16 @@ public function prepend(ContainerBuilder $container)
 
         $container->prependExtensionConfig('doctrine', $doctrineGedmoEntityConfig);
 
+        $frameworkExtensionConfig = [
+            'assets' => [
+                'packages' => [
+                    'asset' => null,
```
acrobat, line 122: I'm thinking, do we need the asset config? Can we use the asset function without the package parameter for all calls in the code, except for the media assets. Which can use the "media" package. Is that something that could work?

dannyvw: @acrobat That works too. I have updated the PR.

dannyvw:
@acrobat Do you have an update for this?

acrobat:
@dannyvw I've assigned this PR to myself, so I can test it in a real project. Can you rebase against the current master as it makes it easier to test it and have the new github actions setup here.

dannyvw:
@acrobat Done

PR #2784, opened 2020-10-09
[MediaBundle] Change media path usage

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

This makes it easier to use one remote storage configuration for the complete application


ProfessorKuma (changes requested):

ProfessorKuma (approved):

Thank you for contributing!

dannyvw:
@acrobat Do you have an update for this?

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -43,7 +43,7 @@ public function load(array $configs, ContainerBuilder $container)
         $container->setParameter('kunstmaan_media.enable_pdf_preview', $config['enable_pdf_preview']);
         $container->setParameter('kunstmaan_media.blacklisted_extensions', $config['blacklisted_extensions']);
         $container->setParameter('kunstmaan_media.web_root', $config['web_root']);
-        $container->setParameter('kunstmaan_media.full_media_path', $config['web_root'] . '%kunstmaan_media.media_path%');
+        $container->setParameter('kunstmaan_media.full_media_path', $config['web_root'] . '%kunstmaan_media.media_path%'); // Not used, deprecate?
```
acrobat, line 46: If this is not used we need to deprecate this, maybe it's only possible to have an entry in the upgrade file to warn users. And add a `// NEXT_MAJOR` command so we know that it should be removed

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
```diff
@@ -74,12 +74,12 @@ public function load(array $configs, ContainerBuilder $container)
...
     public function prepend(ContainerBuilder $container)
     {
-        if (!$container->hasParameter('kunstmaan_media.upload_dir')) {
+        if (!$container->hasParameter('kunstmaan_media.upload_dir')) {  // Not used, deprecate?
             $container->setParameter('kunstmaan_media.upload_dir', '/uploads/media/');
         }
 
         $twigConfig = [];
-        $twigConfig['globals']['upload_dir'] = $container->getParameter('kunstmaan_media.upload_dir');
+        $twigConfig['globals']['upload_dir'] = $container->getParameter('kunstmaan_media.upload_dir'); // Not used, deprecate?
```
acrobat, line 82: The parameter might not be used, but the twig global still is? So do we need to replace that value somehow or is that also a variable that needs to be deprecated?

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
dannyvw: Added an entry to the upgrade file

src/Kunstmaan/MediaBundle/DependencyInjection/KunstmaanMediaExtension.php
dannyvw: I cannot find a reference to the twig global, but i have removed the comment. It is not necessary to remove this.

PR #2805, opened 2020-10-28
[NodeBundle] Improve nodetranslation repository deprecation

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

Follow up for #2716. Don't use the deprecated repository method in our own code and remove the int typehint for now (and deprecate passing other values than integers) to avoid an indirect BC break as passing an object was possible in the past (although the phpdoc didn't indicate this) and actually produced the expected results


ProfessorKuma (changes requested):
* It seems that you should have submitted to master.

PR #2831, opened 2020-11-10
[RedirectBundle] Automatically generate redirects

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

When the slug of a page is changed, if you visit the old url it will result in a 404. This feature will allow Kunstmaan to automatically generate a redirect for each page that implements the `AutoRedirectInterface`

When a slug is changed, the subscriber performs these steps:

- Find all redirects with origin the newly created url. Delete these redirects. (because it is going to be an actual url, you don't need to have an existing redirect for it
- Find all redirects with target the old url. Update the target to the newly created url. (so you don't have to chain redirects going from A to B to C to D, but instead going from A to D and from B to D)
- Create a new redirect with origin the old url and target the newly created url.

Here is an example:

This is the current page:
afbeelding

We rename the page to Page B and change the slug accordingly:
afbeelding

We now have one automatically generated redirect:
afbeelding

We rename the page to Page C and change the slug accordingly:
afbeelding

We now have two automatically generated redirects. Both redirects now target `/page-c` (instead of `page-a` redirecting to `page-b` etc)
afbeelding

We regret all the name changes, and rename the page back to Page A and change the slug accordingly:
afbeelding

We now still have to redirects where `page-b` and `page-c` both redirect back to `page-a`
afbeelding

### DIFF
--- a/src/Kunstmaan/RedirectBundle/AdminList/RedirectAdminListConfigurator.php
+++ b/src/Kunstmaan/RedirectBundle/AdminList/RedirectAdminListConfigurator.php
@@ -140,6 +140,10 @@ class RedirectAdminListConfigurator extends AbstractDoctrineORMAdminListConfigur
             return 'All domains';
         }
 
+        if ($columnName === 'isAutoRedirect' && $item->isAutoRedirect() === null) {
+            return false;
+        }
+
         return parent::getValue(
             $item,
             $columnName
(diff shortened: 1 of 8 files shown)


ProfessorKuma (changes requested):

ProfessorKuma (approved):

Thank you for contributing!

acrobat:
@JZuidema I'm thinking if we need the marker interface at all, would it make sense to have a configuration option in the node bundle to activitate auto redirects for all cms pages?

PR #2811, opened 2020-11-03
[All Bundles] Added PostgreSQL support

igst:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Based on https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1778 and https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/1810

This PR makes it possible to use Kunstmaan Bundles CMS with a PostgreSQL database.

PR #2849, opened 2021-02-10
[All bundles] added postgres support

IntegraConsult:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed tickets | 

Upgraded Kunstmaan 5.7.1 so it supports posgresql. A lot of effort was put into keeping the code compatible with the curent supportted databases, so I do not expect it to break current functionality.
Following test were done before this pull request:
- manual click through the default application => no errors found
- kunstmaan php unit test suite => no errors found
-


ProfessorKuma (changes requested):
* Your answer if this PR is a bugfix seems to be incorrect.
* Your answer if this PR is a new feature seems to be incorrect.
* Your answer if this PR is BC breaking seems to be incorrect.
* Your answer if this PR is has deprecations seems to be incorrect.

ProfessorKuma (dismissed):

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclHelper.php
```diff
@@ -204,6 +223,76 @@ private function getPermittedAclIdsSQLForUser(Query $query)
         return $selectQuery;
     }
 
+    /**
+     * @return string
+     */
+    private function getPermittedAclIdsSQLForUserPlatformPostgres(Query $query)
```
acrobat, line 229: Why is the whole query duplicated? Can't the solution from [#2811](https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/2811/files#diff-2ae42119f573a6a61e26009641aa186d97e0f77b951dc02385d67d9daa075854) be used?

src/Kunstmaan/AdminListBundle/Helper/DoctrineDBALAdapter.php
```diff
@@ -66,10 +71,19 @@ public function getNbResults()
         if ($this->useDistinct) {
             $distinctString = 'DISTINCT ';
         }
-        $statement = $query->select('COUNT(' . $distinctString . $this->countField . ') AS total_results')
-            ->orderBy($this->countField)
-            ->setMaxResults(1)
-            ->execute();
+
+        if ($this->databasePlatform == 'postgresql') {
```
acrobat, line 75: Is this check necessary? I saw that in [#2811](https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/2811/files#diff-ad01d84409afe7ac963af2ce17fd3a076a100eed9c35b6d5e13c69cc75ef46df) the order by was removed and that seems like the right solution as a count with order by isn't needed I think

src/Kunstmaan/MediaBundle/Repository/FolderRepository.php
```diff
@@ -105,10 +105,18 @@ private function deleteChildren(Folder $folder)
      */
     public function getAllFolders($limit = null)
     {
-        $qb = $this->createQueryBuilder('folder')
-            ->select('folder')
-            ->where('folder.parent is null AND folder.deleted != true')
-            ->orderBy('folder.name');
+        if ($this->_em->getConnection()->getDatabasePlatform()->getName() == 'postgresql') {
```
acrobat, line 108: Why is this check necessary?  The mysql variant can also use the new parameter, so we can avoid duplicated queries.

src/Kunstmaan/NodeBundle/Repository/NodeTranslationRepository.php
```diff
@@ -153,7 +174,7 @@ public function getChildrenQueryBuilder(Node $parent, $lang = null)
     public function getOnlineChildrenQueryBuilder(Node $parent, $lang = null)
     {
         return $this->getChildrenQueryBuilder($parent, $lang)
-            ->andWhere('nt.online = true');
+            ->andWhere('nt.online = :onlineTrue')->setParameter('onlineTrue', true);
```
acrobat, line 177:
```suggestion
            ->andWhere('nt.online = :onlineTrue')
            ->setParameter('onlineTrue', true)
        ;
```

src/Kunstmaan/NodeBundle/Repository/NodeTranslationRepository.php
```diff
@@ -307,12 +329,14 @@ public function getAllNodeTranslationsForUrl(
                 'WITH',
                 'b.publicNodeVersion = v.id'
             )
+
```
acrobat, line 332:
```suggestion
```

src/Kunstmaan/NodeBundle/Repository/NodeTranslationRepository.php
```diff
@@ -506,14 +531,16 @@ public function getBestMatchForUrl($urlSlug, $locale)
...
             ->createNativeQuery(
                 'select nt.*
-                from kuma_node_translations nt
-                join kuma_nodes n on n.id = nt.node_id
-                where n.deleted = 0 and nt.lang = :lang and locate(nt.url, :url) = 1
-                order by length(nt.url) desc limit 1',
+                         from kuma_node_translations nt
+                        join kuma_nodes n on n.id = nt.node_id
+                        where n.deleted = :deletedFalse and nt.lang = :lang and locate(nt.url, :url) = 1
+                        order by length(nt.url) desc limit 1',
```
acrobat, line 537: Keep the indentation as it was before

src/Kunstmaan/TranslatorBundle/Repository/TranslationRepository.php
```diff
@@ -62,9 +62,10 @@ public function getLastChangedTranslationDate()
         $flagNew = \Kunstmaan\TranslatorBundle\Entity\Translation::FLAG_NEW;
         $flagUpdated = \Kunstmaan\TranslatorBundle\Entity\Translation::FLAG_UPDATED;
 
+        //use lowercase variable names for postgress. i.e. : newest_date instead of newestDate
```
acrobat, line 65: Can you remove this comment. It makes sense in this PR and the context of adding postgres support, but this info is not needed after it's merged

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclHelper.php
```diff
@@ -55,6 +60,7 @@ class AclHelper
     public function __construct(EntityManager $em, TokenStorageInterface $tokenStorage, RoleHierarchyInterface $rh, $permissionsEnabled = true)
     {
         $this->em = $em;
+        $this->databasePlatform = $this->em->getConnection()->getDatabasePlatform()->getName();
```
acrobat, line 63: When the query is not duplicated any more I would inline this check in the query method, there is no need to have this as a class property

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclNativeHelper.php
```diff
@@ -45,20 +50,126 @@ class AclNativeHelper
...
+     * Apply the ACL constraints to the specified query builder, using the permission definition, for all database platforms
      *
      * @param QueryBuilder         $queryBuilder  The query builder
      * @param PermissionDefinition $permissionDef The permission definition
      *
      * @return QueryBuilder
      */
     public function apply(QueryBuilder $queryBuilder, PermissionDefinition $permissionDef)
+    {
+        if ($this->databasePlatform == 'postgresql') {
```
acrobat, line 69: See comments in `AclHelper.php` about duplicate query

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclWalker.php
```diff
@@ -24,10 +24,19 @@ public function walkFromClause($fromClause)
...
+tempAclView;
+                break;
+            default:
+                $tempAclView = <<<tempAclView
 JOIN ({$extraQuery}) ta_ ON {$tableAlias}.id = ta_.id
 tempAclView;
-
+                break;
+        }
+        //var_dump('ACLWalker::walkFromClause',$sql);echo"<br/>";
```
acrobat, line 39: Cleanup this var_dump

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclNativeHelper.php
```diff
@@ -138,6 +249,7 @@ public function apply(QueryBuilder $queryBuilder, PermissionDefinition $permissi
 AND e.mask & {$mask} > 0
 SELECTQUERY;
 
+        //var_dump('ACLNativeWalker::applyPlatformOther',$joinTableQuery); echo"<br/>";
```
acrobat, line 252: Cleanup this var_dump

acrobat (changes requested):
Hi @IntegraConsult,

Thanks for the PR, the effort is appreciated!
I did a first review and left some comments. As a general comment I would like to keep the query parameters "neutral". So `:onlineTrue` -> `:online`, this to avoid confusion if the value would change in the future but the parameter not.

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclHelper.php
IntegraConsult: As a matter of fact I initially did use the solution from #2811 and tested it with one of KBC bank core products. 2811 breaks access control in mysql, due to various issues (amongst which the databaseprefix I think). In order to keep things backwards compatibility with mysql I thiought it prudent to implement a dedicated method for postgres. However as you can see my implementation keeps many of the good stuff of 2811. 
The alternative would have been to place some more (if (dtatabaseplatform==) {}else{} statements in the opriginal code. I do not think that improves code readibility, but perhaps thsi si a matter of tasete

src/Kunstmaan/AdminListBundle/Helper/DoctrineDBALAdapter.php
IntegraConsult: Yes, I believe this check to be neccesary. In #2811 orderby was removed becaused it creates postgres errors when used in conjunction with COUNT as it is here. Simply removing the order by - because its not neccesary- from the original kunstmaan code is unfortunately not enogh, this may work for simple queries that are provided by querybuilder. However when queries provided by queryBuilder get more complex and contain orderBys in themselves we are back to square one. Hence the check

src/Kunstmaan/MediaBundle/Repository/FolderRepository.php
IntegraConsult: Good point. In hindsight I do not know why I did that either. Perheaps being a little overzealous trying to protect backward compatibility. No matter. check will be removed.

src/Kunstmaan/NodeBundle/Repository/NodeTranslationRepository.php
IntegraConsult: OK, done

IntegraConsult:
> Hi @IntegraConsult,
> 
> Thanks for the PR, the effort is appreciated!
> I did a first review and left some comments. As a general comment I would like to keep the query parameters "neutral". So `:onlineTrue` -> `:online`, this to avoid confusion if the value would change in the future but the parameter not.

See your point, will update

src/Kunstmaan/TranslatorBundle/Repository/TranslationRepository.php
IntegraConsult: done

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclHelper.php
IntegraConsult: Good point ! done

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclWalker.php
IntegraConsult: done

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclNativeHelper.php
IntegraConsult: done

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclHelper.php
IntegraConsult: Agree, done

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclWalker.php
IntegraConsult: done

src/Kunstmaan/AdminListBundle/Helper/DoctrineDBALAdapter.php
IntegraConsult: Sometimes the queryBuilder query itself may contain orderBy clauses which will reek havoc in postgress when used in conjuction with COUNT DISTINCT. Therefore it is not sufficient to simply remove ones Kunstmaans own orderBy clause, but also the orderby clauses of the querybuilder query itself.  IMO the reason that the orderBy was removed from the kunstmaan code was not that it is not neccesary, but in postgres it is detrimental.

src/Kunstmaan/MediaBundle/Repository/FolderRepository.php
IntegraConsult: Absolutuetly right: ccheck is taken out

src/Kunstmaan/NodeBundle/Repository/NodeTranslationRepository.php
IntegraConsult: done

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclNativeHelper.php
IntegraConsult: done

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclNativeHelper.php
IntegraConsult: Same answer as above applies. The original 2811 was tested against one of KBC core system and caused crashes all over the place. To make abolutely sure backward compatibility with msql is guearanted I opted to do a separate method for postgres. This is not critical/essential however and could be refactored to a single method with some mor if(){}else{} clauses. Apart from processing a few more ifs, IMO this makes the code less readable, so I would not do it ifg the product was my own. However I also see that this can be a matter of taste.

IntegraConsult:
Hello Jeroen, Ad Langenkamp here

I went through review comments, and made the appropriate code changes. Some
of the original 2811 PR needed to be reworked so they would work in a more
general case. This particularly pertains to AclHelper, AclNativeHelper and
DoctrineDBALAdapter.

Can you please have a look and see if the changes are OK as they are, or
perhaps a little more needs to be done?.

Thank you

Ad Langenkamp

(currently working for ICAPPS/KBC bank belgium)


Op wo 17 feb. 2021 om 20:32 schreef Jeroen Thora <notifications@github.com>:

src/Kunstmaan/AdminBundle/Repository/ExceptionRepository.php
```diff
@@ -26,8 +27,10 @@ public function markAllAsResolved()
     {
         return $this->createQueryBuilder('e')
             ->update()
-            ->set('e.isResolved', 1)
-            ->where('e.isResolved = 0')
+            ->set('e.isResolved', ':isResolvedTrue')
+            ->where('e.isResolved = :isResolvedFalse')
+            ->setParameter('isResolvedTrue', true, \PDO::PARAM_BOOL)
```
acrobat, line 32: I don't know if it's needed to pass the 3th parameter here (I know it brings a speed improvement for `\DateTime` objects), so we should decide to add or not add it. Because now it's a mix of both, so if it's not required I would suggest not adding it.

EDIT: or is it here required because it's a parameter for an update field?

src/Kunstmaan/AdminListBundle/Helper/DoctrineDBALAdapter.php
acrobat: Yes I know postgres has issues with that, but what I meant was can't this be done for "all" (mysql and postgress) platforms? So we remove the general orderBy and reset the OrderBy on the query object.

src/Kunstmaan/MediaBundle/Repository/FolderRepository.php
```diff
@@ -107,7 +107,8 @@ public function getAllFolders($limit = null)
     {
         $qb = $this->createQueryBuilder('folder')
             ->select('folder')
-            ->where('folder.parent is null AND folder.deleted != true')
+            ->where('folder.parent is null AND folder.deleted != :deleted')
+            ->setParameter('deleted', true, \PDO::PARAM_BOOL)
```
acrobat, line 111: Same question as in `src/Kunstmaan/AdminBundle/Repository/ExceptionRepository.php`. If it's not required I would remove the 3th parameter as we don't set it anywhere else.

src/Kunstmaan/NodeBundle/Repository/NodeTranslationRepository.php
```diff
@@ -255,7 +278,8 @@ private function getNodeTranslationForSlugPart(
                 'WITH',
                 't.publicNodeVersion = v.id'
             )
-            ->where('n.deleted != 1')
+            ->where('n.deleted != :deleted')
+            ->setParameter('deleted', true, \PDO::PARAM_BOOL)
```
acrobat, line 282: See comment/question above about the 3th parameter

src/Kunstmaan/AdminBundle/Helper/Security/Acl/AclHelper.php
acrobat: I will look into this myself to see what the posibilities are. But my gut feeling is that this is not ideal as this just duplicates the full code. Maybe just inner parts need to be duplicated. But let's see what an ideal solution can be here!

acrobat (changes requested):
Thanks for the updated parameter names, they look good to me! I left some extra comments/questions before actually testing this PR in some real life projects.

src/Kunstmaan/AdminListBundle/Helper/DoctrineDBALAdapter.php
IntegraConsult: OK, made the change also for non postgress and extensively retested that against mysql database. You are right. This works also for mysql, so no if statement needed and cleaner code. Thanks!

src/Kunstmaan/AdminBundle/Repository/ExceptionRepository.php
IntegraConsult: Your right, this is not consistent, have removed them

src/Kunstmaan/MediaBundle/Repository/FolderRepository.php
IntegraConsult: same answer: PARAM_BOOL's have been removed

IntegraConsult: Yep, has been removed

src/Kunstmaan/NodeBundle/Repository/NodeTranslationRepository.php
IntegraConsult: parameter has beeen removed

src/Kunstmaan/AdminBundle/Repository/ExceptionRepository.php
IntegraConsult: parameter was removed

IntegraConsult:
As you can see I made the changes you requested. Thanks for posing your review remarks in such a constructive manner. It really helped me!
Ad

acrobat (changes requested):
@IntegraConsult I've created a PR against your fork with extra fixes. Can you review and verify in your project? 

See: https://github.com/IntegraConsult/KunstmaanBundlesCMS/pull/1

acrobat (changes requested):

src/Kunstmaan/NodeBundle/Repository/NodeRepository.php
```diff
@@ -327,12 +345,25 @@ public function getAllMenuNodes(
...
+        if (!$includeHiddenFromNav) {
+            $stmt->bindValue(':hiddenFromNavFalse', false);
+        }
         $stmt->bindValue(':lang', $lang);
         if (!\is_null($rootNode)) {
             $stmt->bindValue(':left', $rootNode->getLeft());
             $stmt->bindValue(':right', $rootNode->getRight());
         }
         $stmt->execute();
+        if ($databasePlatformName == 'postgresql') {
```
acrobat, line 358: This is something I would like to solve in the query, as it could potentially a huge array to be looped over on large websites. If we can "pre-filter" these results in the query, this heavy lifting is done by the database

acrobat (changes requested):

src/Kunstmaan/NodeBundle/Repository/NodeRepository.php
```diff
@@ -266,31 +267,30 @@ public function getAllMenuNodes(
...
             $falseValue
         ) use ($databasePlatformName) {
             switch ($databasePlatformName) {
                 case 'sqlite':
                     $statement = 'CASE WHEN %s THEN %s ELSE %s END';
-
                     break;
-
+                case 'postgresql':
+                    return sprintf('COALESCE(%s, %s)', $trueValue, $falseValue);
```
acrobat, line 280:
```suggestion
                    return sprintf('COALESCE(%s, %s)', $falseValue, $trueValue);
```

As discussed on slack!

acrobat (approved):

acrobat:
Thanks @IntegraConsult!

PR #2857, opened 2021-04-05
[AdminBundle] Add config option to disabled exception logging module

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

This PR adds the config option to disable the exception logging, as most projects use an external service (like sentry) for this. When you disable it the cms doesn't save exception to the database, the exceptions toolbar item is removed and the settings/exceptions menu is removed.


ProfessorKuma (approved):

Thank you for contributing!

PR #2871, opened 2021-05-06
[MediaBundle] file whitelist

JZuidema:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

An audit was performed on one of the websites we've built with kunstmaan. They strongly recommended a whitelist for files that are being uploaded into the admin. Kunstmaan currently only has a blacklisted (which only renames the uploaded file).
This merge requests adds a whitelist, and uses both the whitelist and the blacklist to validate if the uploaded file is actually allowed to be uploaded.
By default the whitelist is not used (empty array) to make sure there are no backward compatibility breaks.


ProfessorKuma (changes requested):

ProfessorKuma (approved):

Thank you for contributing!

acrobat:
Hi @JZuidema, thanks for the PR! Related to #2680 

In the other PR there were some concerns about the proposed solution, which this PR might solve or fix in a different way. I will do review to check if the concerns are addressed in this PR or if any other changes need to be done here!

PR #2905, opened 2021-07-15
[SitemapBundle] Add pre render event for adding sitemap indexes

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Add a custom event for adding extra sitemaps to the sitemap index page. Implementation example
```
final class SitemapListener implements EventSubscriberInterface
{
    public static function getSubscribedEvents(): array
    {
        return [
            PreSitemapIndexRenderEvent::NAME => 'addSitemap',
        ];
    }

    public function addSitemap(PreSitemapIndexRenderEvent $event): void
    {
        $event->addExtraSitemap(new SitemapIndex('https://example.com'));
    }
}
```

### DIFF
--- a/src/Kunstmaan/SitemapBundle/Model/SitemapIndex.php
+++ b/src/Kunstmaan/SitemapBundle/Model/SitemapIndex.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Kunstmaan\SitemapBundle\Model;
+
+class SitemapIndex
+{
+    private $url;
+
+    public function __construct(string $url)
+    {
+        $this->url = $url;
+    }
+
+    public function getUrl(): string
+    {
+        return $this->url;
+    }
+}
--- a/src/Kunstmaan/SitemapBundle/Controller/SitemapController.php
+++ b/src/Kunstmaan/SitemapBundle/Controller/SitemapController.php
@@ -2,6 +2,7 @@
 
 namespace Kunstmaan\SitemapBundle\Controller;
 
+use Kunstmaan\SitemapBundle\Event\PreSitemapIndexRenderEvent;
 use Kunstmaan\SitemapBundle\Event\PreSitemapRenderEvent;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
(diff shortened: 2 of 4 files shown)


ProfessorKuma (changes requested):

ProfessorKuma (approved):

Thank you for contributing!

acrobat:
Hi @dannyvw, I think this feature is already implemented. Can you check #2522 if this solves your case?

dannyvw:
@acrobat This implementation is for the sitemap index (https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/SitemapBundle/Resources/views/SitemapIndex/view.xml.twig). So this is something different than the implementation in #2522. The use case for us is that we want to add a few extra sitemaps to the index page with a lot of links. The implementation in #2522 is only for adding some links to the sitemap (not index) (https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/master/src/Kunstmaan/SitemapBundle/Resources/views/Sitemap/view.xml.twig).

src/Kunstmaan/SitemapBundle/Model/SitemapIndex.php
```diff
@@ -0,0 +1,18 @@
...
+    public function __construct(string $url)
+    {
+        $this->url = $url;
+    }
+
+    public function getUrl(): string
+    {
+        return $this->url;
+    }
+}
```
acrobat, line 18: Can you also implement the other properties defined in the spec? (lastmod, priority, etc) See https://www.sitemaps.org/protocol.html

So we are fully feature complete!

acrobat (changes requested):
@dannyvw that's indeed a different and valid use case! First of all thanks for the PR! I've left some small comments and request to implement some extra properties to be feature complete with the sitemap spec (https://www.sitemaps.org/protocol.html)

src/Kunstmaan/SitemapBundle/Model/SitemapIndex.php
dannyvw: @acrobat Only the lastmod (optional) property is available for the sitemap index. I can add this but this is also not added in the current file. Do you want to add some value for the current lines? Not sure where you want to configure this value then.

PR #2944, opened 2021-09-16
[AllBundles] Add return type declarations and improve docblocks

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This PR adds return type declarations where possible (final, private and test methods) and updated phpdoc with return types. This will also prepare our code to add some return type declarations in 6.0 so we can support symfony 6. See https://wouterj.nl/2021/09/symfony-6-native-typing


ProfessorKuma (approved):

Thank you for contributing!

PR #2949, opened 2021-09-24
[NodeBundle] Deprecate SlugActionInterface and replace it by PageRender setup

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |


ProfessorKuma (approved):

Thank you for contributing!

acrobat:
@Numkil fixed!

Numkil (approved):
looks good 👍

henrypenny:
Is there any documentation on what to do with old projects that use SlugActionInterface extensively.
Not sure how you can justify saying : BC breaks? | no

acrobat:
@henrypenny There is some documentation on how to move forward in the [UPGRADE-5.9 doc](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/5.10/UPGRADE-5.9.md#nodebundle). We've replaced the old slugaction logic with the viewdataprovider and in a way they work exactly the same. This is also called before the view render in the slug controller and also the renderContext object is populated through that service. The only main difference is that this is not linked to a specific controller anymore but it exists as a separate service.

So far we didn't have a case in our upgrades (or got any other reports) that there are less possibilities with the new system in comparison with the old. So that's also the reason why "BC breaks - no" is in the PR. Existing 5.x applications continue to work and as far as we know the new system is an equal replacement.

If you still have cases that are not possible anymore, please open an issue or discussion to bring these things to our attention. From there we can see how these issues can be resolved.

PR #2979, opened 2021-10-12
[NodeBundle] Don't rely on MySQL defaults for page title field lengths

molovo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Prevents uncaught exception on saving nodes if more than 255 characters are entered into the `title` or `page_title` fields of `AbstractPage`.


acrobat (changes requested):
@molovo these changes look good! Can you just check the flintci run and apply the suggested changes? Otherwise the PR is good to go 👍

acrobat (approved):

PR #2981, opened 2021-10-12
[GeneratorBundle] Add webpack encore - keep Groundcontrol as option

dbeerten:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

TODO:

- [x] Check admin JS & CSS
- [x] Check if Webpack encore setup works (with and without demosite flag)
- [x] Check if Groundcontrol setup still works (with and without demosite flag)
- [x] ~~doc install composer require symfony/webpack-encore-bundle~~
added symfony/webpack-encore-bundle to cms-skeleton: https://github.com/Kunstmaan/cms-skeleton/pull/9
- [x] update stylelint & eslint to latest versions (incl configs)
- [x] adjust cookie bundle

BLOCKED BY:
- ~~https://github.com/symfony/webpack-encore-bundle/pull/147~~
- ~~with Groundcontrol setup, the `encore_entry_exists` twig function is not available and results in a render error: https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/3074~~

Feedback?

- Deleted the styleguide for now, should we keep it?
  _We can't use Webpack Encore to build the assets as it is not rendered by Symfony_
- By default, the cms-skeleton installs the symfony/webpack-encore-bundle package. When Groundcontrol is prefered, this package needs to be removed. Where should we document this? (also: by removing this package, the package.json gets deleted)


src/Kunstmaan/GeneratorBundle/Resources/SensioGeneratorBundle/skeleton/layout/webpack-encore/webpack.config.js
```diff
@@ -0,0 +1,112 @@
...
+    // requires WebpackEncoreBundle 1.4 or higher
+    // .enableIntegrityHashes(Encore.isProduction())
+
+    .addPlugin(new BrowserSyncPlugin(
+        {
+            open: 'external',
+            https: {
+                key: path.join(os.homedir(), '/.dinghy/certs/dinghy.key'),
+                cert: path.join(os.homedir(), '/.dinghy/certs/dinghy.crt'),
+            },
```
acrobat: Should we keep this out of the default setup? Specific to our internal local env.

// As discussed see if we can use the default dev-server setup + hmr

src/Kunstmaan/NodeBundle/Resources/views/NodeAdmin/edit.html.twig
```diff
@@ -120,7 +120,7 @@
 {% endblock %}
 
 {% block extracss %}
-    {% set cssPath = 'frontend/css/admin-style.css' %}
+    {% set cssPath = 'build/css/admin-style.css' %}
```
acrobat: Possible bc break for sites not using the new encore setup? Maybe check both paths and include if exists, the old `frontend` can be removed in 7.0

src/Kunstmaan/AdminBundle/Resources/views/Default/_js_footer.html.twig
```diff
@@ -14,7 +14,7 @@
 
         {%- block extra_async_javascripts %}{% endblock -%}
 
-        {%- set jsPath = 'frontend/js/admin-bundle-extra.js' -%}
+        {%- set jsPath = 'build/js/admin-bundle-extra.js' -%}
```
acrobat, line 20: Possible bc break for sites not using the new encore setup? Maybe check both paths and include if exists, the old frontend path can be removed in 7.0

acrobat (changes requested):

acrobat (approved):

Issue #3044, opened 2022-01-07
[DashboardBundle] ConfigHelper requires `kuma_analytics_config` table to be present on constuct


jazzzpor:
Steps to reproduce:

1. Create new checkout of project with KunstmaanDashboardBundle enabled
2. run `composer install`
3. run `bin/console`

Expected result: A list of commands that are available
Actual result: "SQLSTATE[42S02]: Base table or view not found: 1146 Table 'amarant.kuma_analytics_config' doesn't exist "

The problem lies within `\Kunstmaan\DashboardBundle\Helper\Google\Analytics\ConfigHelper`. 
This service calls an `init` method on constuct which in turn assumes the database is up and running. 

I'm working on a PR and will submit it when it's done.

acrobat:
@jazzzpor can you check if #3056 fixes your problem? This should workaround the issue and allows us to really refactor that class in the future.

jazzzpor:
@acrobat #3056 does indeed fix the problem :+1:

acrobat:
Thanks for verifying!

PR #3045, opened 2022-01-07
[DashboardBundle] Fix ConfigHelper that requires kuma_analytics_confi…

jazzzpor:
Fix ConfigHelper that requires kuma_analytics_config table to be present on constuct.

| Q      | A |
| ----------- | ----------- |
| Bug fix?      | Yes       |
| New feature?   | No        |
| BC breaks?   | No        |
| Deprecations?   | No        |
| Deprecations?   | No        |
| Fixed tickets   | #3044         |

Note:
It's not considered good practice to call `init` methods in `__construct` methods.
This PR does not fix the underlying problem and introduces a new query every time this service is used to check if the required table exists if the token is not set in `$token`.
A better solution would be to rework the usage of `\Kunstmaan\DashboardBundle\Helper\Google\Analytics\ConfigHelper` to remove the `init` method from the constructor and enforce the `init` call wherever the service is used. This however would cause BC breaks.

### DIFF
--- a/src/Kunstmaan/DashboardBundle/Helper/Google/Analytics/ConfigHelper.php
+++ b/src/Kunstmaan/DashboardBundle/Helper/Google/Analytics/ConfigHelper.php
@@ -40,6 +40,11 @@ class ConfigHelper
      */
     public function init($configId = false)
     {
+        // check if table exists before moving on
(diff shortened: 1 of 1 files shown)


jazzzpor:
This MR can be closed in favor of #3056

PR #3056, opened 2022-01-14
[Dashboardbundle] Make ConfigHelper service lazy to avoid constructor queries

### LINKED ISSUE #3044: [DashboardBundle] ConfigHelper requires `kuma_analytics_config` table to be present on constuct

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | fixes #3044 closes #3045 

This class needs refactoring to move the init function from the constructor, this PR adds a workaround so this issue can already be avoided and refactored in the future.

### DIFF
--- a/src/Kunstmaan/DashboardBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/DashboardBundle/Resources/config/services.yml
@@ -41,6 +41,8 @@ services:
         class: 'Kunstmaan\DashboardBundle\Helper\Google\Analytics\ConfigHelper'
         arguments: ['@kunstmaan_dashboard.helper.google.analytics.service', '@doctrine.orm.entity_manager']
         public: true
+        # This is needed to avoid db queries in the service constructor (init()). Can be removed after refactor.
+        lazy: true
 
     # query helper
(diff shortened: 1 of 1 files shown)


jazzzpor:
@acrobat will this MR also be merged into 6.*?

acrobat:
Yes, I've upmerged that change to 6.x so this bugfix will be included in all version starting from 5.9.3

jazzzpor:
Awesome thanks!

PR #3097, opened 2022-03-03
[AdminBundle] Fix user entity BC break

### LINKED ISSUE #3077: [AdminBundle] User fields have a different column name
After upgrading to 5.10.2 the user table generates a migration. In https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/5.10/src/Kunstmaan/AdminBundle/Entity/BaseUser.php#L41 there is no column_name defined as before in https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Resources/config/doctrine-mapping/User.orm.xml#L11

`$this->addSql('ALTER TABLE kuma_users ADD usernameCanonical VARCHAR(180) NOT NULL, ADD emailCanonical VARCHAR(180) NOT NULL, ADD confirmationToken VARCHAR(255) DEFAULT NULL, DROP username_canonical, DROP email_canonical, DROP confirmation_token, DROP password_requested_at, CHANGE salt salt VARCHAR(100) DEFAULT NULL, CHANGE password password VARCHAR(100) NOT NULL');`

Also passwordRequestedAt is dropped and the methods (https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/5.10/src/Kunstmaan/AdminBundle/Entity/BaseUser.php#L694) are now returning always a hardcoded value. Is this a BC break?

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | fixes #3077

### DIFF
--- a/src/Kunstmaan/AdminBundle/Entity/BaseUser.php
+++ b/src/Kunstmaan/AdminBundle/Entity/BaseUser.php
@@ -36,7 +36,7 @@ abstract class BaseUser implements UserInterface
      *
      * @var string
      *
-     * @ORM\Column(type="string", length=180, unique=true)
+     * @ORM\Column(name="username_canonical", type="string", length=180, unique=true)
      */
     protected $usernameCanonical;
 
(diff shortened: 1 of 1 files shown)


dannyvw:
@acrobat confirmation_token should also be changed.

acrobat:
Fixed!

dannyvw (approved):

PR #3138, opened 2022-05-06
[AdminListBundle] Fix CSRF token in AdminList delete modal

molovo:
The CSRF token uses an ID which no longer matches that used in `NodeAdminController`, meaning that the delete function does not work for nodes within an AdminList. This PR updates the ID, and the name of the submit button (for consistency).

To reproduce the original error:
* Go to an AdminList containing nodes.
* Click the delete icon for one of the rows
* Confirm when the modal appears
* Instead of deleting the node, the user is redirected to the node edit page.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | N/A

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Resources/views/AdminListTwigExtension/sure-modal.html.twig
+++ b/src/Kunstmaan/AdminListBundle/Resources/views/AdminListTwigExtension/sure-modal.html.twig
@@ -12,7 +12,7 @@
                 </h3>
             </div>
             <form action="{{ path(adminlist.getDeleteUrlFor(item)["path"], adminlist.getDeleteUrlFor(item)[("params")] ) }}" method="POST">
-                <input type="hidden" name="token" value="{{ csrf_token('delete-'~adminlist.configurator.entityName|slugify) }}"/>
+                <input type="hidden" name="token" value="{{ csrf_token('node-delete') }}"/>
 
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/AdminListBundle/Resources/views/AdminListTwigExtension/sure-modal.html.twig
```diff
@@ -12,7 +12,7 @@
                 </h3>
             </div>
             <form action="{{ path(adminlist.getDeleteUrlFor(item)["path"], adminlist.getDeleteUrlFor(item)[("params")] ) }}" method="POST">
-                <input type="hidden" name="token" value="{{ csrf_token('delete-'~adminlist.configurator.entityName|slugify) }}"/>
+                <input type="hidden" name="token" value="{{ csrf_token('node-delete') }}"/>
```
acrobat, line 15: This is not correct as it breaks the other places where an adminlist delete modal is used as this modal is also used in general (non-node related) adminlists.

Let me see if we can switch this in case of a node delete to a generic name

acrobat:
There is currently no easy way to support both csrf token for nodes and adminlist managing specific node types, this will be fixed at a later time. See #3153

PR #3143, opened 2022-06-10
[NodeBundle] issue where max weight could be exceeded and cause irregular behaviour with sorting

Numkil:
…
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/NodeBundle/Form/NodeMenuTabTranslationAdminType.php
+++ b/src/Kunstmaan/NodeBundle/Form/NodeMenuTabTranslationAdminType.php
@@ -2,15 +2,27 @@
 
 namespace Kunstmaan\NodeBundle\Form;
 
+use Doctrine\ORM\EntityManagerInterface;
+use Kunstmaan\NodeBundle\Entity\NodeTranslation;
 use Kunstmaan\NodeBundle\Form\Type\SlugType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
 use Symfony\Component\OptionsResolver\OptionsResolver;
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/NodeBundle/Form/NodeMenuTabTranslationAdminType.php
```diff
@@ -30,6 +42,22 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            if ($nt instanceof NodeTranslation && $nt->getNode()->getParent() !== null) {
+                $maxWeight = $this->em->getRepository(NodeTranslation::class)->getMaxChildrenWeight($nt->getNode()->getParent());
+
+                $event->getForm()->add('weight', ChoiceType::class, [
+                    'label' => 'kuma_node.form.menu_tab_translation.weight.label',
+                    'choices' => array_combine(range(0-$maxWeight, $maxWeight+1), range(0-$maxWeight, $maxWeight+1)),
+                    'required' => false,
+                    'placeholder' => false,
+                    'attr' => ['title' => 'kuma_node.form.menu_tab_translation.weight.title'],
+                    'choice_translation_domain' => false,
```
acrobat: @Numkil Check if you can refactor this to get the original options and only change the choices. To avoid duplicating all options

eg.
```php
$options = $form->get('weight')->getConfig()->getOptions();
// Change options
$form->add('weight', ChoiceType::class, $options);
```

src/Kunstmaan/NodeBundle/Resources/config/services.yml
```diff
@@ -64,6 +64,13 @@ services:
         tags:
             - { name: 'form.type' }
 
+    kunstmaan_node.form.type.nodemenutabtranslation:
+        class: Kunstmaan\NodeBundle\Form\NodeMenuTabTranslationAdminType
```
acrobat: this was no service previously so use the FQCN as the service name

PR #3145, opened 2022-07-06
[MediaBundle] avoid deprecation passing null to php method

Numkil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |

### DIFF
--- a/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
+++ b/src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
@@ -108,18 +108,20 @@ class FileHandler extends AbstractMediaHandler
         return FileType::class;
     }
 
-    /**
-     * @param mixed $object
-     *
-     * @return bool
-     */
-    public function canHandle($object)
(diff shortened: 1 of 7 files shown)


src/Kunstmaan/MediaBundle/Helper/File/FileHandler.php
```diff
@@ -108,20 +108,22 @@ public function getFormType()
...
+        if($object instanceof Media) {
+            if($object->getLocation() === 'local') {
+                return true;
+            }
+
+            if($object->getContent() !== null && is_file($object->getContent())) {
+                return true;
+            }
+        }
+
```
acrobat, line 126:
```suggestion
        if (!$object instanceof Media) {
            return false;
        }
        
        if ($object->getLocation() === 'local') {
            return true;
        }

        if ($object->getContent() !== null && is_file($object->getContent())) {
            return true;
        }
```

1 less level of nesting, makes it easier to read

PR #3225, opened 2022-11-23
[AdminBundle] Store original url so images have the correct url when previewing in editor

### LINKED ISSUE #3224: Image url is removed when converting urls to tokens in WYSIWYG editor.
Hi, I got a problem with  WYSIWYG fields. 

If a user inserts an image the image will be stored as 

`
`
but when the user then wants to edit the text later, the img tag has been reduced to 

`` 

 which results in the images missing when editing texts.

NB:  (The tokens are different - that is not the issue here).

This is using Kunstmaan 6.1.5.

tarjei:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #3224 

Refs #3224. Images do not get shown in the editor because the url is reduced to the token (i.e. src="[xxx]"). By storing the original url in data-src this gets handled propperly in the transform stage.

### DIFF
--- a/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
+++ b/src/Kunstmaan/AdminBundle/Form/MediaTokenTransformer.php
@@ -74,7 +74,8 @@ class MediaTokenTransformer implements DataTransformerInterface
 
                     return;
                 }
-
+                // keep url for reverse transform.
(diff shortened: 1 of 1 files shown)


acrobat:
Good catch, thanks @tarjei!

PR #3237, opened 2023-01-04
[MediaBundle] Only get soundcloud thumbnail if api key is available

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Soundcloud triggered an error because of the api key and you cannot get new keys. So I changed the default value to null and add a check for getting the thumbnail url. 

Official reaction from soundcloud:

> ”Unfortunately we're currently not accepting any new developer requests for API keys and don't have an ETA of when it will be open again. As this is something to be decided and handled by the engineering side of the organisation- any updates will be posted on the developer blog https://developers.soundcloud.com/blog.”

### DIFF
--- a/src/Kunstmaan/MediaBundle/Helper/RemoteAudio/RemoteAudioHandler.php
+++ b/src/Kunstmaan/MediaBundle/Helper/RemoteAudio/RemoteAudioHandler.php
@@ -103,7 +103,7 @@ class RemoteAudioHandler extends AbstractMediaHandler
         // update thumbnail
         switch ($audio->getType()) {
             case 'soundcloud':
-                if ($this->getSoundcloudApiKey()) {
+                if (null !== $this->getSoundcloudApiKey()) {
                     $scData = json_decode(
                         file_get_contents(
                             'http://api.soundcloud.com/tracks/' . $code . '.json?client_id=' . $this->getSoundcloudApiKey()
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/MediaBundle/Helper/RemoteAudio/RemoteAudioHandler.php
```diff
@@ -103,14 +103,16 @@ public function prepareMedia(Media $media)
...
             case 'soundcloud':
-                $scData = json_decode(
-                    file_get_contents(
-                        'http://api.soundcloud.com/tracks/' . $code . '.json?client_id=' . $this->getSoundcloudApiKey()
-                    )
-                );
-                $artworkUrl = $scData->artwork_url;
-                $artworkUrl = str_replace('large.jpg', 't500x500.jpg', $artworkUrl);
-                $audio->setThumbnailUrl($artworkUrl);
+                if ($this->getSoundcloudApiKey()) {
```
acrobat:
```suggestion
                if (null !== $this->getSoundcloudApiKey()) {
```

acrobat (changes requested):

acrobat (approved):

acrobat:
Thanks @dannyvw

PR #3239, opened 2023-01-11
[RedirectBundle] Improve performance of the redirect router

### LINKED ISSUE #3228: [RedirectBundle] Slow performance with many redirects (4000+)
If you have a lot of redirects then the redirect bundle is very slow. In one of our applications this has a performance impact of 53%. So with disabling the method initRoutes in https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/6.x/src/Kunstmaan/RedirectBundle/Router/RedirectRouter.php#L108 the performance increase with 53%.

Blackfire profile for initRoutes
image

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | fixes #3228 

I've tested this with a fresh demo site setup in production mode and with a set of 100 000 redirects and this resulted in a huge improvement in both request time and memory usage.

Overview:
image

Request time:
image

Memory usage:
image

### DIFF
--- a/UPGRADE-6.3.md
+++ b/UPGRADE-6.3.md
@@ -6,6 +6,39 @@ General
 
 - The supported Symfony version is 5.4.
 
+RedirectBundle
+--------------
+
+- The redirect-bundle has now some improved logic in the router itself. Using the old router logic is deprecated and the new will be the default in 7.0.
+  To enable the new and improved router, set the `kunstmaan_redirect.enable_improved_router` config to `true`.
+
+  This improved router has the following changes:
+  - Huge speed improvement with larger sets of cms redirects. See [#3239](https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/3239)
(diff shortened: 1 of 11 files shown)


dannyvw:
This code does not work on 6.1 due too low doctrine/dbal version. For 6.2 there are no issues.

The following use cases does currently not work, but works with the old code:

- if origin does not start with a slash in the db
- if domain is an empty string in the db. Empty string https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/6.x/src/Kunstmaan/RedirectBundle/Form/RedirectAdminType.php#L20

Other than that it looks good :)

dannyvw:
> if origin does not start with a slash in the db

Query does not match 

Extra use case:

- Wildcard redirects working different. Before a request to /test/test with configuration origin (/test/*) and target (/new) redirect to /new/test. Now it redirects to /new

dannyvw (approved):

PR #3242, opened 2023-01-26
[MediaPagePartBundle] Add admin view for media pageparts

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Define admin views for media pageparts so we can do something different in the admin view


acrobat:
@dannyvw I will merge this one for now but I would suggest to move these pageparts to the application itself (The default pagepart generator/command will already do this during the initial setup). As the MediaPartPartBundle will be deprecated in the future because of that reason.

dannyvw:
@acrobat Thanks for the info. We have a different setup so we don't copy the pageparts inside each project. When this bundle is deprecated/removed we will change something on our side.

PR #3254, opened 2023-02-13
[AdminBundle] Fix PHP Deprecated:  trim(): Passing null to parameter #1

tarjei:
PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in PHP 8.2.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | none.


acrobat:
Thanks @tarjei!

PR #3260, opened 2023-03-11
[GeneratorBundle] Only enable tty on platforms supporting it

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | related to Kunstmaan/cms-skeleton#13

@brabo-dan can you check if this solves the error you got with the installer on windows?


brabo-dan:
@acrobat now there is another windows problem. Command `bin/console` cannot be run directly. It needs be called as `php bin/console` on windows.

Original error message:

Step 6: "doctrine:fixtures:load" - [FAILED] Message: '"bin/console"' is not recognized as an internal or external command,
operable program or batch file.

PR #3267, opened 2023-05-09
[AdminBundle] trim(null) is not allowed

tarjei:
This causes errors on modern PHPs.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets |


acrobat:
Thanks @tarjei

PR #3292, opened 2023-06-12
[PagePartBundle] Allow deleting nested sub-entities

### LINKED ISSUE #3218: No longer able to delete sub-sub-entities from page parts
We have recently had reports from many clients that they get server errors when trying to delete items from their pages, most of our sites have been upgraded to 6.1.* over the last year but I haven't been able to work out specifically when this started happening.

We commonly use this structure for page parts:

Page Part -> Sub Entity (1toMany) -> Images (1toMany)

An example could be a page part that has a title, then has `n` rows of which each have `n` images.

The error we are getting is something like

```
call_user_func(): Argument #1 ($callback) must be a valid callback, class App\Entity\PageParts\TwoColImageTextPagePart does not have a method "getTwoColImageTexts_0_images"
```

It looks like the problem is coming from https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/05307f9e8c60e9c074874353ee4e19136fe32f09/src/Kunstmaan/PagePartBundle/PagePartAdmin/PagePartAdmin.php#L166 which looks like it should only be handling 1 level of sub entities.

Is this a bug or just us doing something that Kuma doesn't support?

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | fixes #3218 

As far as I could see from the previous code it should not be possible to delete nested sub entities (pp with sub entities works, pp with sub entities with sub entities could not work regarding the regex). So I've updated the code to recursively look for delete entities so it should work now with any level of sub entities.

@iknowfoobar @jerome2710 can you test this PR if it resolves the issue of #3218

### DIFF
--- a/src/Kunstmaan/PagePartBundle/Dto/PagePartDeleteInfo.php
+++ b/src/Kunstmaan/PagePartBundle/Dto/PagePartDeleteInfo.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Kunstmaan\PagePartBundle\Dto;
+
+/**
+ * @internal
+ */
+final class PagePartDeleteInfo
+{
+    private string $name;
+    private string $id;
+    private ?PagePartDeleteInfo $nestedDeleteInfo;
+
+    public function __construct(string $name, string $id, ?self $nestedDeleteInfo)
+    {
+        $this->name = $name;
+        $this->id = $id;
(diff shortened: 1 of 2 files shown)


iknowfoobar:
@acrobat That seems to be working fine for me, thanks!

PR #3300, opened 2023-07-12
[RedirectBundle] Add order by to redirect admin list

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

If you have 52 results then page 6 will display the first 2 results instead of the last 2.

Other option is to set fetchJoinCollection to false for the redirect overview. This requires some extra work because this is not configurable inside AbstractDoctrineORMAdminListConfigurator.


acrobat:
Thanks @dannyvw

PR #3331, opened 2024-02-07
[SearchBundle, NodeSearchBundle] : change `nGram` to `ngram`

delboy1978uk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes|
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | NA

When using elastic search 8, elastica throws the following Exception

The [nGram] tokenizer name was deprecated in 7.6. Please use the tokenizer name to [ngram] for indices created in versions 8 or higher instead.

Version 7 uses ngram as well as nGram as can be seen here, https://www.elastic.co/guide/en/elasticsearch/reference/7.14/analysis-ngram-tokenizer.html#_example_output_13 , so this pull request simply lower cases the G's, which allows the search population command to successfully run again

### DIFF
--- a/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
+++ b/src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
@@ -327,7 +327,7 @@ class NodePagesConfiguration implements SearchConfigurationInterface
         $ngramDiff = 1;
         if (isset($analysers['tokenizer']) && count($analysers['tokenizer']) > 0) {
             foreach ($analysers['tokenizer'] as $tokenizer) {
-                if ($tokenizer['type'] === 'ngram') {
+                if (\strtolower($tokenizer['type']) === 'ngram') {
                     $diff = $tokenizer['max_gram'] - $tokenizer['min_gram'];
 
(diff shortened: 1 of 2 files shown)


src/Kunstmaan/NodeSearchBundle/Configuration/NodePagesConfiguration.php
```diff
@@ -327,7 +327,7 @@ public function setAnalysis(Index $index, AnalysisFactoryInterface $analysis)
         $ngramDiff = 1;
         if (isset($analysers['tokenizer']) && count($analysers['tokenizer']) > 0) {
             foreach ($analysers['tokenizer'] as $tokenizer) {
-                if ($tokenizer['type'] === 'nGram') {
+                if ($tokenizer['type'] === 'ngram') {
```
acrobat: I would add a strtolower around `$tokenizer['type']` to be backwards compatible with older code, we can remove it later when only ES8+ is supported as everyone using `nGram` will have got the elasticsearch warning by then

acrobat (changes requested):

acrobat:
Thank you @delboy1978uk

acrobat:
Thanks @delboy1978uk

Issue #3343, opened 2024-03-05
[AdminBundle] - removed classes are still in service config


delboy1978uk:
form types deprecated in v6 and removed in v7 are still being referred to in the services yaml

PR #3344, opened 2024-03-05
Refactor service definitions replacing removed classes

delboy1978uk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
Fixed tickets | #3343

### DIFF
--- a/src/Kunstmaan/AdminBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/AdminBundle/Resources/config/services.yml
@@ -173,21 +173,11 @@ services:
         tags:
             - { name: cache.pool }
 
-    kunstmaan_admin.form.type.color:
-        class: Kunstmaan\AdminBundle\Form\ColorType
-        tags:
-            - { name: form.type }
-
     kunstmaan_admin.doctrine_mapping.listener:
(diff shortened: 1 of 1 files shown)


src/Kunstmaan/AdminBundle/Resources/config/services.yml
```diff
@@ -174,7 +174,7 @@ services:
             - { name: cache.pool }
 
     kunstmaan_admin.form.type.color:
```
acrobat: Both the services can be deleted as the symfony form types are registered by symfony themself. We don't need the extra services anymore.

acrobat (changes requested):
That was indeed an incorrect leftover of v6. Can you target the PR against the 7.0 branch? The cleanup was done on that branch.

And see my comment inline

src/Kunstmaan/AdminBundle/Resources/config/services.yml
delboy1978uk: I'll strip them completely out then! Thanks  again!

acrobat (approved):

acrobat:
Thanks @delboy1978uk

PR #3348, opened 2024-03-14
[RedirectBundle] Configure redirect admin list configurator as service

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

We needed some custom logic in this class. Do I need to add some BC layer?

### DIFF
--- a/src/Kunstmaan/RedirectBundle/Resources/config/services.yml
+++ b/src/Kunstmaan/RedirectBundle/Resources/config/services.yml
@@ -22,8 +22,13 @@ services:
             - { name: form.type, alias: kunstmaan_redirect_form_type }
 
     Kunstmaan\RedirectBundle\Controller\RedirectAdminListController:
-        arguments: ['@kunstmaan_admin.domain_configuration']
+        arguments: ['@Kunstmaan\RedirectBundle\AdminList\RedirectAdminListConfigurator']
         calls:
             - setContainer: [ ] # Container service locator will be injected filled by the services from subscribed services
         tags: ['controller.service_arguments', 'container.service_subscriber']
 
+    Kunstmaan\RedirectBundle\AdminList\RedirectAdminListConfigurator:
+        arguments:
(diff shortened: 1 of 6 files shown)


acrobat (changes requested):
This will indeed need a BC layer. I think something like in the translation controller can be implemented in a BC way

https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/cc640556e5cac9874ea021a0a897c00114ff590c/src/Kunstmaan/TranslatorBundle/Controller/TranslatorController.php#L246-L260

src/Kunstmaan/RedirectBundle/Resources/config/services.yml
```diff
@@ -22,8 +22,14 @@ services:
...
         calls:
             - setContainer: [ ] # Container service locator will be injected filled by the services from subscribed services
         tags: ['controller.service_arguments', 'container.service_subscriber']
 
+    Kunstmaan\RedirectBundle\AdminList\RedirectAdminListConfigurator:
+        arguments:
+            $em: '@doctrine.orm.entity_manager'
+            $domainConfiguration: '@kunstmaan_admin.domain_configuration'
+        autowire: true
+        autoconfigure: true
```
acrobat:
```suggestion
            - '@doctrine.orm.entity_manager'
            - '@kunstmaan_admin.domain_configuration'
```

Won't use autowiring/autoconfigure in bundles as the service config should be explicit, also don't use named arguments as the variables names are not included in the bc policy

acrobat (changes requested):

src/Kunstmaan/RedirectBundle/Resources/config/services.yml
dannyvw: @acrobat Done

acrobat (approved):

acrobat:
Thanks @dannyvw!

PR #3406, opened 2024-04-16
Fix bad responsetype in reorder action

tarjei:
IMHO This should warrant an urgent release.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |


acrobat:
Thanks for the PR @tarjei! This fix was included in #3408 together with another bugfix in the reorder action

PR #3407, opened 2024-04-18
[PagePartBundle] Add page to PagePartEvent

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

### DIFF
--- a/src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
+++ b/src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
@@ -2,6 +2,7 @@
 
 namespace Kunstmaan\PagePartBundle\Event;
 
+use Kunstmaan\PagePartBundle\Helper\HasPagePartsInterface;
 use Kunstmaan\PagePartBundle\Helper\PagePartInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Contracts\EventDispatcher\Event;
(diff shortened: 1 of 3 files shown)


src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
```diff
@@ -18,9 +19,12 @@ final class PagePartEvent extends Event
      */
     private $response;
 
-    public function __construct(PagePartInterface $pagePart)
+    private ?HasPagePartsInterface $page;
+
+    public function __construct(PagePartInterface $pagePart, ?HasPagePartsInterface $page = null)
     {
         $this->pagePart = $pagePart;
+        $this->page = $page;
```
acrobat, line 27: Can you add a deprecation here if $page is null, so we can make it madatory in the next major as it is logic to have the page always available. Also add an entry to the upgrade file that the $page will be required in 8.0

src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
```diff
@@ -42,4 +46,14 @@ public function setResponse(Response $response)
     {
         $this->response = $response;
     }
+
+    public function getPage(): ?HasPagePartsInterface
+    {
+        return $this->page;
+    }
+
+    public function setPage(?HasPagePartsInterface $page): void
```
acrobat: Do you specifically need the setter? Otherwise I would propose to leave it out, we can always add afterwards if there is a use case

acrobat (changes requested):

src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
dannyvw: @acrobat  Done

src/Kunstmaan/PagePartBundle/Event/PagePartEvent.php
dannyvw: Not needed so I removed this setter

acrobat (approved):

acrobat:
Thanks @dannyvw

PR #3414, opened 2024-04-22
[AdminBundle] fixed base route name for exception admin lists

virtualize:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

I don't know if this the admin list for exceptions is still used. Or if exceptions are still logged in 7.1. Maybe it's deprecated and there's a config flag to disable it. Haven't found anything on the docs.

Anyway, in our admin UI there's a menu entry for it that throws route-not-found exceptions. So here is a fix for it.


acrobat:
Thanks @virtualize. You can disable the cms error logging with the `kunstmaan_admin.exception_logging.enabled: false` if you don't need it in your application because you use sentry for example.

PR #3421, opened 2024-06-06
[FormBundle] Fix form submission export

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

The old "iterate" method returns an array, but "toIterable" returns a object. It triggers now the following error:

Cannot use object of type Kunstmaan\FormBundle\Entity\FormSubmission as array on kunstmaan/bundles-cms/src/Kunstmaan/FormBundle/AdminList/FormSubmissionExportListConfigurator.php line 130


acrobat:
Thanks @dannyvw

PR #3435, opened 2024-09-05
[UtilitiesBundle] Fix instanceof proxy

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

When fixing this deprecation (https://github.com/doctrine/orm/pull/10837) then the getClass method returns the proxy class instead of the real class. This check works for both enable_lazy_ghost_objects true/false.


acrobat:
Thanks @dannyvw

PR #3444, opened 2024-10-08
[MultidomainBundle] Improve domain based locale router performance

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes?
| Deprecations? | no

This line https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/7.x/src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php#L175 decrease performance when you have many links. The "isMultiLanguage" check is also done inside "getSlugRouteParameters" so it is not necessary to perform that inside the "getRouteCollection" for every link. Runtime checks inside "getRouteCollection" are slow (https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Routing/RouterInterface.php#L29)

We have about 20/25% profit on a page with only a menu and footer.

The only BC break is as the removed methods are overridden. We can solve it maybe with an config setting to switch to the improved router? Is this needed to implement?

### DIFF
--- a/src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
+++ b/src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
@@ -9,14 +9,10 @@ use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 use Symfony\Component\Routing\Generator\UrlGenerator;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\Matcher\UrlMatcher;
-use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
 
 class DomainBasedLocaleRouter extends SlugRouter
 {
-    /** @var RouteCollection */
-    protected $routeCollectionMultiLanguage;
-
     /**
      * @var array|null
      */
@@ -172,17 +168,6 @@ class DomainBasedLocaleRouter extends SlugRouter
      */
     public function getRouteCollection(): RouteCollection
     {
-        if (($this->otherSite && $this->isMultiLanguage($this->otherSite['host'])) || (!$this->otherSite && $this->isMultiLanguage())) {
-            if (!$this->routeCollectionMultiLanguage) {
-                $this->routeCollectionMultiLanguage = new RouteCollection();
-
-                $this->addMultiLangPreviewRoute();
-                $this->addMultiLangSlugRoute();
-            }
-
-            return $this->routeCollectionMultiLanguage;
-        }
-
         if (!$this->routeCollection) {
             $this->routeCollection = new RouteCollection();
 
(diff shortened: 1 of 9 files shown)


src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
```diff
@@ -9,23 +9,12 @@
...
-    /**
-     * @var array|null
-     */
-    private $otherSite;
-
-    /**
-     * @var array
-     */
-    private $cachedNodeTranslations = [];
+    private ?array $otherSite = null;
```
acrobat, line 24: The switch from protected to private properties is a BC break and should be reverted

src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
```diff
@@ -9,23 +9,12 @@
 use Symfony\Component\Routing\Generator\UrlGenerator;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\Matcher\UrlMatcher;
-use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
 
 class DomainBasedLocaleRouter extends SlugRouter
 {
-    /** @var RouteCollection */
-    protected $routeCollectionMultiLanguage;
```
acrobat, line 18: If this property is not needed anymore we need to deprecate it so it can be removed in 8.0

src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
```diff
@@ -202,39 +167,6 @@ protected function addSlugRoute()
         $this->addRoute('_slug', $routeParameters);
     }
 
-    /**
-     * Add the slug route to the route collection
-     */
-    protected function addMultiLangPreviewRoute()
```
acrobat, line 208: These methods need to be deprecated for removal in 8.0

acrobat (changes requested):
As you mentioned already, we need to keep BC and can't remove those properties and methods (or change visibility)

I like the performance improvements here, so indeed with some kind of option/toggle would be ideal to make improvements and keep BC. Do you have any numbers on the improvement itself? Like blackfire profile of some timings of real apps?

src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
dannyvw: This property was already private

dannyvw:
@acrobat I pushed some changed. I cannot generate the same blackfire profile as before because we have changed a lot. But the main thing is that is reduced calls to getHost and isMultiLanuage methods.

src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
```diff
@@ -172,15 +156,19 @@ private function getReverseLocaleMap(): array
      */
     public function getRouteCollection(): RouteCollection
     {
-        if (($this->otherSite && $this->isMultiLanguage($this->otherSite['host'])) || (!$this->otherSite && $this->isMultiLanguage())) {
-            if (!$this->routeCollectionMultiLanguage) {
-                $this->routeCollectionMultiLanguage = new RouteCollection();
+        if (!$this->enabledImprovedRouter) {
+            trigger_deprecation('kunstmaan/multidomain-bundle', '7.2', 'Not enabling the improved router is deprecated and the changed and improved router will be the default in 8.0. Set the "kunstmaan_multi_domain.enable_improved_domain_based_local_router" config to true.');
```
acrobat:
```suggestion
            trigger_deprecation('kunstmaan/multidomain-bundle', '7.2', 'Not enabling the improved router is deprecated and the changed and improved router will be the default in 8.0. Set the "kunstmaan_node.enable_improved_domain_based_local_router" config to true.');
```

The option is configured on the node bundle currently or is it added in the wrong bundle? And if it is the node-bundle config the config option itself is incorrect `enable_improved_router`

acrobat (changes requested):
@dannyvw small remark regarding the deprecation message and suggested config. Otherwise PR is good to go 👌

src/Kunstmaan/MultiDomainBundle/Router/DomainBasedLocaleRouter.php
dannyvw: I didn't get it working when adding the configuration option inside the multi domain bundle. Maybe it missed something, so I added it inside the node bundle.

acrobat (approved):

acrobat:
Thanks @dannyvw!

PR #3453, opened 2024-10-14
[PagePart] (fix) skip checking for not persisted sub entites delete c…

mart-insiders:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Fixed a bug when persisting a pagepart with subentities where one is being deleted whilst not yet persisted. This causes the preg_match checks to not match and cause a 500 error. Later when iterating through the $subPagePartsToDelete, the $deleteInfo object is null and will also cause a 500 error.


acrobat (approved):

acrobat:
Thanks @mart-insiders

PR #3459, opened 2024-11-08
[AllBundles] Fix fixture deprecations

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes

Method "Doctrine\Common\DataFixtures\FixtureInterface::load()" might add "void" as a native return type declaration in the future. Do the same in implementation "Kunstmaan\GeneratorBundle\DataFixtures\ORM\GroupFixtures" now to avoid errors or add an explicit @return annotation to suppress this message.


acrobat (approved):
Technically adding those return types are a BC break, but I don't expect anyone extending the fixture classes so let's merge as is! Thanks @dannyvw!

PR #3473, opened 2024-11-14
array_key_exists only if is_array

jerome2710:
| Q             | A
| ------------- | ---
| Bug fix?      | yes|no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 3449

### DIFF
--- a/src/Kunstmaan/AdminListBundle/Service/ExportService.php
+++ b/src/Kunstmaan/AdminListBundle/Service/ExportService.php
@@ -78,7 +78,7 @@ class ExportService
 
             $iterator = $adminList->getIterator();
             foreach ($iterator as $item) {
-                if (\array_key_exists(0, $item)) {
+                if (\is_array($item) && \array_key_exists(0, $item)) {
                     $itemObject = $item[0];
                 } else {
                     $itemObject = $item;


src/Kunstmaan/AdminListBundle/Service/ExportService.php
```diff
@@ -78,7 +78,7 @@ protected function streamOutput(ExportableInterface $adminList, $format)
 
             $iterator = $adminList->getIterator();
             foreach ($iterator as $item) {
-                if (\array_key_exists(0, $item)) {
+                if (\is_array($item) && \array_key_exists(0, $item)) {
```
dannyvw, line 81: See also https://github.com/Kunstmaan/KunstmaanBundlesCMS/pull/3421, so I think it can be simplified to `$itemObject = $item;` Or remove everything and change the var `$itemObject` to `$item`

acrobat: So what would be the correct fix here?

dannyvw: You can do this: Or remove everything and change the var $itemObject to $item

tarjei:
@acrobat  will you be merging this? It has broken the export functionality

acrobat:
Sorry @tarjei, I've lost track of this issue! Merged now! I need to check if there are other PR's that are ready to merge, after that I will push new releases!

tarjei:
@acrobat NP :)

PR #3477, opened 2024-11-29
[MultiDomainBundle] Prevent duplicate calls for getting node translation

dannyvw:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Use array_key_exists instead of isset because the value inside cachedNodeTranslations can be null


acrobat (approved):

acrobat:
Thanks @dannyvw

PR #3483, opened 2025-01-30
[NodeBundle] | (fix) nullable nodetranslation in slugtype

mart-insiders:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

The buildView method in SlugType can throw a 500 error when a nodeTranslation is not found in  parentNode->getNodeTranslation(). Adding a simple nullcheck resolves this. 
This bug occurs when a multilangual site had a subpage under a parent node which is not translated. This can be resolved by creating the parent node translation but shouldn't crash the CMS


acrobat:
@mart-insiders can you rebase against the current 7.x branch and target the PR to 7.x. Master is no longer used and is an old state of the project. Thanks!

